PR #25, opened 2011-12-13
sync up with gbp stuff

ethanrublee:
This is still a work in progress.  I have a few fixes related to python 2.6 version.
Also a bit of cmake for making a dist tarball and gbp.  Expect gbp stuffs to change a lot as i figure out the work flow.


ethanrublee:
marking invalid.

Issue #50, opened 2012-01-11
install fails on OSX


gerkey:
On OSX (10.7.2, Python 2.7.1), 'make install' yields:

```
COMMAND=
+ '[' -n '' ']'
+ cd /Users/gerkey/code/catkin/src/catkin
+ /usr/bin/env PYTHONPATH=/Users/gerkey/code/catkin/src/catkin/python:/tmp/catkin/lib/python2.7/dist-packages:/Users/gerkey/code/catkin/src/build/gen/py /usr/bin/python /Users/gerkey/code/catkin/src/catkin/setup.py install --install-layout=deb --prefix=/tmp/catkin
usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: setup.py --help [cmd1 cmd2 ...]
   or: setup.py --help-commands
   or: setup.py cmd --help

error: option --install-layout not recognized
CMake Error at catkin/CMakeFiles/safe_execute_install.cmake:5 (message):
  Error executing process with arguments
Call Stack (most recent call first):
  catkin/cmake_install.cmake:106 (INCLUDE)
  cmake_install.cmake:32 (INCLUDE)


make: *** [install] Error 1
```

Removing the `--install-layout=deb` line from cmake/templates/python_distutils_install.sh.in gets further but results in a different error:

```
COMMAND=
+ '[' -n '' ']'
+ cd /Users/gerkey/code/catkin/src/genpy
+ /usr/bin/env PYTHONPATH=/Users/gerkey/code/catkin/src/genpy/src:/tmp/catkin/lib/python2.7/dist-packages:/Users/gerkey/code/catkin/src/build/gen/py /usr/bin/python /Users/gerkey/code/catkin/src/genpy/setup.py install --prefix=/tmp/catkin
running install
Checking .pth file support in /tmp/catkin/lib/python2.7/site-packages/
/usr/bin/python -E -c pass
TEST FAILED: /tmp/catkin/lib/python2.7/site-packages/ does NOT support .pth files
error: bad install directory or PYTHONPATH

You are attempting to install a package to a directory that is not
on PYTHONPATH and which Python does not read ".pth" files from.  The
installation directory you specified (via --install-dir, --prefix, or
the distutils default setting) was:

    /tmp/catkin/lib/python2.7/site-packages/

and your PYTHONPATH environment variable currently contains:

[...]

PR #53, opened 2012-01-16
OSX s/dist/site/g abstraction

kwc:
See comments on Issue #50.  Will also issue pull request on genpy repo for the cmake file there.  I'm not terribly happy with the patch, but it works so far in testing on OS X and Ubuntu.  The IF(APPLE) should be something more like IF(NOT DEB LAYOUT)


straszheim:
Looks good to me.  Don't expect happiness when dealing with cmake.

PR #54, opened 2012-01-17
more helpful error message for missing nose

dirk-thomas:


straszheim:
This is my helpful comment on this helpful error message

PR #87, opened 2012-02-19
Librt Check

stonier:
Properly avoid the librt check on win32/apple/mingw - missed it in part the first time around. This should be right now.


kwc:
NOTE: the librt check is failing for me on Vincent's Oneiric machine.

stonier:
Do you have any output? 

I've no problem with catkin cross-compiling for mingw or natively compiling rostime on oneiric with latest trunk/tips.

kwc:
turns out that the build space can get corrupted, just need to clean it first.  Incremental updates appear to not work properly in catkin

PR #107, opened 2012-03-09
Find modules...and specifically a bzip2 find module

stonier:
I couldn't find a standard place for cmake find modules within the new catkin system so I took what was partially there and modified it a bit. It originally added the cmake/Modules directory to the CMAKE_MODULE_PATH, but the scope of the modified path only holds whilst inside the catkin package. 
Instead, I append it from the toplevel cmake - so that the catkin/cmake/Modules can be used by everyone. 

This works for the short term - long term is the catkin package the right place for an extensive library of find-xxx cmake modules?

The bzip2 find module above - it is a modified version of the official bzip2 module which fails on windows. There is no standard include/lib paths there and the module has no BZIP2_ROOT style helper variables.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,8 +11,6 @@ configure_file(cmake/tests.cmake.in
 
 include(${catkin_EXTRAS_DIR}/all.cmake)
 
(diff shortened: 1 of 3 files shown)


stonier:
Cancelling this - there is a better way to do this I believe via xxxConfig.cmake files with a bzip2 install and other cmake variables.

PR #113, opened 2012-03-15
Catkin tweaks for Archlinux

hauptmech:
Added a cmake option for non deb python packages layout for those of us that need it but aren't apple.

Fixed problem with 'nosetests' not existing in archlinux under that name.


kwc:
Closing due to #114.  Please reissue the nosetests2/3 patch separately.  Thanks!

PR #116, opened 2012-03-23
Utilise an alternative root for installed programs

stonier:
This helps find headers and libraries if alot of deps are installed under a particular root. It works...but are there other uses for CMAKE_XXXX_PATH that this might get in the way of?


gerkey:
Hmm, in that situation, I would usually set CPATH, LIBRARY_DIRS, and (if needed) LD_LIBRARY_PATH.  Is setting CMAKE_PREFIX_PATH and CMAKE_INCLUDE_PATH preferred?

stonier:
I'm a bit inclined to agree - if they wanted it used like that, why didn't they include that functionality already in CMAKE_INCLUDE_PATH? Feels like forcing it.

CPATH, LIBRARY_DIRS could probably be ok. I kind of hate having to set variables in two places (env. variables and cmake variables) though. If it can all be done via the cmake cache, it's easier to directly introspect.

Posted a question on the cmake mailing list re preferred solutions. Will see what comes out of that.

stonier:
Not much response from the cmake list. However testing the compiler - it seems that msvc's cl and gcc use quite different environment variables (e.g. INCLUDE vs CPATH (or CPLUS_INCLUDE_PATH). 

I'd prefer to have this configurable via cmake's cache. It's what cmake is designed to do - abstract platform differences and it's also a one-stop place for configuration and introspection.

However, seeing as it isn't built in, CMAKE_INCLUDE_PATH may not be the correct place to do it. Someone on the cmake list suggested just setting up our own project variable to do the job which can be (pre)configured in the cmake cache.

Thoughts?

stonier:
Updated previous commit to now include a catkin variable, CATKIN_ROSDEPS_PATH for the same effect (see the diff). I still prefer this - it's much easier having the one core cmake variable than to have to remember to set compiler specific environment variables for msvc or mingw.

stonier:
Added a rule for handling dll's in windows as discussed on the catkin mailing list.

gerkey:
Please test the result of the merge.  I tweaked the list management a bit, but it shouldn't produce different behavior.

And I can only assume that the MSVC-specific add_library() override works properly...

stonier:
I'll start the win64 builds daily on hudson so windows builds are more visible.

gerkey:
Fantastic!  Looking forward to keeping this stuff working on Windows.

stonier:
Hit a snag when testing that I didn't notice before. Running 'cd build/ros_comm/tools/rosconsole; nmake' was fine as it hit all targets, but dependency got lost in the wash when nmaking a specific target.

There's a better way by hooking ADD_CUSTOM_COMMAND as a POST_BUILD hook. Will test that and do another pull request.

PR #117, opened 2012-03-24
stack.yaml updates

isucan:
For 3rd party repos it is convenient to have the stack.yaml located only on github, not in upstream. This is used in OMPL, and I believe PCL as well.

In addition, it is important not to make too many assumptions about the tag names in upstream. For that, I added an optional Tag: element to stack.yaml, which is used by catkin-import-upstream.

### DIFF
--- a/bin/catkin_util.sh
+++ b/bin/catkin_util.sh
@@ -104,6 +104,7 @@ read_stack_yaml ()
 
     PACKAGE_NAME=$(/bin/echo $TXT | perl -ne '/Catkin-ProjectName:\s+([^\s]+)/ && print $1')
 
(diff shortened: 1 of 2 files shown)


gerkey:
Please test the result of this merge; I don't have any examples of using either feature.

And I cannot believe that there are calls to perl in there...

gerkey:
Oh, and note that I changed Tag: to Release-Tag:

isucan:
Thanks for merging this Brian!
Things appeared to work just fine locally with the latest catkin, but there
was a build failure anyway:
http://hudson.willowgarage.com:8080/job/ros-fuerte-ompl_sourcedeb/45/console

However, that appears not to have to do with OMPL itself, but to access
rights for uploading to the building server. Is that it?

Ioan

On Wed, Apr 4, 2012 at 6:45 AM, gerkey <
reply@reply.github.com

PR #120, opened 2012-04-09
Fix for the lost windows copy dll dependency.

stonier:
Previously worked if an nmake all was passed to any sub project. But wouldn't trigger if a specific target was built down the line.

Move add_custom_target -> add_custom_command as a post build hook to the real target. Fixes the problem.


gerkey:
Windows .bat programming...fun....

PR #124, opened 2012-04-24
Windows buildspace environment update

stonier:
Updating buildspace env.bat and setup.bat to align with current catkin (previously a bit old, gnarly and broken).

Some of the variables going into it need to go into the environment hooks (e.g. ros/env-hooks/10.ros.buildspace.xxx) but we'll patch them upstream later when we're a bit more sure of what needs to go into them.

`python rosmaster` now works!


gerkey:
Looks reasonable to me.  Happy to hear that rosmaster is working!

PR #126, opened 2012-05-08
Windoze updates enabling the install space

stonier:
Sets up the environment env/setup files for the install space. Also gets python doing its thing right on windows. 
- python distutils script added
- windows peculiarities for python handled in the cmake
- install locations pointing to the right places

The last one is a bit odd - windows python setuptools installs to lib/site-packages and not lib/python2.7/site-packages or lib/python2.7/dist-packages.

Are the latter two options strictly because of the install-layout=deb option to python setuptools? The dist-packages is not so important, but am curious why it doesn't do the python2.7.

There is also a patch coming to genpy to help this.

### DIFF
--- a/cmake/catkin_add_env_hooks.cmake
+++ b/cmake/catkin_add_env_hooks.cmake
@@ -41,26 +41,6 @@ endfunction()
 
 
 function(catkin_generic_hooks)
-  foreach(shell sh bash zsh)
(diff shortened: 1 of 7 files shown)


stonier:
Also added a default install command for windows add_library to make sure runtime destination gets set to bin.

It would be better to get people to always set runtime, archive and library destinations to bin-lib-lib, however...linux programmers never do that by habit. So it would mean alot of patching right across the board. If this is preferred, let me know.

PR #146, opened 2012-07-26
Find root path fixes

housel:
Fix cross-compilation builds that make use of `CMAKE_FIND_ROOT_PATH` by using `find_file()` etc. to locate files and directories that may be within a cross-compilation root environment, and by suppressing searches in the cross-compilation root for things that won't be there (such as files in the source tree).

### DIFF
--- a/cmake/templates/pkg-config.cmake.installable.in
+++ b/cmake/templates/pkg-config.cmake.installable.in
@@ -44,7 +44,8 @@ endif()
 set(@PACKAGE_NAME@_CONFIG_INCLUDED TRUE)
 
 if (NOT "X@PACKAGE_INCLUDE_DIRS@" STREQUAL "X")
-  set(@PACKAGE_NAME@_INCLUDE_DIRS @PKG_INCLUDE_PREFIX@/include) #TODO FIX this.
+  find_file(@PACKAGE_NAME@_INCLUDE_DIRS include PATHS @PKG_INCLUDE_PREFIX@
+            NO_DEFAULT_PATH)
 endif()
 
 foreach(lib @PACKAGE_LIBRARIES@)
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Thanks for contributing the patch. Since we are currently changing a lot of code in catkin I won't apply the patch directly. As soon as these changes have been pushed I will look into applying your changes.

dirk-thomas:
I applied the second changeset.

dirk-thomas:
Regarding the first changeset: I am not very familiar with the details of cross-compilation. Can you please describe why this is necessary for cross-compilation? Than I can avoid these issues in the future.

housel:
Cross compilation environments set CMAKE_FIND_ROOT_PATH to point to one or more root filesystem images for the target environment, and possibly for the cross compilation tools themselves. They also often do:

```
set( CMAKE_FIND_ROOT_PATH_MODE_PROGRAM ONLY )
set( CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY )
set( CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY )
```

in order to prevent any pollution from the host environment's libraries, include files, or CMake modules. When CMake find_\* commands are used for source files or other things that it doesn't expect to find in one of the cross-compilation root filesystems, then FIND_ROOT_PATH needs to be suppressed.

dirk-thomas:
Since a lot of the catkin internals have changed in the meantime could you please retest with the current version?

dirk-thomas:
The pull request was automatically closed when master was renamed to groovy-devel. Please reopen the pull request against groovy-devel if the issue is not yet solved in the meantime. Sorry for the inconvenience.

PR #154, opened 2012-08-23
Minor catkin windows branch updates

stonier:
Fixes a couple of entries where windows branch cmake code fails.

Batch file updates to come.


stonier:
If you wanted to remove sequential black magic in the includes in all.cmake, you could just get empy.cmake to call python.cmake itself.

dirk-thomas:
I also added CURRENT_WORKSPACE in the install case for windows.

Issue #161, opened 2012-09-06
catkin_init_workspace creates broken symlink path


tkruse:
To reproduce, follow the walkthrough:
http://ros.org/doc/groovy/api/catkin/html/walkthrough.html

After calling catkin_init_workspace, the src/CMakelists.txt symlink points to:
src/catkin/cmake/toplevel.cmake
should point to:
catkin/cmake/toplevel.cmake

Pull request #162 (sorry, could have just create pull request)

PR #162, opened 2012-09-06
bugfix catkin_init_workspace creates broken symlink path

tkruse:
To reproduce, follow the walkthrough:
http://ros.org/doc/groovy/api/catkin/html/walkthrough.html

After calling catkin_init_workspace, the src/CMakelists.txt symlink points to:
src/catkin/cmake/toplevel.cmake
should point to:
catkin/cmake/toplevel.cmake

pull request is duplicate of issue #161

### DIFF
--- a/python/catkin/init_workspace.py
+++ b/python/catkin/init_workspace.py
@@ -4,17 +4,23 @@ import shutil
 
 
 def _symlink_toplevel_cmake(src, dst, checked):
+
     # check if toplevel file exists
     src = os.path.join(src, 'toplevel.cmake')
     if not os.path.exists(src):
         checked.append(src)
         return False
 
+    if os.path.isabs(src):
+        # we prefer relative symlinks so that we can move the workspace
+        src_rel = os.path.relpath(src, dst)
+
+
     # try to symlink file
     dst = os.path.join(dst, 'CMakeLists.txt')
     try:
-        os.symlink(src, dst)
-        print('Creating symlink "%s" pointing to "%s"' % (dst, src))
+        os.symlink(src_rel, dst)
+        print('Creating symlink "%s" pointing to "%s"' % (dst, src_rel))
     except Exception:
         # try to copy file
         try:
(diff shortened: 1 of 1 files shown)


python/catkin/init_workspace.py
```diff
@@ -4,17 +4,23 @@
...
     # check if toplevel file exists
     src = os.path.join(src, 'toplevel.cmake')
     if not os.path.exists(src):
         checked.append(src)
         return False
 
+    if os.path.isabs(src):
+        # we prefer relative symlinks so that we can move the workspace
+        src_rel = os.path.relpath(src, dst)
+
```
wjwwood, line 17: Should this have an:

```
else:
    src_rel = src
```

?

python/catkin/init_workspace.py
```diff
@@ -4,17 +4,23 @@
...
+
     # try to symlink file
     dst = os.path.join(dst, 'CMakeLists.txt')
     try:
-        os.symlink(src, dst)
-        print('Creating symlink "%s" pointing to "%s"' % (dst, src))
+        os.symlink(src_rel, dst)
```
wjwwood, line 22: This isn't portable to non-Unix [http://docs.python.org/library/os.html#os.symlink], but maybe that doesn't matter.  Alternatively, we could make a copy of the file instead, what are the draw backs to this approach other than updating catkin doesn't update the symlink?

wjwwood, line 22: Could we use a cmake include in the linked/copied file to mimic a symlink?

python/catkin/init_workspace.py
tkruse, line 17: Yes, sorry about that.

dirk-thomas, line 17: A relative symlink is not possible here.
Consider the case catkin is installed in /op/ros/groovy and your workspace is somewhere - you need an absolute path for the symlink to make the workspace movable. So it is exactly the other way around.

If you would move the catkin installation any symlink (relative and absolute) would be wrong.
A relative symlink would only make sense if you move the workspace containing catkin and the workspace with the symlink together.
The code already does that if catkin is in the workspace where the symlink is created.

dirk-thomas:
I committed a fix cb3c22956544edf8882244cac873f5887084c4ca to create the correct relative symlink when catkin is used from.

Issue #165, opened 2012-09-07
catkin unit tests messy


tkruse:
Hi,

looking at the ctakin "unit" tests, they seem to be only loosely structured, badly documented, redundant, broken, etc.

I want to repair and restructure them, unless somebody thinks i should not do so.

regards,
  Thibault

tkruse:
close with #166

PR #166, opened 2012-09-09
cleanup catkin test folder

tkruse:
Hi,

this is my pull request for issue #165 , general catkin test folder messiness. This commit does a lot of things in catkin/test to clean up:
- creates some folder and moved stray files in there
- added test/README.rst to explain folder structure
- moved all shell scripts into "check" folder quarantine
- refactored all nosetests to be based on unittest.TestCase classes
- made most unit tests run again
- tests use current catkin code (by copying) instead of pulling catkin from github, which defeated the purpose of unit tests (testing before pushing)
- test/.../test.rosinstall for unit tests, rosinstall/unstable.rosinstall for dev environment setup
- unit tests create their workspaces and build folders in /tmps or in test/tmp
- I tried hard to avoid long waiting times, so rosinstall, cmake, and make for the core ros stacks should only ever run once.
- separated long-running from short-running tests

Caveat:
I could not resurrect 2 tests that were based around typelibxml, don't even know what that is. So they are left commented as network_tests/test_basic

One of the active tests stills fails, due to maybe a catkin bug. I could not find the cause, but for the project in test/mock_resources/src/catkin_test, I get:

-- Installing: /tmp/tmpX0w7na/install/share/catkin_test/stack.xml
CMake Error at catkin_test/quux_msgs/cmake_install.cmake:38 (FILE):
  file cannot create directory: /msg.  Maybe need administrative privileges.

So it seems that catkin is not using CMAKE_INSTALL_PREFIX for the .msg file of this project (though it works for other projects like std_msgs). i could not find the cause, maybe someone else can?

Also I tried rebasing against the latest master on github, but I ran into issues with the latest gtest changes.

I hope we can have the unit tests running for all catkin developers again, so that we can soon establish that nothing should be pushed to github unless all tests run.


dirk-thomas:
Since the patch is not small I have a couple of comments before we should merge it.
Please don't understand that wrong - your effort is highly appreciated.
The whole refactoring of not fetching catkin from GitHub for testing is great.

Please mention in the README how the PYTHONPATH must be set in order to run the nosetests.

The script under "check" are really severe. Can you enhance the scripts to ask for user permission with a detailed message what is going to happen? I think else people might just run these scripts without knowing the side effects (like sudo rm -rf /opt/ros/fuerte/*) 

The failing test "quux_msgs" might be due to the CMakeLists.txt. Please try to move the catkin_project() call directly belo the cmake proect() call. This file also seems to lack a find_package(catkin REQUIRED) and the existing find_package() calls should specify the REQUIRED option.

Catkin should now (as of 0.5.14) utilize CMAKE_PREFIX_PATH correctly when passed on the cmake command line. So it does no more have to be set as an env variable anymore. This version also fixes the previously unreliable gtest integration.

Regarding the distros mentioned in various tests: we should only use Groovy in there, not Fuerte anymore since the master branch of catkin is Groovy-only. Fuerte has already been branched some time ago and will not receive these changes.

And the last suggestion is for the various rosinstall files which define the repos to test with. These will be(come) a maintaining nightmare. I would suggest to store only the names of the stacks to test and then fetch the appropriate repository information from the rosdistro repository (i.e. https://github.com/ros/rosdistro/blob/master/releases/groovy-devel.yaml).

I tried to ran the tests but was not sure what the number of expected tests is. Can you mention how many tests you can run per folder?

Thank you for your efforts!

tkruse:
Hi,

thanks for the comments. I updated -f my master (and hence the pull request).

Note I did not create the shell scripts in test/checks, those are by Troy, and I suspect they could be deleted as well.
Nevertheless I added the shell warning and asking for confirmation.

I modified the module structure, so now nosetessts should find all test modules itself regardless of whether it is called from within catkin or within catkin/test.

Also added PYTHONPATH explanation in test/README.rst to test stuff in catkin/python.

I managed to fix the quux project by doing what you said. I had the same tried before, but got confused because projects a,b,c,d all had the same problem, and I must have missed one.

In this pull request I now also added 2 tiny true unit tests just as examples, and a bugfix in topological order that I found when writing these true unit tests. (the other tests I call "integration tests", unit tests test just one single function ideally). I will add other such tests over time.

Note that in test/mock_resources/src there are two more projects, "ros_user" and "python_test_a", which were not used in previous tests, so right now they are useless, but i suspect somebody once wanted to write useful integration tests against those, so for now I kept them, and will see later whether I can write integration tests against them.

I also rebased again against 0.5.19, and all tests run, meaning:
$ nosetests
........
Ran 8 tests in 107.997s

OK

tkruse:
Regarding mentionings of fuerte: I think this is only mentioned in the checks/folder with the ugly stuff by Troy. If nobody uses that, I see no point in cleaning it up.

The same goes for most rosinstall files in test/checks. The usable tests currently only use test/network_tests/test.rosinstall.Not sure whether this one is hard to maintain.

Other than that I created catkin/rosinstalls/unstable.rosinstall and used that in the walkthrough.rst. I am not even sure whether that should be used in the future.

tkruse:
updated to latest master, fixed 2 bugs that I created

dirk-thomas:
I have merged the changes and did some minor modifications afterwards. Please update the testing code to conform to PEP8 in the near future.

PR #167, opened 2012-09-10
Symlink fix

tkruse:
(This is an attempt to create a pull request from a fork non-master branch to the main master branch.)

The logic for creating CMakeLists.txt still seems to be flawed. I wrote a unit test for init_workspace and noticed the docstring of init_workspace specified something different from the latest implementation.

So the attached patch changes the logic to:
when linking to toplevel.cmake within workspace use relative path.
Else use absolute path.

Also there was a problem if a user created a workspace with catkin somewhere, using an installed catkin version instead of the catkin version in his workspace. In that case, the init_workspace function would create an absolute link to the installed catkin, not a relative link to the workspace one.

[The more I see this, the less I think it is a good idea to ever allow catkin to live in the same workspace it is used in, as this will always cause non-future-proof hacks.]

Anyway, the patch also detects catkin in the dest workspace via the line in the patch:
workspaces = ['current:' + workspace_dir]

BTW: The whole CATKIN_WORKSPACES logic does not seem to be documented anywhere, I just assumed from the code that this is what is planned.

### DIFF
--- a/python/catkin/init_workspace.py
+++ b/python/catkin/init_workspace.py
@@ -25,28 +25,28 @@ def _symlink_toplevel_cmake(src, dst):
     return True
 
 
-#
-# Create a toplevel CMakeLists.txt in the root of a workspace.
-#
-# The toplevel.cmake file is looked up either in the catkin
-# workspaces contained in the CMAKE_PREFIX_PATH or relative to this
-# file.  Then it tries to create a symlink first and if that fails
-# copies the file.
-#
-# It installs ``manifest.xml`` to ``share/${PROJECT_NAME}``.
-#
-# .. note:: The symlink is absolute when catkin is found via the
-#   environment (since that indicates a different workspace and it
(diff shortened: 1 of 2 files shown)


tkruse:
hm, this did not quite work as expected on github

wjwwood:
Does it contain commits that you did not intend it to?  You don't have to close it, just update the branch and push -f it, the pull request will update accordingly.

tkruse:
Updated to latest master, also this fixes bugs/vulnerabilities in init_workspace around the .CATKIN_WORKSPACE file read

tkruse:
the "code style" commits should all change nothing to functionality and can be merged even if the last commit is rejected.

dirk-thomas:
I have cherry-picked most of the modifications and made some modifications to the init_workspace part. Please try to keep future pull request as small as possible. I.e. non-essential changes like renaming functions make it more difficult to review.

PR #169, opened 2012-09-10
Fix Python3 compatibility issue in catkin

davidhodo:
The execfile command used in the init.py files autogenerated by catkin does not work in Python 3. This pull request replaces that command with a version that works in both 2.7 and 3.


wjwwood:
This occurs in a few other cases, I am going to fix them and close this when I do.

PR #177, opened 2012-09-27
Test fix

tkruse:
This is a first batch to fix the catkin tests after the changes made towards a package.xml

It should fix all tests in test/unit_tests (run by extending PYTHONPATH and the nosetests test/unit-tests)

the caktin integration tests in test/network_tests should work once this bug gets fixed:
https://github.com/ros/common_msgs/issues/1
Remember that you may need to delete test/tmp if you have not run tests in a long time.

A test test/local_tests discovered a new minor bug in catkin that I cannot easily resolve, maybe someone else will see the solution immediately, so I do not start to search:

If you use an invalid CMakeLists.txt without 
project(foo),
like catkin/test/mock_resource/src-fail/noproject/CMakeLists.txt

then in catkin_package_xml the lines:
  if(NOT PROJECT_NAME)
    message(FATAL_ERROR "catkin_package_xml() PROJECT_NAME is not set. You must call project() before you can call catkin_package_xml().")
  endif()
should generate this error. However, the error does not appear, as 
PROJECT_NAME seems to have the value "Project"
a non-descript error happens as a consequence

That's what you get with
$ nosetests test/local-tests

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,5 +1,5 @@
 catkin
-========
+======
 
 Catkin is a collection of cmake macros and associated python code used
 to build some parts of `ROS <http://www.ros.org>`_
--- a/test/checks/test-nocatkin.rosinstall
+++ b/test/checks/test-nocatkin.rosinstall
@@ -1,7 +1,7 @@
-- git:
-    uri: 'git://github.com/willowgarage/catkin-debs.git'
-    local-name: catkin-debs
-    version: master
+# - git:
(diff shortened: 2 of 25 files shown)


tkruse:
Also this includes a workaround to 
https://github.com/ros/langs-dev/issues/1
not sure what langs-dev is and whether it will ever be required

tkruse:
updated to latest master, and undid changes to langs-dev in rosinstall, since for now this might be confusing. Can be fixed after ros/langs-dev#1

PR #188, opened 2012-10-07
added python sphinx autodoc generation

tkruse:
based on latest master

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -12,6 +12,7 @@
 # serve to show the default.
 
 import os
+import sys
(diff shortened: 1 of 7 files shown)


tkruse:
this depends on 
https://github.com/ros/catkin/pull/176
because else scripts run things when imported rather than remaining passive

tkruse:
sphinx style code comments are being added in several other of my pull requests, so if you feel like fixing docstrings to make sphinx autodoc look nicer, maybe please first merge with the other pull requests

PR #190, opened 2012-10-07
Redesign package_version.py

tkruse:
small changes to make package_version more robust, see commit message
also added docstrings and unit tests

Based on ros/catkin#176


dirk-thomas:
This pull request seem to have changes which are also in #188.

tkruse:
both state that they are based on ros/catkin#176, and those commits they had in common. Unless you meant something else, this should be fixed now with the rebase after 176 has been merged.

PR #191, opened 2012-10-07
Refactor catkin find

tkruse:
refactoring the code to remove duplicate code and avoid bad style.
added unit tests

based on ros/catkin#176

### DIFF
--- a/bin/catkin_find
+++ b/bin/catkin_find
@@ -3,8 +3,15 @@
 from __future__ import print_function
 import argparse
 import sys
+import os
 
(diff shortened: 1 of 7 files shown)


tkruse:
This pull request is more urgent, as I introduced a bug in ros/catkin#176 in bin/catkin_find, where I forgot to add

```
return args
```

in parse_args()

if the pull requests is more lengthy to review, you might consider just fixing that in the meantime.

tkruse:
note that the functionality of workspaces might better go into catkin_pkg instead

PR #194, opened 2012-10-07
Refactor and minor fix run_tests

tkruse:
added docstrings, unit tests, better error message.
Also changed arg --results to be positional argument instead, as it was mandatory


tkruse:
rebased to latest master and fixed bug as detected in review.

PR #199, opened 2012-10-08
More unit tests

tkruse:
- remove topological test after module moved to catkin_pkg
- added test for regression in #190


tkruse:
rebased to latest master, fix failing tests due to invalid package.xml after catkin_pkg update

PR #202, opened 2012-10-09
Catkin fuerte integration tests repaired

tkruse:
This is a backport of the integration tests that I restructured for groovy to fuerte. After the pull requests, nosetests should run through without fail, testing catkin_fuerte against core fuerte branches of ROS. This allows safer backporting of other features to fuerte (mainly catkin_init_workspace next).
I renamed some files from 'unstable' to 'test', so some backports will not be possible by mere cherry-picking, but hopefully the different names prevent confusion.

Issue #170, opened 2012-09-11
Backport catkin_init_workspace to fuerte catkin


tkruse:
For the distro-agnostic tools like rosinstall, it would be very helpful if the different catkin branches for all distros provided a consistent API for how to setup the workspace.

The main challenge for rosinstall is to know what to do for a catkin workspace and where to find the correct catkin instance to use, when there can be multiple instances in the given workspace and a system-wide installation.

rosinstall --catkin could be rewritten to call catkin_init_workspace, and if that fails (because there is no catkin system installation), it could take a path/to/catkin argument for the user to choose. ideally all catkin branches would have a fixed relative path to the catkin_init_workspace script that rosinstall could then invoke.

tkruse:
duplicate of ros/catkin#206

PR #206, opened 2012-10-10
Init workspace for fuerte

tkruse:
fix for #170, adds catkin_init_workspace to fuerte, for consistent workspace setup in fuerte and groovy.

Copied the file without changes, not sure whether all search cases make sense in fuerte, but this makes it easier to backport future patches.

This pull request is based on ros/catkin#202,
so commits belonging to ros/catkin#202 may show until the other pull request was merged. Only the last commit is relevant for this pull request.

### DIFF
--- a/bin/catkin_init_workspace
+++ b/bin/catkin_init_workspace
@@ -1,0 +1,34 @@
+#!/usr/bin/env python
+
+from __future__ import print_function
+import argparse
+import os
+import sys
+
+try:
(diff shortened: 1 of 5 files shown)


tkruse:
On second though, just added the commit to fuerte-devel, since there is no strong dependency to ros/catkin#202, and Dirk does not care about linear commit history.
Just note that the unit tests supposed to run for this pull requests are in test/unit-tests. All other tests may still fail.

PR #209, opened 2012-10-10
refactored and unit-test parse_package_xml.py

tkruse:


dirk-thomas:
I merged the changes and fixed the usage of the Mock() constructor (see related commit). Please use that initialization for mocks for future patches.

tkruse:
Sorry, did the usage of the Mock() contructor with arguments not work for
you, or what is the problem with it?

dirk-thomas:
The mock constructor does not support kwargs - at least not in the version shipped with precise. I fixed that already for another pull request recently (can't find which one).

tkruse:
Ok, I am on oneiric, and it works, that's weird. Sorry about it, if i see
it again I will fix it.

tkruse:
I want to find out what happens. Can you run this please:

> > > import mock
> > > mock.**version**
> > > '1.0b1'

dirk-thomas:
Precise ships with mock version 0.7.2
http://packages.ubuntu.com/precise/python-mock

Issue #203, opened 2012-10-09
catkin_find does not correctly find files in src


tfoote:
I did a standalone install of std_msgs with CMAKE_PREFIX_PATH=/opt/ros/groovy

With the buildspace setup.bash sourced I got the following incorrect fall through to the installed version:

```
buildspace$ catkin_find std_msgs msg/Bool.msg
/opt/ros/groovy/share/std_msgs/msg/Bool.msg
```

After apt-get remove ros-groovy-std-msgs it returns nothing

```
buildspace$ catkin_find --share std_msgs msg/Bool.msg

```

And it does not list the source directory w/o the 2nd argument. 

```
buildspace$ catkin_find --share std_msgs
/tmp/std_msgs/build/buildspace/share/std_msgs
```

tkruse:
Providing a fix in #212, but the current solution using catkin_pkg.packages.find_packages(basepath) over all known catkin workspace source locations is highly inefficient.

This bug shows the general problem with having a name tag in REP127, and a flexible folder name.

Since we went back to allowing hierarchical workspaces, we now have to crawl all source folders for package.xmls and then parse all these package.xmls to find one which has the correct name. This can be tweaked a lot with a single workspace knowing that mostly folder names will be equal to package names, but with n workspaces, all packages of overlaying workspaces need to have their package.xml parsed before proceeding to the next workspace.

tkruse:
This seems to be almost fixed with 41c36787044b4ae9, except that the option first_matching_workspace_only needs to be used as default, with maybe a command line option to also show overlayed paths

dirk-thomas:
catkin_find is used in several cases. What seems to be a reasonable default in this case might not be for other. So I would keep the option as it currently is.

tkruse:
Tully raised this, so maybe Tully wants to comment on the desired behavior of
$ catkin_find std_msgs msg/Bool.msg

in the overlay case. In any case the behavior should be documented in the help, e.g. that by default even overlayed paths appear.

Currently the help is also misleading:

```
parser.add_argument('--first-only', action='store_true', help='Flag if only the results from the first matching workspace should be returned (default is all)')
 find_in_workspaces(... first_match_only=args.first_only...)
```

According to the help, the param passed to find_in_workspaces should be first_matching_workspace_only, not first_match_only. And there is no option to get only overlayed paths currently.

dirk-thomas:
The question is what should be the default behavior?
We have three different use cases imho:
[...]

PR #212, opened 2012-10-10
Fix find in ws

tkruse:
fix for #203


tkruse:
The pull request fixes two main bugs: the code in find_in_workspace still assumed a flat workspace source folder with every subfolder being a package. Also the code did not at all attempt to provide any overlay logic.
This bug indicates to me that we need in catkin_pkg an API to find a single package by name efficiently.

tkruse:
to be fixed in a different way

PR #213, opened 2012-10-11
Reexport plugin declarations into the manifest.xml required for plugins

tfoote:
This will reexport plugin exports from the package.xml into the generated manifest.xml.  And also add the required rospack style dependency onto the package so that rospack plugins will discover it.


dirk-thomas:
Removed generation of manifest files in de420fa9d8da54e0d7fd0acf48505af34b7b1c76

PR #222, opened 2012-10-15
Test and refactor setup py

tkruse:
added refactor of workspace as well, as it is related.
- replaced long dict check in workspace.py continuations with equivalent continuation using get with default.
  This may in some cases do an unecessary ''.split(), but I think it is still preferrable code.
  then refactored functions in setup.py:
- moved in get_workspaces() (Note it would be much better to use catkin/workspace.py, but I am not sure about PYTHONPATH issues to consider)
- renamed a lot of vars to make the code readable
- replaced long continuations with shorter ones using dict.get() with default
- removed / prefix in subfolder argument if any, and use os.path.join instead

Note that I would also like to refactor the CLI api of setup.py.
Instead of
setup.py --prefix --name xxx --value yyy
it would be nicer to use
setup.py prepend xxx --values yyy
setup.py reversed-workspace --exclude yyy
setup.py remove xxx --subfolder yyy

Also it would be much nicer if the ..._dir varssetup.sh did not contain a leading '/'

And it would be better that instead of --prefix returning the prefix we would use --prepend returning the whole new env var value (makes both setup.py and setup.sh cleaner)

But since setup.sh is not unit tested yet, I did not do these changes, and will wait for you to comment before trying them.

### DIFF
--- a/cmake/templates/script.py.in
+++ b/cmake/templates/script.py.in
@@ -1,3 +1,5 @@
 #!/usr/bin/env python
+# creates a relay to a python script source file, acting as that file.
+# The purpose is that of a symlink
 with open("@PYTHON_SCRIPT@", 'r') as fh:
     exec(fh.read())
--- a/cmake/templates/setup.py
+++ b/cmake/templates/setup.py
@@ -5,6 +5,7 @@ import argparse
 import os
 import sys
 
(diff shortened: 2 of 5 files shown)


tkruse:
Note I was also thinking of moving get_workspaces to catkin_pkg, which might make it easier for setup.py to depend on it

PR #224, opened 2012-10-15
print catkin version for easier debugging

tkruse:
prints what version of catkin runs, I think all.cmake works for all use-cases.
Useful for helping other people with catkin-related problems, makes it easier for them to report errors with version.


dirk-thomas:
You can not rely on being able to find_package(catkin) at that point.

tkruse:
changed to check quietly and print "from source" else, which is not optimal
but still better than nothing.

tkruse:
fixed in bc72f6f68966e

PR #236, opened 2012-11-01
fix unit test failures (e.g. after fixing of #194)

tkruse:
patch just changes argument order in unit test to match script order, and some asserts that are outdated


tkruse:
Now also fixes a bad assert in a different test

tkruse:
Now also fixes bad tagnames in mock resources

PR #254, opened 2012-11-12
Fix 2 minor bugs in catkin_python_setup

tkruse:
1st bug is that package_dir is required, but not required for distutils.

2nd bug is that catkin_python_setup.cmake calls catkin_package_xml itself. This can cause plenty of unsuitable error messages. E.g. if call to project() is missing, ${PROJECT_NAME} will still be default value "Project". And consecutive errors.

Also note that there are several other such flaws around genmsg. E.g. in genmsg-extras.cmake.em, PROJECT_NAME is used without checking it has been set.

Also in catkin_python_setup.cmake, the variable ${PROJECT_NAME}_GENERATE_MESSAGES is used, but this is an implicit dependency on genmsg and should be avoided. BUt I do not understand enough about the problems involved to fix that myself. I can raise an issue though.

### DIFF
--- a/cmake/catkin_python_setup.cmake
+++ b/cmake/catkin_python_setup.cmake
@@ -19,6 +19,12 @@ function(catkin_python_setup)
     message(FATAL_ERROR "catkin_python_setup() called with unused arguments: ${ARGN}")
   endif()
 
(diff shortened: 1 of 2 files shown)


dirk-thomas:
I merged the first changeset.

Issue #271, opened 2012-12-03
catkin should warn for unsupported distutils / setuptools / distribute keyword usage


tkruse:
the catkin_python_setup.py macro tries to recreate a develspace that works smilar to the install space, but by relaying source files. This functionality breaks if uses use advanced parameters of the respective setup() functions, like:
- py_modules
- data_files
- any extention module features fromdistutils
- zip-safe
- entry_points
- include_package_data
- exclude_package_data
- setup_requires
- namespace_packages
- use_2to3

catkin should warn the user that those features will not be available in devel space.

See
http://packages.python.org/distribute/setuptools.html#new-and-changed-setup-keywords
http://docs.python.org/2/distutils/setupscript.html

Issue #270, opened 2012-12-03
interrogate_setup_dot_py monkey_patches too much


tkruse:
in interrogate_setup_dot_py.py:

```
  sys.modules['setuptools'] = dummy
  sys.modules['distutils.core'] = dummy
```

This causes the basic setuptools use case to break:
http://packages.python.org/distribute/setuptools.html#basic-use

```
from setuptools import setup, find_packages
setup(
    name = "HelloWorld",
    version = "0.1",
    packages = find_packages(),
)
```

because setuptools.find_package does not exist, then. Suggest only patching setuptool.setup and distutils.core.setup functions, instead of the whole module. I can create a pull request if so desired.

tkruse:
Else documentation should be extended, and maybe the error message improved:

```
Traceback (most recent call last):
  File "/home/kruset/groovy_underlay/src/catkin/cmake/interrogate_setup_dot_py.py", line 167, in <module>
    main()
  File "/home/kruset/groovy_underlay/src/catkin/cmake/interrogate_setup_dot_py.py", line 163, in main
     exec(fh.read())
  File "<string>", line 7, in <module>
AttributeError: Dummy instance has no attribute 'find_packages'
```

dirk-thomas:
Sure, a pull request for that would be great.

PR #272, opened 2012-12-04
patches for #270 and #271

tkruse:


tkruse:
Note I also thought about using a whitelist instead of a blacklist for #271, containing:
'version','package_dir','packages','scripts','name','url','description','long_description','license','keywords','maintainer','maintainer_email','author','author_email','keywords','classifiers','requires','install_requires'

Not sure which approach is better, but I thought it might be better to only war for things we know that would cause breakage in devel space.

PR #275, opened 2012-12-07
improvements to Groovy docs

tkruse:
fix wrong sequence in sketch.rst
adapt setup.py util function call given change in catkin_pkg
Improves catkin_python_setup macro text


tkruse:
Rebased, added fixes to catkin_package documentation in user guide, missing CATKIN_DEPENDS. Else see commit messages

PR #277, opened 2012-12-12
Ignore build type cmake

wjwwood:
This pull request makes catkin ignore packages with the snippet:

```
<export>
  <build_type>cmake</build_type>
</export>
```

in their `package.xml` file.  This allows us to have catkin ignore packages like `pcl` and `opencv2` which have `package.xml`'s but are not build able in a catkin workspace.

Additionally there were some changes to the output of catkin to make things clearer and corresponding changes to the colorization provided by `catkin_make`.

Screen shots of the new output:

Screen Shot 2012-12-11 at 11 27 07 PM

In the topological sort, packages are tagged with metapackage and non-catkin where appropriate.

Screen Shot 2012-12-11 at 11 27 26 PM

The `+++ ...` lines now always output the package _name_ and then the call to `add_subdirectory` is shown with the relative _path_ to the package.


wjwwood:
Fixed, removed the redundant line.

PR #278, opened 2012-12-12
add warnings for deprecated packages

dirk-thomas:
Add warnings when building a deprecated package in a workspace and when find_package()-ing it (using a deprecated package).

### DIFF
--- a/cmake/parse_package_xml.py
+++ b/cmake/parse_package_xml.py
@@ -26,7 +26,7 @@ def _get_output(package):
     values['RUN_DEPENDS'] = ' '.join(['"%s"' % str(d) for d in package.run_depends])
 
     deprecated = [e.content for e in package.exports if e.tagname == 'deprecated']
-    values['DEPRECATED'] = '"%s"' % ((deprecated[0] if deprecated else 'TRUE') if deprecated else '')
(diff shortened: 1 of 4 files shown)


cmake/parse_package_xml.py
```diff
@@ -25,6 +25,9 @@ def _get_output(package):
     values['BUILD_DEPENDS'] = ' '.join(['"%s"' % str(d) for d in package.build_depends])
     values['RUN_DEPENDS'] = ' '.join(['"%s"' % str(d) for d in package.run_depends])
 
+    deprecated = [e.content for e in package.exports if e.tagname == 'deprecated']
+    values['DEPRECATED'] = '"%s"' % ((deprecated[0] if deprecated else 'TRUE') if deprecated else '')
```
wjwwood: Isn't the second `if deprecated else ''` redundant?

wjwwood: Should that be:

```
values['DEPRECATED'] = '"%s"' % ((deprecated[0] if deprecated[0] else 'TRUE') if deprecated else '')
```

dirk-thomas: exactly - updated.

wjwwood:
Looks good to me.

PR #284, opened 2012-12-18
Removed redundant/obsolete RST documentation

mirzashah:
Removed docs that are available on ROS wiki and are more up to date.


dirk-thomas:
Looks like the user_guide is still referenced from at least index.rst. Try running `make html` in the doc folder to confirm that it builds.

mirzashah:
There are still some residual files in user_guide. I'm not sure if they're obsolete and/or still needed.

tkruse:
I'd like to have introduction.rst (at least vital parts of it), supposed.rst and package.rst back in the catkin rst docs. It is true that they are somewhat redundant with the wiki in content, but not in purpose. If anyone downloads catkin or looks at catkin on github, he should get all vital information from the rst docs,without having to look at a ROS wiki. The introduction and package rst files provide that.

The purpose of the rst files should be to document the code provided with a package.
The purpose of the wiki is for ROS users to flesh out minimal information from the .rst files and add extra tipps images, explanations and whatnot. Some duplication cannot be prevented, the only way to deal with that is to minimize the text in the rst files. Not to get rid of the information.

dirk-thomas:
The .rst files should only contain information which is not available on the Wiki. Duplicating any information is not an option. If important content was lost when removing these files (which is not in the Wiki) please feel free to create a pull request reintroducing these information.

PR #285, opened 2012-12-19
Added catkin_make_isolated.

wjwwood:
catkin_make_isolated builds all packages in a
workspace in separate build spaces. By default it
uses separate devel spaces for each package as
well. It can also build plain cmake packages
from the workspace, installing them into the
install space. The --install option will build
and install each catkin and plain cmake package
into the install space.

The recommended use of this tool is for the
source install instructions and for debugging
build problems.

Visuals of the output:

Screen Shot 2012-12-18 at 4 45 42 PM

Screen Shot 2012-12-18 at 4 46 06 PM


wjwwood:
I have made adjustments and rebased (squashed) the commits to master.

tfoote:
@wjwwood can we close this then?

wjwwood:
@tfoote I need to talk to @dirk-thomas about it first probably.

PR #286, opened 2012-12-20
Handling header files generated by configure_file

stonier:
Reasonably often, cmake is used to pregenerate .h/.hpp files - 
usually to define macros describing the platform (results of 
cmake tests etc.). The question is, where to store the configured 
files. 

Converting and storing them in the source directory next to their
respective ___.hpp.in files is unsafe. If you're doing multiple
parallel builds (often happening when cross compiling), then you 
end up with builds including a header that was generated for one 
of the other parallel builds.

A better place that is usually used is CMAKE_BINARY_DIR, however
that doesn't work with catkin. There is no way that I know of 
to pass along the location of that header to subsequent dependent 
packages - catkin only includes the source directories.

One place that would seem to fit might be CATKIN_DEVEL_PREFIX/include.
It would be easy to add this directory to an include_directories() or
add it to catkin_INCLUDE_DIRS like I did here to allow subsequent
packages to find those headers.


dirk-thomas:
This should not be added automatically by catkin.

If your package places files in any location (could be DEVEL/include or even BUILD/somewhere) you should:
- `include(directories(..)` that path explicitly, so that it is found by the compiler
- pass it to `catkin_package(INCLUDE_DIRS ...)`, so that it is passed to downstream packages when not being installed
- install the generated files to the standard location (CATKIN_PACKAGE_INCLUDE_DESTINATION), so that is part of the debian package

stonier:
Awesome, working.

Totally forgot about INCLUDE_DIRS exporting directories for the devel space. Even just found comments about that in my packages.

Sorry for the hassle.

PR #288, opened 2013-01-01
explain which candidate failed

tkruse:
ran into this with a package that I had installed as non-catkin package (console_bridge). Not sure whether a better error message can be created other than using _LIBRARIES

### DIFF
--- a/cmake/catkin_package.cmake
+++ b/cmake/catkin_package.cmake
@@ -149,7 +149,7 @@ function(_catkin_package)
     if(${depend_name}_FOUND)
       # verify that these packages are really catkin packages
       if(NOT ${depend_name}_FOUND_CATKIN_PROJECT)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I don't see how the libraries are related to the error message at all? The name of the package is stated in the error message - what information do you think are lacking from the error message?

tkruse:
If the user hs multiple candidates in his CMAKE_PREFIX_PATH, and a candidate for the given package name is found in a different location than the user expects, the error message does not tell the user so. so the user will look in the wrong place trying to fix things.

dirk-thomas:
Just outputting the LIBRARIES is not a good approach. I.e. the package might not even have libraries?

If the package is found in config mode the path of the found config file is set by CMake. This is imho the only reliable and useful information which can be added to the error message.

PR #291, opened 2013-01-04
Restore some user guide pages in a minimal state

tkruse:
Restored some vital information about catkin in order to have a complete and consistent user guide.

I shortened the original contents removing much that seemed non-essential.

This partly reverts commit 0c866730461ebd063cbfc4744925580e63d1661a.
"Removed redundant/obsolete user documentation that is now found on ROS wiki"

### DIFF
--- a/doc/user_guide/package.rst
+++ b/doc/user_guide/package.rst
@@ -1,0 +1,151 @@
+Packages
+========
+
+.. highlight:: catkin-sh
+
+A catkin package is a folder with a valid ``package.xml`` and a
+``CMakeLists.txt`` file (unless it is a meta-package).
+
+.. _package.xml:
+
+package.xml
+^^^^^^^^^^^
+
+This is used by catkin's dependency-sorting to determine what order to
+traverse CMake projects in (when in a catkin workspace) and by
+packaging utilities when making Debian packages.
+
+The specification for package.xml is in `REP 127 <http://www.ros.org/reps/rep-0127.html>`_.
+
+A small example is::
+
+  <?xml version="1.0"?>
+  <package>
+    <name>example</name>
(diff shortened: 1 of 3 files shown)


dirk-thomas:
This content looks still highly redundant to the documentation we have in the Wiki. Why should that be in the .rst files and not just be a plain link to the corresponding Wiki page?

tkruse:
Some reasons: The documentation in the rst files changes over time with the
source code.It is the specificatino of the source code. In 2 years, the
code may have changed a lot, so will the wiki. But if a user still uses the
current version of catkin, the docs will still be accurate.

Also the wiki is under no supervision, anyone can write anything there, the
rst docs are the reviewed design authority.

A wiki link would be better than nothing at all, but I see no reason why a
software package on github should be dependent in its documentation on a
ROS wiki.

dirk-thomas:
Simple answer: no duplication.
It is a trade off for sure. But duplication is considered being worse than people running outdated versions and only finding up-to-date docs on the Wiki.

tfoote:
I understand that it would be great if everything was freestanding and documented nicely in small units.  However, as the ROS wiki is the primary portal for both developers and users for docs that's where we should keep the docs.  As you mentioned over time things change significanly and if we have a secondary documentation area, it will likely be overlooked by most developers and users.  And the poor user who shows up down the line and runs into the overlooked documentation will be signifincantly worse off than if the secondary documentation did not exist, as they would find the primary documentation by continuing their search.  As such I think we should stick with the ROS wiki as primary documentation and we could put a link into the code, but shouldn't duplicate documentation if at all possible.  

If there are more technical things which are more developer focused and are not for the user facing documentation those would be good to put into the documentation here.

PR #311, opened 2013-01-14
Merged run_commands in catkin.builder module

wjwwood:


wjwwood:
This is basically @tkruse's patch but with some stuff I have been meaning to do as well.

PR #310, opened 2013-01-15
catkin_make doesn't handle colorizing of unicode text properly

wjwwood:
We're building a couple third party packages using catkin, and wrapping their build systems. Some of them output non-ascii characters, which breaks the colorizing in catkin:

```
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2018' in position 38: ordinal not in range(128)
<caktin_make> color formatting problem: 'ascii' codec can't encode character u'\u2018' in position 38: ordinal not in range(128)
  File "/opt/ros/groovy/lib/python2.7/dist-packages/catkin/builder.py", line 132, in run_command_colorized
    out.write(unicode(colorize_line(line)))
  File "/opt/ros/groovy/lib/python2.7/dist-packages/catkin/builder.py", line 58, in colorize_line
    cline = str(sanitize(line))
```

### DIFF
--- a/python/catkin/builder.py
+++ b/python/catkin/builder.py
@@ -56,7 +56,7 @@ def cprint(msg, end=None):
 
 
 def colorize_line(line):
-    cline = str(sanitize(line))
+    cline = sanitize(line)
     cline = cline.replace(
         '-- ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~',
         '-- @{pf}~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~@|'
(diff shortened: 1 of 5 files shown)


wjwwood:
I have reproduced the bug once, but I am having trouble reproducing it multiple times, even when clearing the downloaded sources from the source tree (bad mojo btw).

I am looking into the build code which outputs the symbols.

wjwwood:
Note, I had to disable an out of date test.  I can try and update it if you want @dirk-thomas

jonbinney:
I looked up the unicode value \u2018 it is complaining about. It is a "Unicode Character 'LEFT SINGLE QUOTATION MARK' (U+2018)"

In general it looks like the sanitize() function in terminal_color.py only handles a small subset of non-ascii characters. If the output of that function is really supposed to be  valid ascii, might need to add a final:

def removeNonAscii(s): return "".join(filter(lambda x: ord(x)<128, s))

As suggested here: http://stackoverflow.com/questions/1342000/how-to-replace-non-ascii-characters-in-string

wjwwood:
Reopening...

Also I have converted this to a pull request with a fix.

<str>.decode('utf8', 'replace') will replace non-ascii values with ?

I also now have a test which I could reproduce it with.

wjwwood:
Also in my terminal the funny ' came out ok after colorization.

PR #323, opened 2013-01-19
Support inverted color scheme

wjwwood:
catkin_make colors the output, and uses bright foreground colors over custom background colors. That means if the user had a white background, he will not see certain text.

Attaching file

### DIFF
--- a/python/catkin/builder.py
+++ b/python/catkin/builder.py
@@ -154,7 +154,7 @@ def run_command(cmd, cwd, quiet=False, colorize=False):
         raise subprocess.CalledProcessError(proc.returncode, ' '.join(cmd))
     return out.getvalue() if quiet else ''
 
-blue_arrow = '@!@{bf}==>@{wf}'
+blue_arrow = '@!@{bf}==>@|@!'
 
 
 def _check_build_dir(name, workspace, buildspace):


tkruse:
console

wjwwood:
So from a few of my terminal styles:

Screen Shot 2013-01-18 at 3 34 42 PM

In the above theme you can see that it makes the bold white slightly grey so it shows up better.

Screen Shot 2013-01-18 at 3 34 28 PM

I would say that your theme should make sure that all colors are visible against the background.  The theme dictates the background and foreground colors.

That being said I am going to try and make things better by removing explicit colorization from some of the text, which should make the important information more readable.

wjwwood:
Ok, I was explicitly using white foreground at some point in there.

wjwwood:
Baseline:

Screen Shot 2013-01-18 at 3 48 46 PM

Without color:

Screen Shot 2013-01-18 at 3 48 59 PM

With bold only:

Screen Shot 2013-01-18 at 3 49 32 PM

I prefer the bold only, thoughts?

wjwwood:
@tkruse can you get this branch and try it out on your machine?

tkruse:
works for me

PR #322, opened 2013-01-24
catkin_make_isolated should allow building individual packages

wjwwood:
Currently going into build_isolated/packagename and running make or cmake does not work, nor do I see any other easy way of running that build in isolation, I guess the env.sh file is required, but it should be easier to do this, IMO.

Would be less urgent if catkin_make_isolated had rosbuild parallelity.


wjwwood:
@dirk-thomas and @tkruse for review

PR #327, opened 2013-01-25
catkin_make_isolated should provide setup files

wjwwood:
setup files with similar functionality as for the non-isolated case should be available for an isolated devel space. I think we discussed this, just tracking the issue.


wjwwood:
@dirk-thomas and @tkruse for review.

wjwwood:
Updated.

PR #355, opened 2013-02-06
Convert windows path to cmake path

stonier:


dirk-thomas:
Can you mention in which use case this was an issue?

stonier:
It failed topological ordering as soon as it hit the ros stack. Incoming path strings looked like "ros\core\mk" which would cause CMake to abort.

stonier:
This was something that got lost in fuerte -> groovy. You can see we had it implemented in

https://github.com/ros/catkin/blob/832c1312cccc79ac7325bfe13dc22b87a5389a93/cmake/em/topologically_traverse.cmake.em

dirk-thomas:
Weird. Actually the topological order worked fine for me yesterday on Windows. I will look into that soonish.

stonier:
Closing this and re-opening from the pull request from my fork.

PR #359, opened 2013-02-11
Minor windows fixes

stonier:
Some trivial fixes...bigger fixes are pending and probably need some discussion.

21cfef6c45e6ab2220180e175cd3f92b7cc5d805: it converts windows paths (e.g. ros\core\mk) into *nix paths so that cmake doesn't fall over when it sees backslashes. Further discussion is at #355

c691275f56cb519ecabd1ab04d416c1e99d0dcad: haven't used it here, but will in a coming patch. Useful to have there at any rate.

10fb4ead188114f5156e6819c06e3755bf27aeeb: comments in the commit.

### DIFF
--- a/cmake/all.cmake
+++ b/cmake/all.cmake
@@ -114,6 +114,7 @@ foreach(filename
     parse_arguments
     safe_execute_process
     stamp
+	platform/cross_platform_utilities
     platform/lsb
     platform/ubuntu
     platform/windows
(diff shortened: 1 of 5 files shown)


dirk-thomas:
I just applied the first and third patch to go on with local testing.

dirk-thomas:
The second patch should be part of the next pull request which requires the variable introduced in the changes.

stonier:
Thanks dirk and yes, I should have unwound the second one before pull requesting. I'll be more strict on it in future.

Thanks.

PR #378, opened 2013-03-10
Handling dependencies that are imported libraries

stonier:
Carry over from our email discussion - better having it here.

**Problem**: sometimes dependencies are provided as import targets. This causes a problem when creating a catkin package's Find... cmake module as it stores the target name which is often different to the actual library name. Packages using this then fail to find the required libraries.

**About Imported Libraries**

This is a convenience as it lets you use them like a real target. Refer to the [cmake documentation](http://cmake.org/cmake/help/v2.8.8/cmake.html#command:add_library) and [wiki tutorials](http://www.cmake.org/Wiki/CMake/Tutorials/Exporting_and_Importing_Targets) for more information. Boost's cmake modular build creates files in `lib/Boost.cmake`, `lib/Boost-debug.cmake`,  `boost-1.47.0/cmake/BoostConfig.cmake` and `boost-1.47.0/cmake/BoostConfigVersion.cmake` as an example.

**Use Case**

We're using cmake-boost for windows. Note that this isn't specifically a windows problem. It would apply to any xxx-config.cmake  files which were written setting found libraries as imported targets.

The above diff handles the problem for windows. I'm not sure about handling the case where you have multiple imported configurations though and how you should choose the configuration you require. This works for boost as each configuration (DEBUG, RELEASE) has only one imported configuration. I don't have any other examples to test on.

**This Approach**

Simply digs into the imported library property information and dumps the actual library name into the `xxxConfig.cmake` for the package.

**Other Approaches**

I did initially think of recreating the imported targets in the xxxConfig.cmake modules, but this will fail as soon as another package tries to `find_package(Boost...)` again as suddenly there will be two `add_library(xxx IMPORTED)` calls.


dirk-thomas:
The patch looks great - thanks. I will commit a slightly modified version without tabs and trailing whitespaces.

stonier:
Sorry - configuring my editor now to lose the tabs.

PR #381, opened 2013-03-12
Better nosetest build folder naming

stonier:
Windows failing to build genmsg which uses 

```
catkin_add_nosetests(${PROJECT_SOURCE_DIR})
```

and it falls all over the `C:\` prefix to the path. This led to finding this bit of code:
- Was it actually intending to strip the source directory and not the binary directory? I can't see the reason you'd want to strip the binary directory and stripping the source directory makes it more convenient for introspection.
- This regex replace code is quite a bit simpler for removing prefixes
- Provides a name when ends up with the empty string (could maybe choose a better name?)
- Also replaces colons with dots to avoid problem windows path characters.

### DIFF
--- a/cmake/test/nosetests.cmake
+++ b/cmake/test/nosetests.cmake
@@ -55,21 +55,18 @@ function(catkin_add_nosetests path)
     set(_covarg " --with-coverage")
   endif()
 
-  # strip PROJECT_BINARY_DIR from output_file_name
+  # strip PROJECT_SOURCE_DIR from output_file_name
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Can you please test the branch cleaner_nosetest_target_names, verify that it does what you wanted to achieve and report back?

stonier:
:+1: - branch changes work fine.

dirk-thomas:
New pull request pending: #382

PR #383, opened 2013-03-13
Avoid creating nmake inference rule problems

stonier:
**Problem**

rostest's `add_ros_test` calls on windows would run through the cmake ok, but when hitting nmake, it would stop short with the error message:

```
makefile(1035) : fatal error U1086: inference rule cannot have dependents
Stop.
```

In the `Makefile`, you would find a target like:

```
.run_tests_rostest_rostest_test_hztest0.test: cmake_check_build_system
    $(MAKE) -f CMakeFiles\Makefile2 /nologo -$(MAKEFLAGS) .run_tests_rostest_rostest_test_hztest0.test
```

Nmake would read this like an inference rule (aka `.c.obj:`) and abort when it saw the dependency (http://msdn.microsoft.com/en-us/library/d8k6a82k.aspx).

**Solution**

Is there a special reason the _hidden_ targets are using a dot? I've used underscore here.

### DIFF
--- a/cmake/test/tests.cmake
+++ b/cmake/test/tests.cmake
@@ -43,17 +43,17 @@ function(catkin_run_tests_target type name xunit_filename)
   if(NOT TARGET run_tests_${PROJECT_NAME})
     add_custom_target(run_tests_${PROJECT_NAME})
     # create hidden meta target which depends on hidden test targets which depend on clean_test_results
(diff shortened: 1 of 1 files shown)


dirk-thomas:
No specific reason why we selected `.`. `_` should do the same job of hiding the target name on auto completion and work well with nmake.

PR #385, opened 2013-03-13
modify extra file handling to reference them relatively if possible to make the generated code relocatable (#369)

dirk-thomas:


dirk-thomas:
Confirmed to work in #369.

@tfoote @wjwwood Please review and merge.

wjwwood:
+1

PR #394, opened 2013-03-17
Remove redundant install-scripts option.

stonier:
The `--install-scripts` option gets added to `SETUPTOOLS_ARG_EXTRA` and doubled up in `cmake/templates/python_distutils_install.bat.in` and `cmake/templates/python_distutils_install.sh.in`.

We can remove it here, or in the `python_distutils_install.bat.in` files. I did it here in `python.cmake` since I was not sure if it was necessary with `--install-layout=deb`.


dirk-thomas:
Seems to addressed by #393 already?

stonier:
Indeed, thanks.

PR #402, opened 2013-03-27
catkin_make* should allow for `MAKEFLAGS`

wjwwood:
Currently `caktin_make*` programs always give job  arguments to `make`, even if none are passed to them directly, e.g. `catkin_make` -> `make -j8 -l8` and `catkin_make -j1` -> `make -j1 -l1`.

This prevents the Make environment variable `MAKEFLAGS` from working with `catkin_make*`. I propose a new logic for handling `-j`, `--jobs`, `-l`, `--load-average` arguments:

catkin_make_jobs

This will allow users to override the make job options with direct arguments to `catkin_make*`, `MAKEFLAGS`, or `ROS_PARALLEL_JOBS`.

_EDIT:_ Removed references to `CATKIN_PARALLEL_JOBS` because it is overshadowed by `MAKEFLAGS` and direct arguments, and I did not want it to have the same weird specification as `ROS_PARALLEL_JOBS` where it has flags for make not just a number.

### DIFF
--- a/bin/catkin_make
+++ b/bin/catkin_make
@@ -156,8 +156,8 @@ def _parse_args(args=sys.argv[1:]):
         'If no -j/-l arguments are given, then the MAKEFLAGS environment variable is searched for -j/-l flags. '
         'If found then no -j/-l flags are passed to make explicitly (as not to override the MAKEFLAGS). '
         'If MAKEFLAGS is not set then the environement variable ROS_PARALLEL_JOBS is used. '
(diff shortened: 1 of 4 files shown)


wjwwood:
Ok, now @dirk-thomas for review. I accidentally messed up pull request #400.

dirk-thomas:
The graph lacks an arrow from "Does MAKEFLAGS have any job flags in it?" to "Is ROS_PARALLEL_JOBS set?" in case of "No".

dirk-thomas:
+1

wjwwood:
@dirk-thomas updated, review and merge please

PR #403, opened 2013-04-01
Corrects catkin_make

chadrockey:
Better user output for catkin, much more intuitive - compiling is like assembling a multi-layer cake.


dirk-thomas:
Duplicate - already implemented since a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1a1

PR #405, opened 2013-04-02
Remove old command, which is annoying when tab completing

wjwwood:
This command is annoying when you are trying to tab complete. It is also misleading as I doubt it works anymore, users should not see it and try to use it.

Is it being used or can we get rid of this?


dirk-thomas:
It would be great if catkin would provide such a command again so that people can try build their own deb and verify if it contains what it needs.
Anyway the current file is indeed useless and can be removed.

wjwwood:
https://github.com/ros-infrastructure/bloom/issues/121

PR #404, opened 2013-04-05
catkin_prepare_release should validate metapackages

wjwwood:
Specifically that they have valid `CMakeLists.txt`. A lot of people are going to run into this problem while releasing with bloom. It would be nice to have this tool check.

It could output:

```
cat <<EOF
Metapackage <metapackage_name> does not have a valid CMakeLists.txt, therefore it does not conform to REP-0127, see: http://ros.org/reps/rep-0127.html#metapackage

The CMakeLists.txt should look like this:
<CMakeLists.txt that was expected>
EOF
```

Instead of its normal command output, which would be safe for people doing something like this:

```
`catkin_prepare_release`
```

### DIFF
--- a/bin/catkin_prepare_release
+++ b/bin/catkin_prepare_release
@@ -6,6 +6,7 @@ import os
 import sys
 
 from catkin_pkg.packages import find_packages, verify_equal_package_versions
+from catkin_pkg import metapackage
 
 # find the import relatively if available to work before installing catkin or overlaying installed version
 if os.path.exists(os.path.join(os.path.dirname(__file__), 'CMakeLists.txt')):
@@ -68,6 +69,14 @@ def main():
     if not packages:
         print('No packages found', file=sys.stderr)
         sys.exit(1)
+    for path, package in packages.iteritems():
+        if metapackage.is_metapackage(package):
+            try:
+                metapackage.validate_metapackage(path, package, warnings=True)
+            except metapackage.InvalidMetapackage as e:
+                print("Invalid metapackage at path '%s':\n  %s\n" % (os.path.abspath(path), str(e)))
+                print("See requirements for metapackages: %s" % metapackage.DEFINITION_URL)
+                sys.exit(1)
     old_version = verify_equal_package_versions(packages.values())
     new_version = bump_version(old_version, args.bump)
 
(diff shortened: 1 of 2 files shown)


wjwwood:
@dirk-thomas and @tfoote for review

wjwwood:
Error output looks something like this:

```
Invalid metapackage at path '/Users/william/devel/catkin_pkg/test/data/metapackages/no_cmake':
  Metapackage 'no_cmake': No CMakeLists.txt

See requirements for metapackages: http://ros.org/reps/rep-0127.html#metapackage
```

Or

```
WARNING: Metapackage "invalid_depends" should not have other dependencies besides a buildtool_depend on catkin and run_depends.
Invalid metapackage at path '/Users/william/devel/catkin_pkg/test/data/metapackages/invalid_depends':
  Metapackage 'invalid_depends': Has build, buildtool, and/or test depends, but only run depends are allowed (except buildtool catkin)

See requirements for metapackages: http://ros.org/reps/rep-0127.html#metapackage
```

The extra `WARNING: ...` line is coming from `find_packages` which checks a few things about metapackages.

tfoote:
+1

bin/catkin_prepare_release
```diff
@@ -68,6 +69,14 @@ def main():
     if not packages:
         print('No packages found', file=sys.stderr)
         sys.exit(1)
+    for path, package in packages.iteritems():
+        if metapackage.is_metapackage(package):
+            try:
+                metapackage.validate_metapackage(path, package, warnings=True)
+            except metapackage.InvalidMetapackage as e:
+                print("Invalid metapackage at path '%s':\n  %s\n" % (os.path.abspath(path), str(e)))
+                print("See requirements for metapackages: %s" % metapackage.DEFINITION_URL)
```
dirk-thomas: Should these error message be printed to `, file=sys.stderr`?

wjwwood:
@dirk-thomas updated for review again.

dirk-thomas:
+1

wjwwood:
I will wait to merge this until https://github.com/ros-infrastructure/catkin_pkg/pull/40 is merged and catkin_pkg 0.1.11 is released.

Issue #407, opened 2013-04-09
When auth with repo fails, catkin_package_version should not increment version nor repeat commit


130s:
See [this question thread](http://answers.ros.org/question/60141/catkin_prepare_release-pushes-twice-wo-incrementing-tag/?answer=60349#post-id-60349). 

In my case, I might have failed auth once, then succeeds. That way the version in `package.xml` was 0.2.13 and `tag` in repo stayed 0.2.12 after one `catkin_package_version` execution. Also `commit` occurred as many as I failed auth + 1 success, with its msg remaining 0.2.12 for all commits.

dirk-thomas:
The message remains the same since you have not run `catkin_prepare_release` again but just the commands printed from the initial invocation.

#417 will ensure for non-svn repositories that the user has push access (else it will not perform any action).

In case the pushing still fails with #417 `catkin_prepare_release` will output a message that the user has to manually invoke the push (instead of triggering another release).

PR #417, opened 2013-05-02
rewrite catkin_prepare_release to execute the commands rather than printing them

dirk-thomas:
@tfoote @wjwwood Please review and comment.

Also addresses #407.

### DIFF
--- a/bin/catkin_prepare_release
+++ b/bin/catkin_prepare_release
@@ -3,47 +3,57 @@
 from __future__ import print_function
 import argparse
 import os
+import subprocess
 import sys
 
-from catkin_pkg.packages import find_packages, verify_equal_package_versions
 from catkin_pkg import metapackage
+from catkin_pkg.package import PACKAGE_MANIFEST_FILENAME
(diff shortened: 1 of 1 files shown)


bin/catkin_prepare_release
```diff
@@ -54,54 +64,94 @@ def get_commands(path, packages, vcs_type, bump, new_version, svn_url=None):
             raise RuntimeError('Could not determine base URL of SVN repository "%s"' % svn_url)
         tag_url = '%s/tags/%s' % (base_url, new_version)
         cmd = 'svn cp -m "tagging %s" %s %s' % (new_version, svn_url, tag_url)
-        result.append(cmd)
-    return result
+    else:
+        assert False, 'Unknown vcs type: %s' % vcs_type
+    try:
+        output = subprocess.check_output(cmd, cwd=base_path, stderr=subprocess.STDOUT)
```
tfoote, line 70: This is not python 2.6 compatible. but if we're only supporting groovy and higher that should be ok.

wjwwood:
+1

tfoote:
+1

PR #419, opened 2013-05-06
Documentation + base.rosinstall update

bcharrow:
I encountered two errors following the directions to build a ROS underlay:
http://ros.org/doc/groovy/api/catkin/html/adv_user_guide/underlay.html
1) The URL for the rosinstall file was wrong
2) Some of URI's in the rosinstall file were out of date / broken.

As everything in the rosinstall file is for groovy, perhaps it'd also be a good idea to rename base.rosinstall -> groovy_base.rosinstall or something similar?

### DIFF
--- a/doc/adv_user_guide/underlay.rst
+++ b/doc/adv_user_guide/underlay.rst
@@ -22,7 +22,7 @@ Pick a temp directory::
 
 Get the rosinstall file::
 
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Since these information always tend to be outdated we should just remove them from the repo and instead use the rosinstall generator webservice to generate an up-to-date list on demand.

Renaming the file would also not be a good idea since the content already belongs to the groovy-devel branch.

PR #422, opened 2013-05-14
fix chaining of plain cmake packages: remove exec '' statement

smits:
Signed-off-by: Ruben Smits ruben.smits@mech.kuleuven.be

### DIFF
--- a/python/catkin/builder.py
+++ b/python/catkin/builder.py
@@ -480,10 +480,7 @@ fi
 export PATH="{path}$PATH"
 export PKG_CONFIG_PATH="{pkgcfg_path}$PKG_CONFIG_PATH"
 export PYTHONPATH="{pythonpath}$PYTHONPATH"
-
-exec "$@"
-""".format(**subs)
-        )
+""".format(**subs))
 
 
 def build_package(


dirk-thomas:
I agree that the exec should be removed since setup.sh is never invoked with arguments. But could you please describe what use case was not working for you?

smits:
Op 14 mei 2013 22:16 schreef "Dirk Thomas" notifications@github.com het
volgende:

> I agree that the exec should be removed since setup.sh is never invoked
> with arguments. But could you please describe what use case was not working
> for you?

I had multiple plain cmake packages in my workspace which depend on each
other (orocos toolchain) , due to the exec, cmake was executed in the
setup.sh environment of the first package of my workspace chain instead of
building up the environment with all preceding packages, which resulted in
a failure of building a package that depends on a package which is not the
first one. Removing the exec makes sure the entire environment is chained
up before building a package.

Ruben

dirk-thomas:
Can you please describe a reproducible scenario (at best with step-by-step commands) since I can't reproduce it with a simple workspace consisting of two plain CMake packages where A build_depends on B?

smits:
This is expected, what you need are at least three packages build in the
order A B C and with C depending on B. Due to the early exec only the
environment of A is used to build C, with the required environment of B
missing.

I'll send an example later today if still needed.

smits:
Do you still need that example?

dirk-thomas:
I will give it a try with three packages and get back to you.

dirk-thomas:
After I was able to reproduce it with three plain packages I can understand how and why it breaks like that.

Thank you for describing it in detail and of course the patch.

Issue #410, opened 2013-04-17
provide task-oriented how-to documentation for catkin users


jack-oquin:
I consider this the biggest gap in the catkin user documentation.

There is a lot of information already, but it's not very user-oriented. Answers to specific questions are hard to find. Some pages assume a much deeper understanding of CMake than a lot of people have, including me. Other places assume an understanding of the overall catkin design, which is quite complex and very different from rosbuild.

I am working on a task-oriented how-to section in my [jack-oquin/catkin](https://github.com/jack-oquin/catkin) fork. I will periodically post formatted versions here: http://farnsworth.csres.utexas.edu/docs/catkin/html/howto/howto.html.

So far there is only an outline and part of one section. I'll make a note here when more text is available. 

Please provide suggestions, feedback and corrections.

jack-oquin:
A couple more sections written, outline expanded. Done for today.

http://farnsworth.csres.utexas.edu/docs/catkin/html/howto/howto.html

tfoote:
That looks great Jack!

jack-oquin:
Thanks!

It's coming along, slowly. Lots left to do.

I renamed the top-level howto page. It's here now: 
- http://farnsworth.csres.utexas.edu/docs/catkin/html/howto/index.html

jack-oquin:
There is now has a fairly complete first draft, missing only the Python and unit test pages. 
- http://farnsworth.csres.utexas.edu/docs/catkin/html/howto/index.html

jack-oquin:
First draft complete, except for the unit tests. I'd like some feedback on what is there now:
- http://farnsworth.csres.utexas.edu/docs/catkin/html/howto/index.html

dirk-thomas:
catkin_overview: CMakeLists.txt: Todo:
There is no specific order required between test and install invocations.

jack-oquin:
Noted. Thanks, Dirk!

dirk-thomas:
catkin_library_dependencies: package.xml:
"If all the C++ code is defined in header files and no shared object libraries are provided, the <run_depend> is not needed and should be omitted."

This is not correct. The condition when to declare a run_depend is if another package is required for executing this package. This is the case the the other package contains a shared library which is used (not for static libraries since they are included at build time). Another case is if the headers of the other package are "visible" to downstream packages - usually that is the case when the headers of this package transitively include headers of the other package. If the headers of the other package are only included in cpp files a run_depend is not required.

jack-oquin:
[...]

PR #429, opened 2013-05-28
Add how-to documentation

jack-oquin:
This is only a documentation change, safe to release to both Groovy and Hydro.

This partially fixes ros/catkin#410. The unit test sections still need to be written, but I am still learning how to do that myself, and it is difficult enough to significantly delay completion. 

So, I recommend releasing the other sections, which have already undergone considerable review on **ros-sig-buildsystem**. I see many questions on **answers.ros.org** that could benefit from this information.


jack-oquin:
A formatted draft, which currently matches this update, is here:
- http://farnsworth.csres.utexas.edu/docs/catkin/html/howto/index.html

doc/howto/building_libraries.rst
```diff
@@ -0,0 +1,94 @@
...
+  catkin_package(INCLUDE_DIRS include
+                 LIBRARIES your_library)
+
+  install(TARGETS your_library
+          RUNTIME DESTINATION ${CATKIN_GLOBAL_BIN_DESTINATION}
+          ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+          LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION})
+
+  install(DIRECTORY include/
+          DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION})
```
dirk-thomas, line 92: This looks wrong to me. The include folder should have a subfolder which namespaces the package headers but with this the content of the include folder would get installed into the package specific destination. It should be:

```
install(DIRECTORY include/${PACKAGE_NAME}/ DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION})
```

or (which is less strict because the user could put headers in the include root or different subfolders):

```
install(DIRECTORY include/ DESTINATION ${CATKIN_GLOBAL_INCLUDE_DESTINATION})
```

dirk-thomas, line 92: The install() invocation should also filter only header files (.h, .hpp) or explicitly exclude .svn folders.

doc/howto/building_msgs.rst
```diff
@@ -0,0 +1,91 @@
...
+
+  generate_messages(DEPENDENCIES std_msgs)
+
+A good ROS practice is to collect related messages, services and
+actions into a separate package with no other API.  That simplifies
+the package dependency graph.
+
+However, you *can* provide scripts and programs with the message
+package.  If you do, message generation targets need to be built
+before any of your other programs::
```
dirk-thomas: ... if they depend on the generated code for the messages.

doc/howto/building_msgs.rst
```diff
@@ -0,0 +1,91 @@
...
+
+  add_dependencies(your_program ${your_package_EXPORTED_TARGETS})
```
dirk-thomas, line 92: Should `${catkin_EXPORTED_TARGETS}` also be mentioned here?

doc/howto/building_msgs.rst
```diff
@@ -0,0 +1,91 @@
...
+
+Finally, add this near the end of ``CMakeLists.txt`` to export the
+messages, services and actions for use by other packages::
```
dirk-thomas: `catkin_package()` must appear before any usage of catkin variables (e.g. destination). Therefore it should not recommend to be called near the end.

doc/howto/catkin_library_dependencies.rst
```diff
@@ -0,0 +1,102 @@
...
+
+When your package depends on a catkin C++ library, there are usually
+several kinds of dependencies which must be declared in your
+``package.xml`` and ``CMakeLists.txt`` files.
+
+
+package.xml
+:::::::::::
+
+Binary and source package dependencies are declared in
```
dirk-thomas: I am not sure if the terminology of "binary" and "source" package dependencies is good here. It might be counter intuitive to introduce these term since they do not fit the build/run depends in all cases.

doc/howto/catkin_overview.rst
```diff
@@ -0,0 +1,127 @@
...
+  <url type="repository">https://github.com/ros-drivers/camera1394.git</url>
+  <url type="bugtracker">https://github.com/ros-drivers/camera1394/issues</url>
+
+
+Metapackages
+::::::::::::
+
+These are special-purpose catkin packages for grouping other packages.
+Users who install a metapackage binary will also get all packages
+directly or indirectly included in that group.  Metapackages should
```
dirk-thomas: "should not" -> "must not"

doc/howto/catkin_overview.rst
```diff
@@ -0,0 +1,127 @@
...
+
+In addition to the required ``<buildtool_depends>`` for catkin,
+metapackages list the packages in the group using ``<run_depends>``
+tags::
+
+  <run_depend>your_custom_msgs</run_depend>
+  <run_depend>your_server_node</run_depend>
+  <run_depend>your_utils</run_depend>
+
+Metapackages should not include any other ``package.xml`` elements.
```
dirk-thomas: "should not" -> "must not"

doc/howto/cpp_msg_dependencies.rst
```diff
@@ -0,0 +1,54 @@
...
+  <build_depend>std_msgs</build_depend>
+  <build_depend>sensor_msgs</build_depend>
+
+  <run_depend>std_msgs</run_depend>
+  <run_depend>sensor_msgs</run_depend>
+
+Although C++ message references are resolved at build time via
+headers, providing a ``<run_depend>`` is a good idea, making the
+message definitions available for useful Python programs, like
+rosmsg_.
```
dirk-thomas: The run_depend is not really optional. The generated C++ headers are useless without the run_depends since transitive includes would be missing. So it should state that run_depends are always needed.

doc/howto/installing_other.rst
```diff
@@ -0,0 +1,20 @@
...
+Installing other files
+----------------------
+
+Sometimes your package needs to install additional files, like
+`roslaunch scripts`_ or parameter settings.
+
+In most cases these data are platform-independent, so install them
+within your package's share directory::
+
+  install(TARGETS your_data your_parameters
```
dirk-thomas: `TARGETS` seems to be wrong for platform-independent files. I guess it should be `FILES`?

doc/howto/installing_other.rst
```diff
@@ -0,0 +1,20 @@
...
+          DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
+
+This example installs everything in your **launch/** subdirectory::
+
+  install(DIRECTORY launch/
+          DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/launch)
```
dirk-thomas: This either needs an exclusion list for e.g. .svn stuff and/or use `SCRIPTS` to install whitelisted files only.

doc/howto/installing_python.rst
```diff
@@ -0,0 +1,89 @@
...
+ROS executables are installed in a per-package directory, not the
+distributions's global ``bin/`` directory.  They are accessible to
+rosrun_ and roslaunch_, without cluttering up the shell's ``$PATH``,
+and their names only need to be unique within each package.  There are
+only a few core ROS commands like ``rosrun`` and ``roslaunch`` that
+install in the global ``bin/`` directory.
+
+Standard ROS practice is to place all your executable Python programs
+in a ``scripts/`` subdirectory.  To keep the user API clean,
+executable script names generally to not include a ``.py`` suffix.
```
dirk-thomas: "to" -> "do"

doc/howto/installing_python.rst
```diff
@@ -0,0 +1,89 @@
...
+Your ``CMakeLists.txt`` can conveniently install all the scripts in
+that directory::
+
+  install(DIRECTORY scripts/
+          DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION})
```
dirk-thomas: Again excluding stuff like .svn or whitelisting.

doc/howto/system_library_dependencies.rst
```diff
@@ -0,0 +1,132 @@
...
+which only require the library's run-time binary files.  But, it is
+also needed when the dependency provides a header file included in
+some header exported from your package.::
+
+  <run_depend>libgstreamer0.10-0</run_depend>
+
+Unfortunately, many existing rosdep entries only resolve the library's
+"development" package.  If no appropriate run-time package key is
+defined, consider `contributing the missing rules`_ so users need not
+install unnecessary build files.
```
dirk-thomas: The run_depend can only be a non-dev package if none of its headers is exposed by the headers of this package. But commonly that is the case why the run_depend still has to state the -dev package.

doc/howto/system_library_dependencies.rst
```diff
@@ -0,0 +1,132 @@
...
+
+  <build_depend>curl</build_depend>
+  <run_depend>curl</run_depend>
+
+Some C++ packages, like Eigen, have no run-time library.  All code is
+defined in the header files.  Since those files are only needed at
+compile time, you should not specify a ``<run_depend>``, unless you
+also export headers that include them::
+
+  <build_depend>eigen</build_depend>
```
dirk-thomas, line 75: Again that is only try if the package uses Eigen only inside of cpp files. As soon as Eigen is imported in an exposed header it must be a run_depend too.

doc/howto/system_library_dependencies.rst
```diff
@@ -0,0 +1,132 @@
...
+
+  find_package(Boost REQUIRED)
+
+But, the boost thread impementation *does* require a library, so
+specify "COMPONENTS thread" if you need it::
+
+  find_package(Boost REQUIRED COMPONENTS thread)
+
+These ``find_package()`` calls define CMake variables that will be
+needed later for the compile and linkedit steps.
```
dirk-thomas: You should explicitly mention that the naming of the variable is not always consistent and should be checked before using them.

doc/howto/system_library_dependencies.rst
```diff
@@ -0,0 +1,132 @@
...
+package also depends on other catkin packages, add
+``${catkin_INCLUDE_DIRS}`` to the list.
+
+Exporting interfaces
+''''''''''''''''''''
+
+Declare the non-catkin system library and header packages needed by
+all the interfaces you export to other ROS packages::
+
+  catkin_package(DEPENDS Boost GSTREAMER)
```
dirk-thomas, line 147: This only works when the exposed variables follow the recommended schema. If the package has different variables the libraries and include paths must be passed explicitly to `catkin_package(INCLUDE ... LIBRARIES ...)`.

doc/index.rst
```diff
@@ -71,3 +71,4 @@ Contents
    user_guide/user_guide
    adv_user_guide/adv_user_guide
    dev_guide/dev_guide
+   howto/index
```
dirk-thomas: Should this be earlier in this list e.g. after the user guide?

dirk-thomas:
One very general question: how does all these new how-to's relate to the existing documentation in the Wiki as well as the guides in the repo? Should they all be kept? Is there no overlap? Or should parts of the existing docs be removed in favor of these new how-to's?

doc/index.rst
jack-oquin: Perhaps.

I was unable to discern the organizational outline of the user guide, so I linked it at the end, hoping to at least do no harm.

jack-oquin:
These pages were written because I could not find the information I needed in a form I could use.

As you know, I favor holding a complete catkin doc review. That would be an appropriate forum to answer your overall question. Although I don't find most of the catkin docs useful, there may be others who do.

In my view, most of the existing catkin documentation is either pitched at too high a level of expertise, or not organized to meet user needs. Every doc should begin with an understanding of what readers already know and what they are trying to learn how to do. People don't read technical documentation for pleasure, they always have a goal in mind.

Most of the catkin sections in other packages' tutorials are necessary, although many could use improvement. Perhaps some of them could cross-reference to various how-to pages, if appropriate.

doc/howto/building_libraries.rst
jack-oquin, line 92: It does work correctly, barring `.svn` subdirectories. But, it is not as restrictive as you would like.

The preceding text already requested that people place their headers in `include/your_package_name/`. To be consistent with the other pages, that should have been `include/your_package`. But, there are cases where a different subdirectory name may be appropriate, usually for API compatibility reasons. I am trying to provide the simplest solution that works, although I do not wish to provide examples of bad practice.

Filtering the file extensions does not appeal to me. Who knows what extensions people may be using? Other variants include `.hh`, `.H`, and `.hxx`, and others that don't come to mind right now. Why drag the discussion into those unnecessary details? 

I would prefer to add a sentence after "that last command installs the entire subtree of headers", suggesting that it is a bad idea to mix other files into the header folder, but if you must, then by all means install each header explicitly by name.

The `.svn` issue _is_ a problem, however. Do you recall the exact syntax for ignoring it?

dirk-thomas, line 92: I don't think that the include installation works as you expect it to. Please see two example where it is used and proven to work correctly:
The prefered way when all header are part of the subfolder PROJECT_NAME:
https://github.com/ros-visualization/rqt_common_plugins/blob/groovy-devel/rqt_image_view/CMakeLists.txt#L59
The generic way where folders might be in an arbitrary subfolder of include:
https://github.com/ros/ros_comm/blob/groovy-devel/tools/rosconsole/CMakeLists.txt#L59

I agree, that excluding .svn is sufficient and header files should not need to be listed explicitly (if the user does not put other files in there - but that should be best practice anyway).

For an example for the exclude rule please see:
https://github.com/ros/ros/blob/groovy-devel/tools/roscreate/CMakeLists.txt#L10

jack-oquin, line 92: Yes, I see now that I had mixed up `${CATKIN_PACKAGE_INCLUDE_DESTINATION}` and `${CATKIN_GLOBAL_INCLUDE_DESTINATION}`. 

Thanks for catching that, Dirk!

jack-oquin, line 92: Commit 7da463838ad9d8b49c7f0453a0a172bfb2b738f0 should fix the problems with this section.

Please check it for me.

doc/howto/catkin_library_dependencies.rst
jack-oquin: This could be greatly simplified by recommending `<depend>`, instead of trying to explain how `<build_depend>` and `<run_depend>` actually work.

Should we do that?

doc/howto/cpp_msg_dependencies.rst
jack-oquin: Here again, `<depend>` would simplify things.

doc/howto/catkin_library_dependencies.rst
dirk-thomas: @wjwwood suggested some time ago that we might not allow packages to be released when they use that <depend> tag since it is likely used as shortcut and might result in bloated dependencies. If that is the case all tutorials should still teach how to use the individual dependency tags.

jack-oquin: Surely you are joking?

dirk-thomas: I think that was more related to the usage of catkin_simple and the assumption that people will just always use depend and that released packages should have more specific dependencies.

When just being used as a replacement for build and run dependencies if both are necessary it is find to use depend.

doc/howto/system_library_dependencies.rst
jack-oquin: What is the best way to check them?

dirk-thomas: Checking if the variable is defined and/or outputting them and verifying the value.

jack-oquin: You really don't want me writing my own CMake code. 

I am hoping for a concrete example that can be used or adapted for the text.

jack-oquin:
Note to self: add a new page for "Building and installing CMake files".

See: this [ros-sig-buildsystem thread](https://groups.google.com/forum/?hl=en#!msg/ros-sig-buildsystem/q-HXY-tS_20/BH1KlBrKaKYJ).

doc/index.rst
jack-oquin: I agree: it should be put _much_ earlier on the top Sphinx page.

That probably requires moving much of the top-level text into subordinate pages. Should I give it a try?

jack-oquin: I'll move it after the user guide for now.

After this large batch has been merged, I plan to work on the top-level index page to make the information more accessible to first-time readers. There is a whole screen full of verbiage to wade through top before getting to the "Contents". The "benefits" and "Installation" sections should move down a level.

jack-oquin:
I have made all the recommended changes. Is this ready to merge now?

dirk-thomas:
Looks good to be merged.

Could you squash some / all of the commits? Since it includes so many commits and spans a long time it would look very confusing if not being squashed.

jack-oquin:
Note that this PR assumes existence of the `<depend>` tag added by ros-infrastructure/catkin_pkg#43. It should not be released until after that.

While slogging through the wretched git rebase squash results (I only mangled my local repository once), I noticed a few more minor editorial changes. They will be coming soon in a separate commit.

dirk-thomas:
Awesome job - thank you very much for your continuous effort.

Issue #426, opened 2013-05-23
download_test_data: not documented and difficult to use


jack-oquin:
There is no documentation for this command.

I am converting a rosbuild package to catkin. It used `rosbuild_download_test_data()`, which I changed to `download_test_data()`.

The file gets downloaded to the build tree, but in rostest `$(find PACKAGE)/tests/FILENAME.pcap` can't find it there.

See [this question](http://answers.ros.org/question/63189/how-do-i-use-download_test_data/) for details, also the discussion in #416.

jack-oquin:
I tried to hack around the problem by using an absolute file name in **/tmp**. 

```
download_test_data(
    http://pr.willowgarage.com/data/velodyne/32e.pcap
    /tmp/${PROJECT}/tests/32e.pcap
    e41d02aac34f0967c03a5597e1d554a9)
```

But, `download_test_data()` insists on prefixing the name with `${PROJECT_BINARY_DIR}/`, even when the it is fully qualified. It _always_ goes in the build tree, although it cannot be resolved there.

jack-oquin:
The following appalling hack seems to work:

```
set(SAVE_PROJECT_BINARY_DIR ${PROJECT_BINARY_DIR})
set(PROJECT_BINARY_DIR /tmp/velodyne_driver)
download_test_data(http://pr.willowgarage.com/data/velodyne/class.pcap
                   tests/class.pcap
                   65808d25772101358a3719b451b3d015)

download_test_data(http://pr.willowgarage.com/data/velodyne/32e.pcap
                   tests/32e.pcap
                   e41d02aac34f0967c03a5597e1d554a9)
set(PROJECT_BINARY_DIR ${SAVE_PROJECT_BINARY_DIR})
```

I even had to spell out the package name, because I could not get CMake to resolve `/tmp/${PROJECT}` correctly in the `set()` command.

dirk-thomas:
The CMake variable for the project name is called `PROJECT_NAME` not `PROJECT`.

I don't have a package at hand which does it the way you describe it. Can you try the following (or point me to an existing repo with which I could give it a try)?

Since the test data is downloaded to a location relative to `PROJECT_BINARY_DIR` have you tried to register your test with a `WORKING_DIRECTORY` which points to `PROJECT_BINARY_DIR`? Than it might be possible to just reference the test data with a relative path (without any find logic).

jack-oquin:
[...]

PR #431, opened 2013-05-31
add new CMake function catkin_download_test_data

dirk-thomas:
mark download_test_data as deprecated.

Based on the discussion in #426 I have added a new download function with the necessary API to use it in any way suggested in there. It also exposes the target name explicitly (as any other CMake function) to enable the user land code to add custom dependencies.

### DIFF
--- a/cmake/all.cmake
+++ b/cmake/all.cmake
@@ -120,6 +120,7 @@ foreach(filename
     platform/lsb
     platform/ubuntu
     platform/windows
+    test/catkin_download_test_data
     test/download_test_data
     test/gtest
     test/nosetests
--- a/cmake/test/catkin_download_test_data.cmake
+++ b/cmake/test/catkin_download_test_data.cmake
@@ -1,0 +1,48 @@
+#
+# Download a file containing test data from a URL.
+#
+# It is commonly used to download larger data files for unit tests
+# which should not be stored in the repository.
+#
+# .. note:: The target will be registered as a dependency
+#   of the "tests" target.
+#
+# .. note:: If the tests should be run on the ROS buildfarm the URL
+#   must be publically and reliably accessible.
+#
+# :param target: the target name
+# :type target: string
+# :param url: the url to download
+# :type url: string
+
+# :param DESTINATION: the directory where the file is downloaded to
+#   (default: ${PROJECT_BINARY_DIR})
+# :type DESTINATION: string
+# :param FILENAME: the filename of the downloaded file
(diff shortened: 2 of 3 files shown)


jack-oquin:
The new interface looks good, Dirk. I'll test it as soon as I find time.

tfoote:
reviewed +1

jack-oquin:
It would be extremely helpful to know when something like this is actually released, so I can change my own tests to depend on it.

dirk-thomas:
This is what the changelog files are for. The maintainer can either give an expected date for the forthcoming section in there or use external infrastructure like milestones.

This particular change will be released either today or tomorrow (plus the time it takes to rebuilding everything plus time until it get synced).

jack-oquin:
How does the change log cause me to be notified that the release happened?

dirk-thomas:
The maintainer can provide the information in the forthcoming section, e.g. "expected to be released around mid June". Or he can create a milestone "next release" with a due date.

If that is not the case you can only track the ros/rosdistro/DISTRO/release.yaml file for changes or wait for the announcement from Tully once the packages have been synced.

jack-oquin:
That is polling, rather than notification. 

Why not just leave the issue open until the fix is released?

dirk-thomas:
It is common procedure to close issues when the fix gets applied to the source repo.

Closing issues when they are available in public is even more cumbersome, since a) multiple distros might have very different sync times and b) the person performing the sync does not even have the necessary information to close all affected issues in the upstream repos.

The closest I could imaging would be an rss feed on when a specific package changes in a given distro but someone would need to write a web service like this.

jack-oquin:
Closing at the time of release would be good enough. Once it's released, other things that depend on it can be pre-release tested and then released themselves.

That can be done via github references to the issues fixed. The change log should contain all those references, anyway, so it's easily done.

dirk-thomas:
Just for the record: the merge with the hash above does not appear in the groovy-devel branch (looks like i screwed up the merge somehow). It is now merged in 1480bf788b6e150fb056a76a50d83f77978a38b8.

jack-oquin:
The new command does not appear on the extracted CMake API reference page:
- http://ros.org/doc/groovy/api/catkin/html/dev_guide/generated_cmake_api.html

jack-oquin:
Seems to work:
- https://github.com/ros-drivers/velodyne/commit/ac1f34be45b3032c90641f7ed0c800e7d3ea0e73

Issue #440, opened 2013-06-19
documentation: separate format1 and format2 how-to pages


jack-oquin:

jack-oquin:
Github won't let me assign myself to fix this. I guess it requires admin authority.

jack-oquin:
Please assign this to me. I'll work on it as soon as I can find time.

wjwwood:
That'd be great @jack-oquin, let me know if you need help or someone to review it.

jack-oquin:
I'll certainly need reviewers, but the actual writing should not take long.

I plan to keep a copy of the format1 version, so users can refer to either. Format2 will be the recommended version, but there will still be format1 packages around for years. I think clearly separate docs will be less confusing for the readers.

jack-oquin:
I just noticed that catkin has a separate Indigo source branch, which differs from the Groovy and Hydro branch. Is format2 only supported on Indigo, or does it work in the other branch, too?

I see there are already documentation differences between `indigo-devel` and `groovy-devel`.

dirk-thomas:
Since the package.xml format is implemented in catkin_pkg it is independent of the catkin version.  As soon as a catkin version with #619 is released format 2 can be used. The changes will also be released back into Groovy / Hydro. Then format 1 as well as 2 can be used with any current ROS distro (G, H and I).

Several things have changed between the groovy-devel and indigo-devel branch. Please compare the changelogs for details: https://github.com/ros/catkin/blob/groovy-devel/CHANGELOG.rst https://github.com/ros/catkin/blob/indigo-devel/CHANGELOG.rst

Especially deprecated functionality has been removed as well as the Eigen CMake config file. Furthermore some Python 3 specific changes have been added as well as some restrictions removed to run unit tests multi threaded. And there is an Indigo migration guide page in the docs.

jack-oquin:
That's good. 

It means we'll have to update the docs in both branches. I plan to provide an indigo-devel PR for review. Once it looks OK, we can merge those commits into groovy-devel, too.

dirk-thomas:
That would be great. Thank you for your effort.

jack-oquin:
Any idea where these links have moved?

```
WARNING: intersphinx inventory 'http://ros.org/doc/api/rospkg/html/objects.inv' not fetchable due to <class 'urllib2.HTTPError'>: HTTP Error 404: Not Found
[...]

PR #441, opened 2013-06-19
Separate format1 and format2 documentation.

jack-oquin:
Fixes ros/catkin#440.
- The main how-to index is for format2. It refers format1 users to a subdirectory.
- The format2 version recommends `<depend>`, but does not yet mention `<devel_depend>`
  or `<exec_depend>`.
- The format1 version does not mention `<depend>`.

There is a formatted version here: http://farnsworth.csres.utexas.edu/docs/catkin/html/howto/index.html

### DIFF
--- a/doc/howto/downloading_test_data.rst
+++ b/doc/howto/downloading_test_data.rst
@@ -1,0 +1,56 @@
+.. _downloading_test_data:
+
+Downloading test data
+---------------------
+
+When unit tests require large data files, it is better to download
+them from the web than include them in your source repository.
+
+Each of these files declares a target name.  Running ``"make tests"``
+or ``"make run_tests"`` will automatically download them all.
+Individual test targets do *not* depend on the ``tests`` target, but
+you can use it to download them before running one::
+
+  $ make tests run_tests_your_package
+
+Catkin provides a convenient command, used like this::
+
+  catkin_download_test_data(
+    ${PROJECT_NAME}_32e.pcap
+    http://pr.willowgarage.com/data/velodyne/32e.pcap
+    MD5 e41d02aac34f0967c03a5597e1d554a9)
+
+The first parameter is the target name, which should normally include
+your project name to avoid conflicts with other packages.
+
+The second parameter is the URL to read.  If you release your package
+to the ROS build farm, make sure this URL is available reliably.
+Otherwise, tests will fail randomly, annoying everyone.  Contact
+``ros-release@code.ros.org`` if you need to host some data near the
+build servers.
+
+The MD5 argument is a good way to avoid testing with corrupted data.
+
+The default destination for the file downloaded above is within your
+package's build directory.  The file name is the base name of the URL.
+You can use the DESTINATION argument to put it somewhere else.
+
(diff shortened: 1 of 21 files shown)


jack-oquin:
@dirk-thomas, @tfoote, @wjwwood:

I feel some urgency about this update to avoid misleading users. Also, I have other test pages queued up behind it.

Would you like to review what I've done? Should I just go ahead and merge it?

dirk-thomas:
We should not merge any of these pull request until we do have an updated REP for that and catkin_pkg actually accepting what is documented in here.

jack-oquin:
If you look  closely, you will see that this push is to handle the fact that the REP update has _not_ been accepted. 

The update you should _not_ have accepted was  #429, which explictly states that it should not be accepted until after ros-infrastructure/catkin_pkg#43.

dirk-thomas:
I will look into reverting the parts from #429 until they are available.

This one should obviously not be merged: e.g. https://github.com/jack-oquin/catkin/commit/19861f2c653c463acd890df55f18d0c57cd26f95#L0R16

jack-oquin:
This is an attempt to follow the instructions of @tfoote:  

> The docs need to be readable in the recommended format.  For this change we should write the docs for v2 as it will be globally recommended, with side notes/assides about differences to format 1.  

Have you even looked at it? [Here](http://farnsworth.csres.utexas.edu/docs/catkin/html/howto/index.html) it is.

This doc describes format2, with a link at the top to a separate set of format1 pages. Those format1 pages will work correctly with the current release. Did I misunderstand Tully's instructions?

If you have substantive comments, please make them. I welcome constructive criticism. If you want it done differently, say so. But, don't waste my time by rejecting work thoughtlessly.

dirk-thomas:
As stated in my comment above this pull request mentions format 2 in crucial parts of the docs without mentioning that this part is not even available yet (e.g. https://github.com/jack-oquin/catkin/commit/19861f2c653c463acd890df55f18d0c57cd26f95#L0R16). A user might jump to the page http://farnsworth.csres.utexas.edu/docs/catkin/html/howto/catkin_overview.html (e.g. via a Google search) without visiting the page with the note you refer to.

Since the documentation will be generated every night and would reflect these changes immediately on the ROS wiki it is imho not a good idea to merge this until (!) the REP has been updated and the tools can actually handle the examples.

If other people think it is a good idea to merge it anyway - please go ahead - I will leave the decision to merge this pull request to @tfoote.

tfoote:
We should hold this one, and we should have held the previous one, but it's
probably not worth merging it back out at this point.  Lets hold this until
we've formalized the new format.

jack-oquin:
Having done all the work to separate the two formats, I can fairly quickly provide a version with only the format1 information. I would have done that already had you said to do it that way. It was one of the two options I mentioned on the mailing list.

I _strongly_ recommend merging that rather than leaving the current in-between version, which mentions `<depend>` but not `format="2"`, `<devel_depend>` or `<exec_depend>`.

jack-oquin:
I created a new format1 branch with no format2 features. 

Closing this pull request in favor of #443.

PR #443, opened 2013-06-21
Format 1 (only) version of how-to documentation

jack-oquin:
This version is written for only the `<package format="1">` package.xml version.

It does include some new pages for unit tests, which were already written but not submitted.


jack-oquin:
The format1-only documentation is here: 
- http://farnsworth.csres.utexas.edu/docs/catkin-format1/html/howto/index.html

tfoote:
+1 That looks great.

dirk-thomas:
+1

doc/howto/gtest_configuration.rst
```diff
@@ -2,3 +2,34 @@
...
+  <test_depend>rosunit</test_depend>
+
+
+CMakeLists.txt
+::::::::::::::
+
+Declare each gtest like this::
+
+  catkin_add_gtest(test_your_node tests/test_your_node.cpp)
+  target_link_libraries(test_your_node ${catkin_LIBRARIES})
```
herbrechtsmeier, line 25: The `target_link_libraries` need to be wrapped with a conditional block of `GTEST_FOUND` because the `catkin_add_gtest` function is skipped if `GTEST_FOUND` isn't true.

jack-oquin, line 25: Thanks, @herbrechtsmeier. Would this work?

```
if (GTEST_FOUND)
  catkin_add_gtest(test_your_node tests/test_your_node.cpp)
  target_link_libraries(test_your_node ${catkin_LIBRARIES})
endif()
```

It seems a little cleaner and easier to read, and it could be wrapped around multiple test declarations.

herbrechtsmeier, line 25: This should work, but there is currently a discussion about this topic on the mailing list: https://groups.google.com/forum/#!topic/ros-sig-buildsystem/-spU8G49r8U.

jack-oquin, line 25: I will wait for that to be resolved before updating anything.

PR #448, opened 2013-06-25
add options to configure without tests

dirk-thomas:


vrabaud:
At the bottom of the main CMakeLists.txt of catkin, you want:
if(CATKIN_ENABLE_TESTING)
catkin_add_nosetests(test/local_tests)
catkin_add_nosetests(test/unit_tests)
endif()

vrabaud:
Otherwise, I get an error:
CMake Error at catkin/cmake/test/tests.cmake:125 (add_dependencies):
  add_dependencies Adding dependency to non-existent target:
  run_tests_object_recognition_tod_gtest_tod_clique
  catkin/cmake/test/gtest.cmake:58 (catkin_run_tests_target)
  object_recognition_tod/test/CMakeLists.txt:20 (catkin_add_gtest)

add_test does not create a target. I know, the doc says it "adds a test target" but that is not a proper target (I believe it's an internal target like "test" and "install"). What you want I believe is replace line 238 by:

```
add_dependencies(run_tests_${PROJECT_NAME}_${type} ${_testing_DEPENDENCIES})
```

(works for me) as run_tests_${PROJECT_NAME}_${type}_${name} is not a proper target

dirk-thomas:
Oh, good catch. The CMake doc is pretty unspecific about it - but after knowing how it is meant it becomes clear what they mean. I have integrated the condition for the test and came up with a different solution for the add_test() stuff. I updated the pull request.

PR #454, opened 2013-07-01
Add the --from-pkg option to catkin_make_isolated

wjwwood:
This is mutually exclusive with --pkg and takes
a single package which all packages leading up to
it are skipped, it is built and then the process
continues as normal.

### DIFF
--- a/bin/catkin_make_isolated
+++ b/bin/catkin_make_isolated
@@ -50,8 +50,11 @@ def parse_args(args=None):
         help='Runs cmake explicitly for each catkin package.')
     add('--no-color', action='store_true', default=False,
         help='Disables colored output (only for catkin_make and CMake)')
-    add('--pkg', nargs='+', metavar='PKGNAME', dest='packages',
-        help='Invoke "make" on specific packages (only after initial invocation)')
+    pkg = parser.add_mutually_exclusive_group(required=False)
+    pkg.add_argument('--pkg', nargs='+', metavar='PKGNAME', dest='packages',
+                     help='Invoke "make" on specific packages (only after initial invocation)')
+    pkg.add_argument('--from-pkg', metavar='PKGNAME', dest='from_package',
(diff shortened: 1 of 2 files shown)


bin/catkin_make_isolated
```diff
@@ -114,11 +117,12 @@ def main():
...
         force_cmake=opts.force_cmake,
         colorize=not opts.no_color,
-        build_packages=opts.packages,
+        build_packages=opts.packages or ([] if opts.from_package is None else [opts.from_package]),
         quiet=opts.quiet,
         cmake_args=cmake_args,
         make_args=opts.make_args,
-        catkin_make_args=opts.catkin_make_args
+        catkin_make_args=opts.catkin_make_args,
+        continue_from_pkg=True if opts.from_package else False
```
dirk-thomas: Can you update that to something like `opts.from_package is not None`?

wjwwood:
Done.

bin/catkin_make_isolated
```diff
@@ -114,11 +117,12 @@ def main():
...
         force_cmake=opts.force_cmake,
         colorize=not opts.no_color,
-        build_packages=opts.packages,
+        build_packages=opts.packages or ([] if opts.from_package is None else [opts.from_package]),
         quiet=opts.quiet,
         cmake_args=cmake_args,
         make_args=opts.make_args,
-        catkin_make_args=opts.catkin_make_args
+        catkin_make_args=opts.catkin_make_args,
+        continue_from_pkg=True if opts.from_package is not None else False
```
dirk-thomas: I actually meant to remove the `True/False` part. `opts.from_package is not None` is already boolean enough.

wjwwood: Ok, I understand what you mean, sorry I just barely glanced at your comment and didn't really think about it.

wjwwood:
Updated correctly this time.

Issue #447, opened 2013-06-25
catkin_prepare_release: needs more flexible tag naming option


jack-oquin:
I would like to use this tool, but most of my packages use the `package_name-:{version}` rule for release tags.

Since `catkin_prepare_release` creates a tag with just the new version number, I have to delete that tag and create the correct one manually. So, I find myself doing the whole thing by hand, rather than take advantage of this useful tool.

Adding an appropriate option looks straightforward. I will submit a patch when I get the chance, but will not be able to work on it for at least a couple of weeks. Anyone else is welcome to fix it before then.

jack-oquin:
With this patch, I can do this and have the tags generated correctly:

```
$ catkin_prepare_release -t package_name-
```

dirk-thomas:
Just out of curiousity: do you have more than one package in a repo or why do you prefer the additional prefix?

jack-oquin:
More than one package: yes. But, it's really the repo name that is used as a tag prefix.

As to why: it's mostly a matter of longstanding habit. I believe rosrelease did it that way. Bloom supports that style, but catkin_prepare_release does not.

jack-oquin:
With SVN repositories, which do contain multiple collections of packages, it is probably necessary.

dirk-thomas:
For SVN there no real difference regarding tagging:
- if it uses the structure `package / trunk|branches|tags` it is fine to just tag with the version number.
- if it uses the structure `trunk|branches|tags / package` it would require a prefix but would also tag _too much_ since the tag will include everything.

jack-oquin:
Maybe it's not _necessary_, but there still exist older repositories from pre-catkin days using that tag syntax. The rosrelease tool created them.

Why are you so reluctant to support it?

dirk-thomas:
I wrote:

```
Just out of curiousity
```

I am not reluctant to support it at all. As soon as the PR is ready it will be merged immediately. I asked these questions only to understand why people want to do that in order to understand their use case better.

jack-oquin:
I see. 

The main reason is that old repositories used a certain tag style, and it's convenient not to have to change it.

wjwwood:
Can this not be closed?

PR #488, opened 2013-07-24
catkin_prepare_release: add --tag-prefix argument (#447)

jack-oquin:


bin/catkin_prepare_release
```diff
@@ -176,6 +176,7 @@ def main():
         description='Runs the commands to bump the version number, commit the modified %s files and create a tag in the repository.' % PACKAGE_MANIFEST_FILENAME)
     parser.add_argument('--bump', choices=('major', 'minor', 'patch'), default='patch', help='Which part of the version number to bump? (default: %(default)s)')
     parser.add_argument('--no-color', action='store_true', default=False, help='Disables colored output')
+    parser.add_argument('-t', '--tag-prefix', help='release tag name prefix', default='')
```
dirk-thomas: Can you please update the help text to match phrasing and style of the other argument.

bin/catkin_prepare_release
```diff
@@ -279,22 +280,22 @@ def main():
...
 
     else:
         # for other vcs types the changes are first done locally
         print(fmt('@{gf}Committing the package.xml files...'))
         commit_package_xml_files(base_path, vcs_type, packages, new_version)
 
-        print(fmt("@{gf}Creating tag '@{boldon}%s@{boldoff}'..." % new_version))
-        tag_repository(base_path, vcs_type, new_version)
+        print(fmt("@{gf}Creating tag '@{boldon}%s@{boldoff}'..." % (args.tag_prefix+new_version)))
+        tag_repository(base_path, vcs_type, args.tag_prefix+new_version)
```
dirk-thomas: For all of the above changes lines please keep whitespaces around the `+` operator (as spec. in PEP8).

Furthermore I think it would make sense to set a variable `tag_name` somewhere before which has the value `args.tag_prefix + new_version` and then this variable is used in the five locations.

bin/catkin_prepare_release
jack-oquin: Sure. I'll fix them all if you like. Every one has a different style. 

Which do you prefer?

I suggest emulating the -h option, since we have no choice about that one.

dirk-thomas: I think a real sentence makes sense - the question for bump is actually ugly. And the help should state information about the default value.

I would not make it like -h since the majority of our tools does not follow that style.

dirk-thomas: But for this PR you can just update the help for the newly added argument. Rephrasing the other ones can be done separately.

jack-oquin: OK. 

I ask again, which style do you prefer?

A sentence is capitalized and ends with a period. Do you prefer imperative voice or third person singular indicative?

dirk-thomas: Proposal:

```
Add a prefix to the created tag
```

jack-oquin: OK, imperative voice, no period at the end. I'll mention the default value, too.

dirk-thomas: If the help says _add_ it might not need a default value. Makes only more sense for stuff like _bump_ where a specific value is selected by default.

jack-oquin:
I believe all the initial review issues are resolved. Is this version ready to merge, now?

dirk-thomas:
GitHub does not notify people when extending a pull request with a new commit so I did not noticed your change until you commented.

The changes look good - thank you for adding this option.

PR #507, opened 2013-09-09
add RUBYLIB=lib/ruby/vendor_ruby to setup.bash for ruby programs

OTL:
I want to add ruby library path for setup.sh.

I'm releasing rosruby, and it needs additional environmental variable " RUBYLIB=/opt/ros/hydro/lib/ruvy/vendor_ruby".
If catkin_init_workspace supports this variable generation, it is more easy to use ros ruby libraries.

### DIFF
--- a/cmake/templates/_setup_util.py.in
+++ b/cmake/templates/_setup_util.py.in
@@ -55,6 +55,7 @@ ENV_VAR_SUBFOLDERS = {
     'PATH': '@CATKIN_GLOBAL_BIN_DESTINATION@',
     'PKG_CONFIG_PATH': 'lib/pkgconfig',
     'PYTHONPATH': '@PYTHON_INSTALL_DIR@',
+    'RUBYLIB': 'lib/ruby/vendor_ruby',
 }
 
 


tfoote:
@dirk-thomas should we be able to inject this from the ruby package?

OTL:
ah, we should avoid edit catkin directory because it does not officially support ruby.
But if we use catkin file layout, we should store ruby libraries in ~/catkin_ws/lib/ruby/vendor_ruby/PACKAGE_NAME  or something..

Easiest way is adding environmental variable like below in ~/.bashrc.

```
export RUBYLIB=/opt/ros/groovy/lib/ruby/vendor_ruby
```

If you want to do this automatically, what is the best way?

wjwwood:
Rather than inject this with catkin, you should use an environment hook, which you can "export" from your `rosruby` package where catkin will pick it up and execute it when someone "sources" the workspace containing it. For example see how `roslib` sets the `ROS_PACKAGE_PATH`:

https://github.com/ros/ros/blob/groovy-devel/core/roslib/env-hooks/10.ros.sh.em#L33
https://github.com/ros/ros/blob/groovy-devel/core/roslib/CMakeLists.txt#L27

wjwwood:
http://docs.ros.org/groovy/api/catkin/html/dev_guide/generated_cmake_api.html#catkin-add-env-hooks

dirk-thomas:
@wjwwood Thanks for the great description and links.

@OTL The links to roslib should provide you with enough information to add an environment hook to the rosruby package. I will close this pull request since the change will not affect catkin. But if you need further help please don't hesitate to comment on this.

OTL:
thanks. i did not know catkin_add_env_hooks!

PR #508, opened 2013-09-13
support multilple package name for catkin_make --pkg (fix #504)

### LINKED ISSUE #504: Support multiple package names with catkin_make --pkg
Would it be easily possible to extend the `catkin_make --pkg` command to support building multiple packages at once?

I am using a rosbuild workspace that overlays a catkin devel-space. In the meantime, there are quite a number of packages in the catkin workspace. I would like to be able to run something like

```
cd $MY_CATKIN_WORKSPACE &&
  catkin_make --pkg `rospack depends my_rosbuild_package`
```

just to get the idea. Of course in this simple use case `catkin_make` would have to ignore package names that are not part of the catkin workspace.

dirk-thomas:
@wjwwood Please review

### DIFF
--- a/bin/catkin_make
+++ b/bin/catkin_make
@@ -100,8 +100,14 @@ def main():
     # verify that specified package exists in workspace
     if args.pkg:
         packages_by_name = {p.name: path for path, p in packages.iteritems()}
-        if args.pkg not in packages_by_name:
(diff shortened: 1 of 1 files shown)


wjwwood:
+1

PR #519, opened 2013-09-30
Support for spaces in path

raphendyr:
There is very commonly spaces in ros workspace path...

EDIT: Fixed setup.bash and setup.zsh too.

### DIFF
--- a/cmake/templates/env.sh.in
+++ b/cmake/templates/env.sh.in
@@ -8,6 +8,6 @@ if [ $# -eq 0 ] ; then
 fi
 
 # source @SETUP_FILENAME@.sh from same directory as this file
-_CATKIN_SETUP_DIR=$(cd `dirname $0`;pwd)
+_CATKIN_SETUP_DIR=$(dirname "$0")
 . "$_CATKIN_SETUP_DIR/@SETUP_FILENAME@.sh"
 exec "$@"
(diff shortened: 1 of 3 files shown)


dirk-thomas:
The usage of `dirname` only is not sufficient. I will commit a working patch soon.

raphendyr:
Thanks for the better implementation.

ps. what's the reason to clean up path using pwd?

dirk-thomas:
Try you code in e.g. `/tmp/test.sh` and source it from `/` using `. tmp/test.sh`. `dirname` will return nothing which is just wrong. The modified version always returns an absolute path.

raphendyr:
Hmmm... ineresting... I got these results:

``` sh
/ $ sh tmp/foo\ bar/test.sh 
/tmp/foo bar
tmp/foo bar
tmp/foo bar
/ $ dash tmp/foo\ bar/test.sh 
/tmp/foo bar
tmp/foo bar
tmp/foo bar
/ $ bash tmp/foo\ bar/test.sh 
/tmp/foo bar
tmp/foo bar
tmp/foo bar
/ $ zsh tmp/foo\ bar/test.sh 
/tmp/foo bar
tmp/foo bar
tmp/foo bar
/ $ cat tmp/foo\ bar/test.sh
#!/bin/sh

echo $(cd "`dirname "$0"`" && pwd)
echo $(dirname "$0")
echo ${0%/*}
```

ps. The last line could also be used?

dirk-thomas:
You must `source` the shell script and not invoke it in order to see the difference. Run `source tmp/foo\ bar/test.sh`.

raphendyr:
Well. Can't reproduce, but never mind. Not my problem, it works.

PR #521, opened 2013-10-01
fix env.sh for plain cmake packages to not define _CATKIN_SETUP_DIR

dirk-thomas:
This was breaking isolated builds by not passing along the environment of the packages _before_ the plain CMake package. By setting `_CATKIN_SETUP_DIR` in the plain CMake env.sh it breaks the parent setup.sh since there is no _setup_util.py in that folder.

@wjwwood Please review.

### DIFF
--- a/python/catkin/builder.py
+++ b/python/catkin/builder.py
@@ -463,7 +463,20 @@ def build_cmake_package(
             'SETUP_FILENAME': 'setup'
         }
         with open(os.path.join(new_env_path), 'w') as f:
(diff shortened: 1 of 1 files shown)


wjwwood:
@scpeters also

wjwwood:
Looks good to me. +1

scpeters:
Testing it now, though it will be a few minutes for gazebo to build.

scpeters:
+1, thanks!

Issue #518, opened 2013-09-26
python2.6 compatibility errors


tkruse:
Not sure if anyone still cares about python2.6 (default with lucid), but it seems in few places catkinb code uses python2.7 features:

```
File "src/catkin/bin/catkin_make", line 102
  packages_by_name = {p.name: path for path, p in packages.iteritems()}
```

and

```
File "src/catkin/bin/catkin_make", line 15, in <module>
  from catkin.builder import cmake_input_changed
File "src/catkin/bin/../python/catkin/builder.py", line 698
  packages_by_name = {p.name: path for path, p in packages.iteritems()}
```

and

```
File "/pr/1/tkruse/work/groovy/build/catkin_generated/generate_cached_setup.py", line 21, in <module>
  code = generate_environment_script('/pr/1/tkruse/work/groovy/devel/env.sh')
File "/pr/1/tkruse/work/groovy/src/catkin/cmake/../python/catkin/environment_cache.py", line 61, in generate_environment_script
  output = subprocess.check_output([env_script, sys.executable, '-c', python_code])
AttributeError: 'module' object has no attribute 'check_output'
```

IMHO, staying 2.6 compatible is worth more than saving 2-3 lines of code for now.

dirk-thomas:
Since the current development branch of catkin does only work for Groovy and Hydro and those distributions do neither support Lucid nor Python 2.6 (see http://ros.org/reps/rep-0003.html#groovy-galapagos-oct-2012) I don't think it is worth keeping Python 2.6 compatibility.

Especially with the goal of making the whole code base compatible with Python 3.x for the next ROS distribtion it would be even more effort to also keep 2.6 compatibility.

tkruse:
Indeed maybe it is silly to think of catkin as useful software outside the ROS context.
ROS users trying older platforms still can patch catkin themselves if need be.

Issue #513, opened 2013-09-23
catkin_prepare_release) Handle package paths beneath trunk in svn repository


130s:
Same as asked [here](http://answers.ros.org/question/70054/catkin_prepare_release-fails-with-svn-upstream/), our team stores multiple packages in a single `svn` repository, by placing them BENEATH `/trunk` folder, which current `catkin_prepare_release` returns error with.

What about adding an option (eg. `--pkg`) to accept those packages under `/trunk`?

**Usecase-1**: releasing a package directly beneath /trunk

```
URL of the package: `https://foo.googlecode.com/svn/trunk/bar`
Version current: 1.0.0
```

then:

```
$ catkin_prepare_release --pkg bar
```

yields:

```
tag: `https://foo.googlecode.com/svn/tag/bar-1.0.1`
version 1.0.1
```

(Apparently this can utilize the same logic with `--tag-prefix` option internally)

**Usecase-2**: releasing a package at deeper level

```
URL of the package: `https://foo.googlecode.com/svn/trunk/bar/baz`
Version current: 1.0.0
```

then:

```
$ catkin_prepare_release --pkg baz
```

yields:

```
tag: `https://foo.googlecode.com/svn/tag/baz-1.0.1`
version 1.0.1
```

---

There maybe insufficiency in the usecases but hope you get the basic idea of this suggestion. 

What I am also not clear about `catkin_prepare_release` is what other situations need considered; for example, `trunk` isn't the only suffix that's inspected [in tag_repository](https://github.com/ros/catkin/blob/1b2366e1cee733e6ffab842202098025d11189fe/bin/catkin_prepare_release#L121) but do we need same mechanism that I suggested above for `branch`, `tag`?

If this is good and we get to consensus I can try making a patch (if it doesn't get too complicated).

Again I honestly don't know if this is a valid enhancement with respect to svn convention. However, it'll be very helpful in our release workflow (at JSK lab). Also I just guess that there could be developers who might not strictly follow rules in detail.

...gets so long. Thanks for finishing until the end!

dirk-thomas:
You might consider restructuring your SVN repository so that each unit of release resides in its own trunk/brnaches/tags subfolder. The layout you currently use is not well supported by any of the tools in the chain and requires additional manual intervention which is otherwise not necessary. Just simply because it is not automatically possible to determine what the users layout is and what he wants.

[...]

PR #522, opened 2013-10-02
support svn repository

k-okada:
I'm not sure if I can follow #518 (https://github.com/ros/catkin/issues/513) discussion precisely, but hits what I wrote for using catkin_prepare_release for svn repository. 
c91b875 update "bump version number code" section for new commit_files() and tag_repository() files and 8a1fb7 uses tag_name ( = tag_prefix + new_version ) for commit log, instead of using just a new_version

68364c1 changes behavior, but search trunk/branch/tags from base_url rather than assuming trunk/branches/tags is end of base_url. This can handle more flexible svn repository directory structure.


dirk-thomas:
Next time you can just commit changes to the branch you used for this pull request (even with rewriting history and force pushes). No need to create multiple new pull requests.

dirk-thomas:

PR #523, opened 2013-10-03
fix to work with svn, 

k-okada:
now commit_files() and tags_repository() returns commands although it node used except svn, see https://github.com/k-okada/catkin/commit/c91b8754a90b16aac395058b3f247a0bc0716750#commitcomment-4234735


dirk-thomas:

PR #524, opened 2013-10-03
Use tagname

k-okada:
see https://github.com/k-okada/catkin/commit/8a1f8b7111332b836d82c5b83792b6719b143240#commitcomment-4234687

### DIFF
--- a/bin/catkin_prepare_release
+++ b/bin/catkin_prepare_release
@@ -97,7 +97,7 @@ def check_clean_working_copy(base_path, vcs_type):
 
 
 def commit_files(base_path, vcs_type, packages, packages_with_changelogs, message, dry_run=False):
-    cmd = [_find_executable(vcs_type), 'commit', '-m', message]
+    cmd = [_find_executable(vcs_type), 'commit', '-m', '"%s"' % message]
     cmd += [os.path.join(p, PACKAGE_MANIFEST_FILENAME) for p in packages.keys()]
     cmd += [path for path, _, _ in packages_with_changelogs.values()]
     if not dry_run:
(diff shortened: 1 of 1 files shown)


bin/catkin_prepare_release
```diff
@@ -300,7 +300,7 @@ def main():
     if vcs_type in ['svn']:
         # for svn everything affects the remote repository immediately
         commands = []
-        commands.append(commit_files(base_path, vcs_type, packages, missing_changelogs_but_forthcoming, new_version, dry_run=True))
+        commands.append(commit_files(base_path, vcs_type, packages, missing_changelogs_but_forthcoming, tag_name, dry_run=True))
```
dirk-thomas, line 303: The same fix would be necessary in line 310 and 316.

dirk-thomas:

bin/catkin_prepare_release
```diff
@@ -97,7 +97,7 @@ def check_clean_working_copy(base_path, vcs_type):
 
 
 def commit_files(base_path, vcs_type, packages, packages_with_changelogs, message, dry_run=False):
-    cmd = [_find_executable(vcs_type), 'commit', '-m', message]
+    cmd = [_find_executable(vcs_type), 'commit', '-m', '"%s"' % message]
```
dirk-thomas, line 100: This was a pretty bad idea since commit messages are now surrounded by quotes. Can you please describe why you introduced this in the first place?

dirk-thomas, line 100: I rolled back the additional quoting: a889894e9c7acccca931289de888e3dba1537504

k-okada, line 100: sorry for slow reply,
Originally, my patch calls commit_files() with string surrounded by quotes, so now it not required any more, I think.
https://github.com/k-okada/catkin/commit/8a1f8b7111332b836d82c5b83792b6719b143240

PR #525, opened 2013-10-03
search trunk/branches/tags from base_url

k-okada:
see https://github.com/k-okada/catkin/commit/68364c1fdea5c5a33c7628be74de391d27debe69#commitcomment-4234427


dirk-thomas:
This will not work that way. Both scenarios are valid: having "/trunk" at the end of the url as well as "/trunk/" somewhere in the middle. I will provide a modified pull request soon.

dirk-thomas:

PR #527, opened 2013-10-03
improve svn support in catkin_prepare_release

dirk-thomas:
The commit message should contain the tag name - not only the version number.

SVN should check earlier if the tagging structure is supported (or not) to not do changes if it will fail later anyway.

Besides support for tagging trunk or a specific branch:
-  allow tagging arbitrary subfolders but then require a tag name prefix

@k-okada Please check if that version suits your needs by trying to do some test releases from your svn repos.


tfoote:
+1

wjwwood:
+1

k-okada:
Thanks, https://github.com/ros/catkin/commit/60a49f9062d218069c516ba4a9dc8fae0a30b0c4 works ok for me

PR #528, opened 2013-10-04
(catkin_prepare_release) Prompt tag name when it differs from new version number.

130s:
Old title: (catkin_prepare_release) Add hyphen b/w tag and the new version number for svn.

### DIFF
--- a/bin/catkin_prepare_release
+++ b/bin/catkin_prepare_release
@@ -251,8 +251,15 @@ def main():
     old_version = verify_equal_package_versions(packages.values())
     new_version = bump_version(old_version, args.bump)
     tag_name = args.tag_prefix + new_version
+    
+    # When tag_name differs from new_version, create a string to show
+    # the whole tag name in addition to version (eg. In the case discussed in
+    # https://github.com/ros/catkin/pull/528#issuecomment-25739909).
+    promptmsg_tag_name = ''
+    if tag_name is not new_version:
+        promptmsg_tag_name = fmt(" (tag name '@{bf}@{boldon}%s@{boldoff}@{reset}') " % tag_name)
 
-    if not args.non_interactive and not prompt_continue(fmt("Prepare release of version '@{bf}@{boldon}%s@{boldoff}@{reset}'" % new_version), default=True):
+    if not args.non_interactive and not prompt_continue(fmt("Prepare release of version '@{bf}@{boldon}%s@{boldoff}@{reset}'" % new_version) + promptmsg_tag_name, default=True):
         raise RuntimeError(fmt("@{rf}Aborted release, use option '--bump' to release a different version."))
 
     # check for changelog entries


130s:
Forgot to describe what this PR does. Notice the hyphen between the `tag` and the new version number:

now
    svn cp -m "tagging agentsystem_ros_tutorials1.0.2" https://rtm-ros-robotics.googlecode.com/svn/trunk/agentsystem_ros_tutorials https://rtm-ros-robotics.googlecode.com/svn/tags/agentsystem_ros_tutorials1.0.2

new
    svn cp -m "tagging agentsystem_ros_tutorials-1.0.2" https://rtm-ros-robotics.googlecode.com/svn/trunk/agentsystem_ros_tutorials https://rtm-ros-robotics.googlecode.com/svn/tags/agentsystem_ros_tutorials-1.0.2

```
```

k-okada:
+1
That's good for me

dirk-thomas:
The user can just specify a tag prefix with a trailing hyphen. I would rather not force him to have that hypen since else a tag like `foo1.2.3` becomes impossible to do.

130s:
That makes sense. Then how about showing the tag that will be generated
before bumping the version do that users can terminate the command when
they don't like it? If that's good then I'll update my PR or open a new one.

dirk-thomas:
You could extend the question `Prepare release of version ...` to contain the tag name if it is different from the version. That would be good to add.

130s:
@dirk-thomas Please review my updated pull request, with which the result would look like:

```
$ cd agentsystem_ros_tutorials
$ catkin_prepare_release
Prepare the source repository for a release.
Repository type: svn
Found packages: agentsystem_ros_tutorials, move_base_stage_tutorial
Prepare release of version '1.0.2' [Y/n]?n
$ catkin_prepare_release -t agentsystem_ros_tutorials
Prepare the source repository for a release.
Repository type: svn
Found packages: agentsystem_ros_tutorials, move_base_stage_tutorial
Prepare release of version '1.0.2' (tag name 'agentsystem_ros_tutorials1.0.2')  [Y/n]?n
```

(The title of PR is also updated.)

130s:
Dirk, wouldn't your commit return something like:

```
$ catkin_prepare_release -t foo
Prepare release of version 'foo1.0.2' [Y/n]?
```

Precisely speaking, `foo1.0.2` is a tag name, not a version. Because in my experience releasing needs extra amount of care, I would rather separately display version and tag name in order to avoid users' confusion as much as possible.

dirk-thomas:
Yes, that would be displayed. This is the name the developer wants to give to that release. Since you are potentially tagging a part of the repo the version does not only the number since that would not be sufficient to identify the release.

dirk-thomas:
Would you prefer this?

```
if not args.non_interactive and not prompt_continue(fmt("Prepare release of version '@{bf}@{boldon}%s@{boldoff}@{reset}'%s" % (new_version, " (tagged as '@{bf}@{boldon}%s@{boldoff}@{reset}')" % tag_name if args.tag_prefix else '')), default=True):
```

130s:
Looks nice. Works great like this:

```
$ cd foo
$ catkin_prepare_release 
Prepare the source repository for a release.
Repository type: svn
Found packages: foo
Prepare release of version '1.0.2' [Y/n]?n
Aborted release, use option '--bump' to release a different version.
$ catkin_prepare_release  -t foo
Prepare the source repository for a release.
Repository type: svn
Found packages: foo
Prepare release of version '1.0.2' (tagged as 'foo1.0.2') [Y/n]?n
Aborted release, use option '--bump' to release a different version.
```

130s:
Thanks Dirk, this tremendously lowers our mental barrier for releasing our svn packages!

PR #531, opened 2013-10-09
catkin_make_isolated: fix setup generation when plain cmake package is last in workspace (fix #530)

### LINKED ISSUE #530: Isolated build of plain cmake package has problem with setup scripts
Several files are missing from the devel_isolated folder after building a plain cmake package on OSX. Build sdformat for example:

```
mkdir -p ~/ws/src
cd ~/ws/src
git clone https://github.com/ros/catkin.git
hg clone https://bitbucket.org/osrf/sdformat -u package_xml
cd ~/ws
./src/catkin/bin/catkin_make_isolated
```

Then the catkin package will have some setup files that sdformat does not:

```
$ ls ~/ws/devel_isolated/catkin
_setup_util.py env.sh         lib            setup.sh       share
bin            etc            setup.bash     setup.zsh
```

```
$ ls ~/ws/devel_isolated/sdformat
env.sh   include  lib      setup.sh share
```

The missing files are a problem because `devel_isolated/setup.bash` tries to reference them:

```
$ cat devel_isolated/setup.bash 
#!/usr/bin/env bash
# generated from catkin.builder module

. "~/ws/devel_isolated/sdformat/setup.bash"
$
```

It won't work properly if I source `devel_isolated/setup.bash`; I have to source `devel/setup.sh` instead.

I also tested building a single catkin package (genmsg) and it did not have this problem.

dirk-thomas:
First, it generates setup.bash|zsh script for plain cmake package (to make the next step easier).
Second, the setup.sh for plain packages now relays to the parent setup with the currently selected CATKIN_SHELL.

@wjwwood @scpeters Please review

### DIFF
--- a/python/catkin/builder.py
+++ b/python/catkin/builder.py
@@ -509,9 +509,13 @@ exec "$@"
 #!/usr/bin/env sh
 # generated from catkin.builder module
 
+# remember type of shell if not already set
+if [ -z "$CATKIN_SHELL" ]; then
+  CATKIN_SHELL=sh
+fi
 """)
             if last_env is not None:
-                file_handle.write('. %s\n\n' % last_setup_env)
+                file_handle.write('. %s.$CATKIN_SHELL\n\n' % last_setup_env[:-3])
             file_handle.write("""\
 # detect if running on Darwin platform
 _UNAME=`uname -s`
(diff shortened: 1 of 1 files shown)


wjwwood:
+1

python/catkin/builder.py
```diff
@@ -509,9 +509,13 @@ def build_cmake_package(
...
 # generated from catkin.builder module
 
+# remember type of shell if not already set
+if [ -z "$CATKIN_SHELL" ]; then
+  CATKIN_SHELL=sh
+fi
 """)
             if last_env is not None:
-                file_handle.write('. %s\n\n' % last_setup_env)
+                file_handle.write('. %s.$CATKIN_SHELL\n\n' % last_setup_env[:-3])
```
tfoote, line 518: Is this guaranteed to always be sh? not bash?

dirk-thomas, line 518: Yes, env is always `.sh`.

PR #536, opened 2013-10-17
catkin_FOUND variable was never set

meyerj:
Shouldn't this be

```
  set(catkin_FOUND 1)
```

instead of

```
  set(catkin_FOUND)
```

?

From http://www.cmake.org/cmake/help/v2.8.3/cmake.html#command:set:

> If <value> is not specified then the variable is removed instead of set. See also: the unset() command.

Seems that `catkin_FOUND` was never set since the beginning of time...

### DIFF
--- a/cmake/catkinConfig.cmake.in
+++ b/cmake/catkinConfig.cmake.in
@@ -34,7 +34,9 @@ endif()
 
 # mark as found
 if(NOT catkin_FOUND)
-  set(catkin_FOUND)
+  # cmake will set catkin_FOUND automatically after this config file has been read
+  # but other cmake configuration files of packages mentioned in the COMPONENTS list might need it already
+  set(catkin_FOUND True)
   set(CATKIN_PACKAGE_PREFIX "" CACHE STRING "Prefix to apply to package generated via gendebian")
 endif()
 


wjwwood:
Wow, that is an oversight, thanks for the catch.

I would use `True` rather than `1` though.

dirk-thomas:
Have you actually tried to run and did it returned "false" for you:

```
find_package(catkin REQUIRED)
message("catkin_FOUND ${catkin_FOUND}")
if(catkin_FOUND)
    message("  true")
else()
    message("  false")
endif()
```

The line you pointed out is imo not a problem. It is a remnant of old catkin times. The `catkin_FOUND` variable is implicitly set when `find_package(catkin)` is invoked and the `catkinConfig.cmake` file is found (the same as e.g. `catkin_DIR`). So it works fine as it is - at least for me - and as it looks like also for you: https://github.com/orocos-toolchain/rtt/pull/6#issuecomment-26547810

wjwwood:
@dirk-thomas is right, CMake is actually setting `catkin_FOUND` when it successfully finds the config file (`catkinConfig.cmake`). You only have to manually set the `_FOUND` variable when writing a module (which would be `Findcatkin.cmake`), which we don't. Though the line you pointed out could probably be removed.

meyerj:
Hmm, at least I had a package that did not compile without this patch. After having applied this patch to `/opt/ros/hydro/share/catkin/cmake/catkinConfig.cmake` manually it did, in a clean build environment and with direct invocation of cmake and make (without catkin_make). I will try your test snippet tomorrow and I will also try to remove this line completely.

Perhaps this one bites us:

> The package configuration file may set <package>_FOUND to false to tell find_package that component requirements are not satisfied.
> (from http://www.cmake.org/cmake/help/v2.8.12/cmake.html#command:find_package)

meyerj:
I think the problem is that a `CFG_EXTRAS` file in package [rtt_ros](https://github.com/jhu-lcsr-forks/rtt_ros_integration/tree/9601e99a4c45b449404cd198ad7dd19f60dc2dbc/rtt_ros) includes the `UseOROCOS-RTT.cmake` script, that checks `catkin_FOUND` to decide which output directories should be used. If `catkin_FOUND` is already set to True in `catkinConfig.cmake`, everything is fine. Otherwise `catkin_FOUND` will be set by cmake **after** all configs for packages mentioned in `COMPONENTS` and their `CFG_EXTRAS` have been parsed and `UseOROCOS-RTT.cmake` will see it as not set.

Can we assume that the [rtt_ros-extras.cmake](https://github.com/jhu-lcsr-forks/rtt_ros_integration/blob/9601e99a4c45b449404cd198ad7dd19f60dc2dbc/rtt_ros/cmake/rtt_ros-extras.cmake.in) file is only used in catkin environments or does e.g. rosbuild also uses it under the hood if a package depends on rtt_ros? In the first case we could add `find_package(catkin REQUIRED)` or `set(catkin_FOUND True)` to rtt_ros-extras.cmake.

dirk-thomas:
The extra files are used the same way when (recursively) being found from a dry package.

But the `rtt_ros` package already depends on catkin. Why not call `find_package(catkin REQUIRED)` in the extra file? That should define the boolean flag and it should not matter if it is used later from a wet or dry package.

I will close the pull request since this change should not be done in catkin itself. catkin (or better cmake) already defines the flag perfectly well.

meyerj:
Just to answer your question: The `UseOROCOS-RTT.cmake` script tries to figure out whether it is included from a rosbuild or a catkin package or none of these two in order to setup output directories and other settings accordingly.

So what we currently do is

``` cmake
if(COMMAND rosbuild_init AND ROSBUILD_init_called)
  # assume rosbuild...
  # Note: catkin_FOUND is also True!
  # ...
elseif(catkin_FOUND)
  # assume catkin
  # ...
else()
  # anything else
  # ...
endif()
```

and this script is included from the `rtt_ros` env-hook. I just wasn't sure if `rosbuild_init()` also includes the env-hook internally or if it only uses the pkg-config files to get the flags for wet packages. But even if yes, the test should be okay as rosbuild is tested first. We also added `set(catkin_FOUND True)` to the env-hook explicitly in the meantime and there haven't been any problems anymore since then. Without that line catkin was not recognized properly as `catkin_FOUND` might not have been set yet.

I would still consider this as a (minor) bug in catkin as long as `catkin_FOUND` is not set to `True` or the line is removed as suggested by @wjwwood.

PR #538, opened 2013-10-17
update toplevel.cmake to use catkin_find_pkg instead of trying a fixed location to find catkin in the workspace

dirk-thomas:

### DIFF
--- a/cmake/toplevel.cmake
+++ b/cmake/toplevel.cmake
@@ -5,12 +5,27 @@ cmake_minimum_required(VERSION 2.8.3)
 
 set(CATKIN_TOPLEVEL TRUE)
 
-# include catkin directly or via find_package()
-if(EXISTS "${CMAKE_SOURCE_DIR}/catkin/cmake/all.cmake" AND EXISTS "${CMAKE_SOURCE_DIR}/catkin/CMakeLists.txt")
-  set(catkin_EXTRAS_DIR "${CMAKE_SOURCE_DIR}/catkin/cmake")
+# search for catkin within the workspace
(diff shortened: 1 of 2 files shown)


cmake/toplevel.cmake
```diff
@@ -5,12 +5,27 @@ cmake_minimum_required(VERSION 2.8.3)
...
+)
+if(NOT _res EQUAL 0 AND NOT _res EQUAL 2)
+  # searching fot catkin resulted in an error
+  string(REPLACE ";" " " _cmd_str "${_cmd}")
+  message(FATAL_ERROR "Search for 'catkin' in workspace failed (${_cmd_str}): ${_err}")
+endif()
+
+# include catkin from workspace or via find_package()
+if(_res EQUAL 0)
+  set(catkin_EXTRAS_DIR "${CMAKE_SOURCE_DIR}/${_out}/cmake")
```
wjwwood, line 25: Should you check to see if this resulting directory exists?

wjwwood:
Otherwise +1

cmake/toplevel.cmake
dirk-thomas, line 25: I thought about what should happen then.

The scenario would be that something which looks like catkin is in the workspace does not behave like catkin. I would say it should be ok if the following include will fail?

wjwwood, line 25: That's fine.

tfoote:
+1

dirk-thomas:
catkin_pkg 0.1.21 has been released (ros-infrastructure/catkin_pkg@6de9e90727311f68e34cdfdc2f3a3d6a55855f9f) and is available in the public repos by now.

PR #544, opened 2013-10-23
python 3 compatibility

dirk-thomas:

### DIFF
--- a/python/catkin/builder.py
+++ b/python/catkin/builder.py
@@ -186,7 +186,11 @@ def run_command(cmd, cwd, quiet=False, colorize=False, add_env=None):
     out = io.StringIO() if quiet else sys.stdout
     if capture:
         while True:
-            line = unicode(proc.stdout.readline().decode('utf8', 'replace'))
+            line = proc.stdout.readline().decode('utf8', 'replace')
+            try:
+                line = unicode(line)
+            except NameError:
+                pass
             if proc.returncode is not None or not line:
                 break
             try:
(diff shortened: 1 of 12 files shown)


python/catkin/builder.py
```diff
@@ -186,7 +186,11 @@ def run_command(cmd, cwd, quiet=False, colorize=False, add_env=None):
     out = io.StringIO() if quiet else sys.stdout
     if capture:
         while True:
-            line = unicode(proc.stdout.readline().decode('utf8', 'replace'))
+            line = proc.stdout.readline().decode('utf8', 'replace')
+            try:
+                line = unicode(line)
+            except NameError:
```
tfoote, line 193: This is awkward, It might be worth a comment that this is for python2 backwards compatability so that sometime in the future we would come back and clear all backwards compatibility things out.  

Maybe we should be flagging everything like that with a unique string for ease of finding.

test/unit_tests/test_builder.py
```diff
@@ -23,7 +23,10 @@ def __init__(self, popen):
 
             def readline(self):
                 self.__popen.returncode = 0
-                return unichr(2018)
+                try:
+                    return unichr(2018)
+                except NameError:
```
tfoote, line 29: ditto above

tfoote:
+1 pending comments

python/catkin/builder.py
dirk-thomas, line 193: Added note in 0436f6c.

test/unit_tests/test_builder.py
dirk-thomas, line 29: Added note in 0436f6c.

tfoote:
+1

wjwwood:
+1

PR #546, opened 2013-10-25
fix order of catkin_LIBRARIES

dirk-thomas:
Instead of not adding duplicate library names at the end the libraries should always be appended to maintain correct order (especially valid when using static libraries). Before appending them existing duplicates are being removed from the list.

As long as we have no circular dependencies between libraries this should work fine. Since packages already forbid such a dependency cycle that should never be the case.

### DIFF
--- a/cmake/all.cmake
+++ b/cmake/all.cmake
@@ -112,6 +112,7 @@ foreach(filename
     empy
     find_program_required
     legacy
+    list_append_deduplicate
     list_append_unique
     list_insert_in_workspace_order
     parse_arguments
--- a/cmake/catkinConfig.cmake.in
+++ b/cmake/catkinConfig.cmake.in
@@ -81,7 +81,7 @@ foreach(component ${catkin_FIND_COMPONENTS})
 
     # append component-specific variables to catkin_* variables
     list_append_unique(catkin_INCLUDE_DIRS ${${component}_INCLUDE_DIRS})
-    list_append_unique(catkin_LIBRARIES ${${component}_LIBRARIES})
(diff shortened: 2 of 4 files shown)


tfoote:
One question on the disallowed circular dependencies. That cannot happen between packages, but what about within a package?

dirk-thomas:
Imo you are doing something wrong when your libraries have circular dependencies. E.g. you can create static libraries that way.

wjwwood:
> E.g. you can create static libraries that way.

can == can't?

dirk-thomas:
Ups, yes, can't. I am pretty sure it will be chicken-egg-problem which one to link first.

mihelich:
It is possible to link static libraries with circular dependencies, by listing libraries more than once or using special linker flags: http://stackoverflow.com/questions/9380363/resolving-circular-dependencies-by-linking-the-same-library-twice. I guess this is sometimes useful in large legacy codebases.

I'd agree with Dirk though. Catkin already forbids circular dependencies between packages, and a single package should not be so large and complex that circular dependencies become an intractable issue.

dirk-thomas:
I was able to perform a hydro-desktop build from source with this catkin branch - so chances for a regression should be minimal (hopefully).

@tfoote @wjwwood Any votes on the current version of the pull request?

wjwwood:
Sounds good to me.

tfoote:
Is list_append_unique still used in the system? If not we should clear it out. 

Otherwise +1

dirk-thomas:
Still used. One line above/below the new function.

I will go ahead and merge.

PR #550, opened 2013-11-15
Fix to actually build run_depends before the things that depend on them with catkin_make_isolated

eitanme:
This is the quick fix I made to be able to run `catkin_make_isolated` with the `--only-pkg-with-deps` argument for a source build of ROS. I'm not sure if changing the behavior of `get_package_names_with_recursive_dependencies` to also recurse to run_depends is kosher. If so, this patch should fix things. If not, `--only-pkg-with-deps` probably needs to call another function that does.

Thoughts?

### DIFF
--- a/python/catkin/builder.py
+++ b/python/catkin/builder.py
@@ -931,7 +931,7 @@ def get_package_names_with_recursive_dependencies(packages, pkg_names):
         if pkg_name in packages_by_name:
             pkg = packages_by_name[pkg_name]
             dependencies.add(pkg_name)
-            for dep in [dep.name for dep in (pkg.build_depends + pkg.buildtool_depends)]:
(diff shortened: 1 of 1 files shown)


eitanme:
I was running in a custom env for a client that I can't share, but you can probably reproduce this issue with the bare bones install. Something like:

```
rosinstall_generator ros_comm --rosdistro hydro --deps --wet-only > hydro-ros_comm-wet.rosinstall
wstool init -j8 src hydro-ros_comm-wet.rosinstall
./src/catkin/bin/catkin_make_isolated --merge --only-pkg-with-deps=roscpp

```

dirk-thomas:
For making the build work it would be sufficient to use the first-level build dependencies and recursive build/run dependencies. But the users expectation will be that he is able to run the stuff from the built package. So we should include the run dependencies of the package requested too.

So, yes, your patch fixes that problem well. Thanks for the contribution.

PR #560, opened 2013-12-12
Update download_checkmd5.py

### LINKED ISSUE #559: download_checkmd5 ContentTooShortError
Hi

I'm using download_checkmd5.py to download a 200MB file from my server, and it always seems to fail with a ContentTooShortError. My server is quite reliable though: using wget I have no issues at all downloading the file.

I tried using urllib2 but it's the same.

```
def dlfile(url, dest):
    # Open the url
    f = urllib2.urlopen(url)

    # Open our local file for writing
    with open(dest, "wb") as local_file:
        local_file.write(f.read())
```

I am now working on a solution using wget...

any idea?

Brice

bricerebsamen:
Created a new urlretrieve function that handles retrying if the connection drops. Fixes #559.

### DIFF
--- a/cmake/test/download_checkmd5.py
+++ b/cmake/test/download_checkmd5.py
@@ -227,8 +227,8 @@ def download_md5(uri, dest):
     sys.stdout.flush()
 
     downloader = FileDownloader()
-    downloader.progressFn = progress_report
-    downloader.verbose = True
+    #downloader.progressFn = progress_report
+    #downloader.verbose = True
     downloader.download_file(uri, dest)
     sys.stdout.write('Done\n')
 
--- a/cmake/test/download_checkmd5.py
+++ b/cmake/test/download_checkmd5.py
@@ -4,9 +4,9 @@ from __future__ import print_function
 import os
 import sys
 try:
-    from urllib2 import urlopen, BaseHandler, build_opener, Request
+    from urllib2 import urlopen, BaseHandler, build_opener, Request, addinfourl
 except ImportError:
-    from urllib.request import urlopen, BaseHandler, build_opener, Request
+    from urllib.request import urlopen, BaseHandler, build_opener, Request, addinfourl
 import time
 import hashlib
 from optparse import OptionParser
@@ -23,7 +23,7 @@ class HTTPRangeHandler(BaseHandler):
 
     def http_error_206(self, req, fp, code, msg, hdrs):
         # 206 Partial Content Response
-        r = urllib.addinfourl(fp, hdrs, req.get_full_url())
+        r = addinfourl(fp, hdrs, req.get_full_url())
         r.code = code
         r.msg = msg
         return r
@@ -71,6 +71,8 @@ def download_file(url, localfile=None, nRetry=20, append=False, progressFn=None)
     if not append and os.path.exists(localfile) and os.path.isfile(localfile):
         os.remove(localfile)
 
+    lastProgressReportPercent = None
+    
     range_handler = HTTPRangeHandler()
     opener = build_opener(range_handler)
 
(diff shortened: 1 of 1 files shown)


cmake/test/download_checkmd5.py
```diff
@@ -3,16 +3,132 @@
 from __future__ import print_function
 import os
 import sys
-try:
-    from urllib.request import urlretrieve
-except ImportError:
-    from urllib import urlretrieve
+import urllib2
```
dirk-thomas: The code must be Python 2 and 3 compatible. Please keep using the conditional import and make sure the rest of the code works with Python 3.

cmake/test/download_checkmd5.py
```diff
@@ -4,15 +4,133 @@
...
+# Support retry on dropped connection: #559
+class HTTPRangeHandler(BaseHandler):
+    """
+    handler that enables HTTP Range headers.
+    """
+    # Taken from urlgrabber: http://urlgrabber.baseurl.org
+
+    def http_error_206(self, req, fp, code, msg, hdrs):
+        # 206 Partial Content Response
+        r = urllib.addinfourl(fp, hdrs, req.get_full_url())
```
dirk-thomas: `urllib` is not imported.

cmake/test/download_checkmd5.py
```diff
@@ -4,15 +4,133 @@
...
+                _outfile.close()
+                return
+
+        # we reach this point if the read returned no bytes. In which case we
+        # want to retry
+        _netfile.close()
+        _outfile.close()
+
+        # is it necessary ???
+        time.sleep(1)
```
dirk-thomas: This will let the function take at least 20s if the download fails. Is there a specific reason to have this?

cmake/test/download_checkmd5.py
```diff
@@ -4,15 +4,133 @@
...
+        chunk_size = 8 * 1024
+        while data is None or len(data)>0:
+            data = _netfile.read(chunk_size)
+            _outfile.write(data)
+            count += len(data)
+
+            if progressFn:
+                progressFn(count, filesize)
+
+            if count >= filesize:
```
dirk-thomas: Is `count > filesize` not an error?

cmake/test/download_checkmd5.py
```diff
@@ -4,15 +4,133 @@
...
+        _outfile = open(localfile, "ab")
+        data = None
+        chunk_size = 8 * 1024
+        while data is None or len(data)>0:
+            data = _netfile.read(chunk_size)
+            _outfile.write(data)
+            count += len(data)
+
+            if progressFn:
+                progressFn(count, filesize)
```
dirk-thomas: Calling the progress function for every chunk (8k) might generate a lot of progress calls for larger downloads.

dirk-thomas:
I tried using the code from this pull request and ran into several issue. The first thing is that it currently fails the unit tests since it does not support file urls anymore. The second problem is that is assumes that the server provides the `Content-Length` header. For some servers that might not be true so the script needs to check for that and should also validate that the response contains the corresponding `Content-Range` header. I also think that the retry and resume logic should not be tied to the same counter. We should only retry a low number of times (e.g. 3) but resume should be possible for an arbitrary amount as long as there is progress.

I can look into this but am not sure when I will find time to implement all this cleanly.

bricerebsamen:
Thanks for reviewing. I'll work on it tomorrow.

bricerebsamen:
I think I addressed all your comments. Please let me know if there is something wrong.

dirk-thomas:
The pull request if very lengthy and I tried to come up with a much conciser version of it. Please take a look at #575. It works fine for me (but I don't think that any of our download servers actually stops the download before it is complete). Can you please check it against the server which you had the original problem with to verify that it actually resumes the download correctly?

dirk-thomas:
Resolved by #575.

PR #564, opened 2013-12-17
add support for py_modules in setup.py (fix #399)

### LINKED ISSUE #399: catkin should support the python setup.py argument `py_modules`
When using this argument to setup.py's `setup()` function catkin warns:

```
*** Arguments ['py_modules'] to setup() not supported in catkin devel space in setup.py of camera_info_manager_py
```

It ges installed into the install space, but it does not create a redirection placeholder for the devel space.

Since this is a core functionality of python's distribution system we should support it.

Here is an example setup.py:

```
from distutils.core import setup
from catkin_pkg.python_setup import generate_distutils_setup

d = generate_distutils_setup(
    py_modules=['camera_info_manager'],
    package_dir={'': 'src'}
)

setup(**d)
```

dirk-thomas:
@tfoote @wjwwood Please review (and give it a try).

### DIFF
--- a/cmake/catkin_python_setup.cmake
+++ b/cmake/catkin_python_setup.cmake
@@ -103,6 +103,37 @@ function(catkin_python_setup)
     endforeach()
   endif()
 
+  # generate relay-script for each python module (and __init__.py files) if available
+  if(${PROJECT_NAME}_SETUP_PY_MODULES)
+    list(LENGTH ${PROJECT_NAME}_SETUP_PY_MODULES modules_count)
+    math(EXPR modules_range "${modules_count} - 1")
+    foreach(index RANGE ${modules_range})
+      list(GET ${PROJECT_NAME}_SETUP_PY_MODULES ${index} module)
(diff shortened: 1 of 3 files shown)


cmake/catkin_python_setup.cmake
```diff
@@ -103,6 +103,37 @@ function(catkin_python_setup)
     endforeach()
   endif()
 
+  # generate relay-script for each python module (and __init__.py files) if available
+  if(${PROJECT_NAME}_SETUP_PY_MODULES)
```
tfoote, line 107: Where does `${PROJECT_NAME}_SETUP_PY_MODULES` get set?

dirk-thomas, line 107: See below in file `cmake/interrogate_setup_dot_py.py` line 145 (https://github.com/ros/catkin/pull/564/files#diff-1a64629722ecefc0444795a8016e892eR145).

tfoote:
+1

wjwwood:
+1

PR #566, opened 2013-12-21
added atomic_configure_file function to catkin

wjwwood:
Also made catkin use it in a few places. This was done to prevent errors when building multiple catkin packages into the same devel space in parallel.

### DIFF
--- a/cmake/atomic_configure_file.cmake
+++ b/cmake/atomic_configure_file.cmake
@@ -1,0 +1,5 @@
+function(atomic_configure_file input output)
+  set(atomic_file "${CMAKE_BINARY_DIR}/atomic_configure_file")
+  configure_file("${input}" "${atomic_file}" ${ARGN})
+  file(RENAME "${atomic_file}" "${output}")
+endfunction()
--- a/cmake/catkin_add_env_hooks.cmake
+++ b/cmake/catkin_add_env_hooks.cmake
@@ -83,7 +83,7 @@ function(catkin_add_env_hooks file_prefix)
       else()
         set(in_template ${base}.in)
       endif()
-      configure_file(${in_template}
+      atomic_configure_file(${in_template}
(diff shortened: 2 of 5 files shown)


wjwwood:
This is related to #330.

cmake/catkin_add_env_hooks.cmake
```diff
@@ -46,6 +46,8 @@
 #
 # @public
 #
+include(${catkin_EXTRAS_DIR}/atomic_configure_file.cmake)
```
dirk-thomas: Instead of including the file in multiple location please add it to the global list of included files: https://github.com/ros/catkin/blob/groovy-devel/cmake/all.cmake#L100

cmake/atomic_configure_file.cmake
```diff
@@ -0,0 +1,9 @@
+function(atomic_configure_file input output)
+  set(_TEMP_CONFIGURE_DIR ${CMAKE_BINARY_DIR}/_temp_configure_dir)
```
dirk-thomas: The the function scope the variable doesn't need to be prefixed with `_` and could be shorter to.

The folder could be named `atomic_configure_file`.

cmake/atomic_configure_file.cmake
```diff
@@ -0,0 +1,9 @@
...
+  if(NOT EXISTS "${_TEMP_CONFIGURE_DIR}" AND NOT IS_DIRECTORY "${_TEMP_CONFIGURE_DIR}")
```
dirk-thomas: The `IS_DIRECTORY` check is not necessary. `EXISTS` already covers that.

wjwwood:
Ok, I addressed each of the concerns.

wjwwood:
Squashed.

dirk-thomas:
+1

PR #572, opened 2014-01-05
update handling of library deduplication to support build configuration keywords in the whole process (fix #565)

### LINKED ISSUE #565: catkinConfig.cmake should strip out 'debug' and 'optimized' attributes from catkin_LIBRARIES
I'm have a catkin package `A` which adds `debug;bar.so;optimized;bar.so` to the cmake variable `${A_LIBRARIES}`, and thus also adds to `${catkin_LIBRARIES}`:

I also have package `B` that has a build depends on `A`.  When `BConfig.cmake` is generated, the `debug` and `optimized` attributes are stipped out of `${B_LIBRARIES}`. This is done is [catkin/cmake/catkin_package.cmake:249-264](https://github.com/ros/catkin/blob/0.5.77/cmake/catkin_package.cmake#L249).

Suppose I add a catkin package `C` which depends on `A` and `B`, in that order.  In catkinConfig.cmake, `A_LIBRARIES` will be added to `catkin_LIBRARIES` with `list_append_deduplicate` at [line 84](https://github.com/ros/catkin/blob/0.5.77/cmake/catkin_config.cmake#L84), and then `B_LIBRARIES` will be added to `catkin_LIBRARIES` with `list_append_deduplicate`.  How the dedeuplicate call will strip out `bar.so` , leaving the words `debug` and `optimized` in `catkin_LIBRARIES`.

Thus, we get:
`catkin_LIBRARIES=debug;optimized;bar.so`

I'm assuming that the solution is for `catkinConfig.cmake` to do something similar to the debug/optimize stripping that's done in `catkin_package.cmake`.

**This is seen on catkin version 0.5.77**

dirk-thomas:

@wjwwood Can you please run a build with a big workspace on OS X with this branch? (I ran a successful build of (most of) desktop_full and looked at some generated CMake config files and they looked good.)


tfoote:
From a read through +1 

To confirm there are no other keywords we might encounter and want to support?

dirk-thomas:
`debug`, `optimized` and `general` are the only build configuration keywords (http://www.cmake.org/cmake/help/v2.8.12/cmake.html#command:target_link_libraries) and they should all be supported.

wjwwood:
+1

vpradeep07:
Thanks for making these changes.  I tested this branch on one of our codebases that has catkin packages that export `DEBUG` and `RELEASE` tags, and things seem to be working smoothly.

Looking forward to seeing these changes merged back in!

wjwwood:
Seems to work fine on my local machine, building PCL and `pcl_ros` which are known to use these library configuration keywords.

PR #575, opened 2014-01-07
update download script to resume downloads when server supports it (fix #559)

### LINKED ISSUE #559: download_checkmd5 ContentTooShortError

dirk-thomas:

### DIFF
--- a/cmake/test/download_checkmd5.py
+++ b/cmake/test/download_checkmd5.py
@@ -4,13 +4,81 @@ from __future__ import print_function
 import os
 import sys
 try:
-    from urllib.request import urlretrieve
+    from urllib.request import addinfourl, BaseHandler, build_opener, Request, URLError
 except ImportError:
-    from urllib import urlretrieve
+    from urllib2 import addinfourl, BaseHandler, build_opener, Request, URLError
 import hashlib
 from optparse import OptionParser
 
 NAME = "download_checkmd5.py"
+
+
+class HTTPRangeHandler(BaseHandler):
+
+    def http_error_206(self, req, fp, code, msg, hdrs):
+        r = addinfourl(fp, hdrs, req.get_full_url())
+        r.code = code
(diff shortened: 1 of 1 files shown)


cmake/test/download_checkmd5.py
```diff
@@ -4,15 +4,83 @@
...
+        if content_length is None:
+            break
+        # or when enough data has been downloaded (> is especially a valid case)
+        if offset >= content_length:
+            break
+        if not accept_ranges:
+            raise IOError('Server does not accept ranges to resume download')
+
+
 def download_md5(uri, dest):
```
tfoote, line 84: This whole second code branch could be avoided by using the above code block. The fetch of the md5sum file hopefully won't ever trigger a partial redownload, but the API is the same and this name poor anyway as it's not md5 specific at all.

dirk-thomas, line 84: This is not touched by the pull request. Simply because I did not want to change existing API.

But what avoidable "second code branch" are you referring to?

bricerebsamen:
I tested it with my server that was making problems and it's working now...
For reference, here is the file that I use for testing:
http://cs.stanford.edu/group/driving/data/ladybug/ladybug_intrinsics.dat
it's likely to stay at this location for a long time, so feel free to use it in the future if you need to make changes.

wjwwood:
+1

PR #577, opened 2014-01-08
Fix UnicodeEncodeError

yukota:
if fail to encode to utf8, unencoded charactor is written.


dirk-thomas:
Hiding the line from the output is not a good option. It should rather print unreadable / wrongly encoded characters as they are then ever hide anything.

PR #580, opened 2014-01-27
fix library deduplication (fix #565)

### LINKED ISSUE #565: catkinConfig.cmake should strip out 'debug' and 'optimized' attributes from catkin_LIBRARIES

dirk-thomas:
`pack()` / `unpack()` must be wrapped around every `list_append_deduplicate()` call since the input data might come from a recursive invocation which is not packed anymore.

### DIFF
--- a/cmake/catkinConfig.cmake.in
+++ b/cmake/catkinConfig.cmake.in
@@ -81,16 +81,19 @@ foreach(component ${catkin_FIND_COMPONENTS})
 
     # append component-specific variables to catkin_* variables
     list_append_unique(catkin_INCLUDE_DIRS ${${component}_INCLUDE_DIRS})
+
     # merge build configuration keywords with library names to correctly deduplicate
+    catkin_pack_libraries_with_build_configuration(catkin_LIBRARIES ${catkin_LIBRARIES})
     catkin_pack_libraries_with_build_configuration(_libraries ${${component}_LIBRARIES})
     list_append_deduplicate(catkin_LIBRARIES ${_libraries})
+    # undo build configuration keyword merging after deduplication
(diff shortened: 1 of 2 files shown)


tfoote:
+1

wjwwood:
+1, if it always has to be wrapped, did you consider just moving those pack and unpack calls into the deduplication macro?

dirk-thomas:
No, since the deduplication macro is not specific to libraries. It can work with arbitrary lists.

PR #589, opened 2014-02-04
remove restriction to run unit tests single threaded

dirk-thomas:
ros/ros_comm#349

### DIFF
--- a/bin/catkin_make
+++ b/bin/catkin_make
@@ -161,7 +161,7 @@ def main():
 
     # invoke make
     cmd = ['make']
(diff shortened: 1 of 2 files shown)


tfoote:
+1 following other release.

wjwwood:
I would leave it on for Windows, since the file locking in `rostest` will not work there. We could just not do locking in `rostest` on Windows and continue to enforce `-j1`, but only for Windows, or more specifically machines without `fcntl`.

dirk-thomas:
Closing in favor of #597. Might be considered for backport when next release into older distro is happening.

PR #590, opened 2014-02-07
use underlay workspaces when calculating topological order

dirk-thomas:
Depends on ros-infrastructure/catkin_pkg#87 being released as 0.1.26.
Should be backported to _Groovy_ and _Hydro_.

Note: this patch will significantly increase the time to compute the topological order since all available package.xml files from all underlays must be crawled and parsed.


tfoote:
+1

wjwwood:
+1

PR #593, opened 2014-02-14
support using nosetests with Python3

dirk-thomas:
If Python 3 is used find the matching version of nosetests and actually output what exactly is being used.


tfoote:
+1

wjwwood:
+1

PR #594, opened 2014-02-15
implement list_insert_in_workspace in plain CMake without python

xqms:
This speeds up the CMake run from 12s to 8s on a workspace with ~50
packages.

The logic follow exactly the order_paths() function from workspaces.py.
The helper order_paths.py is not needed anymore.


dirk-thomas:
From reading the patch it looks like that the prefix is compared with the list element without adding a trailing slash which would break behavior in certain cases (see current implementation https://github.com/ros-infrastructure/catkin_pkg/blob/master/src/catkin_pkg/workspaces.py#L83).

Please consider to also convert the corresponding unit test (https://github.com/ros/catkin/blob/indigo-devel/test/unit_tests/test_order_paths.py) to verify that the patched code complies.

xqms:
Thanks for catching that issue. I converted the unit test, but didn't know how to correctly register a CMake unit test. I hope my solution is not too hackish...

dirk-thomas:
I just compared the time to configure a workspace. My workspace contains 63 packages - all packages up to ros_tutorials. The configure step when starting from a non-existing build space took 34.3s (catkin) vs. 32s (this patched version). For a repeated configure step it took 31.8s vs. 30.8s.

While the difference is measurable the overall improvement is rather insignificant (for my example). In order to decide if replicating the existing Python code in CMake is worth the effort (one-time implementation as well as maintenance in the future) I would like to see more performance comparisons. Can you please provide more detailed information where you experienced the more significant improvement (e.g. a rosinstall file of your workspace)?

dirk-thomas:
Without any further information provided I will close this and mark it as "wontfix". Please feel free to provide more information (the ticket can always be reopened).

PR #601, opened 2014-02-28
Fixed bug with length calculation

cehberlin:
Because the string length is calculated with an additional '/' it is necessary to add this character as well to the substring creation. If not it is possible to get out of range errors like below. Another possibility would be to remove the extra '/' from the length calculation in line 396 and 397, but I do not know if this has any side effects

Fixed build errors like below:
-- Using CATKIN_TEST_RESULTS_DIR:
~/ros_catkin_ws/build_isolated/rqt_image_view/test_results
-- Found gtest sources under '/usr/src/gtest': gtests will be built
-- catkin 0.5.84
CMake Error at
/opt/ros/hydro/share/catkin/cmake/catkin_package.cmake:399
(string):
  string end index: 57 is out of range -1 - 56


dirk-thomas:
Thank you for catching this case.

The suffix must be there to avoid matching two paths like `/foo` with `/foobar`.

PR #602, opened 2014-03-03
Include dir handling

dirk-thomas:
@tfoote Please review.


cmake/catkin_package.cmake
```diff
@@ -281,12 +281,16 @@ function(_catkin_package)
   set(PROJECT_CMAKE_CONFIG_INCLUDE_DIRS "")
   set(PROJECT_PKG_CONFIG_INCLUDE_DIRS "")
   foreach(idir ${PROJECT_INCLUDE_DIRS})
-    if(IS_ABSOLUTE ${idir} AND IS_DIRECTORY ${idir})
```
dirk-thomas, line 284: This block of changes (lines 284-293) should have no functional changes. The previously combined error message is now separated into two more specific messages.

cmake/catkin_package.cmake
```diff
@@ -391,15 +395,17 @@ function(_catkin_package)
...
-    string(LENGTH "${CATKIN_DEVEL_PREFIX}/" devel_length)
-    string(LENGTH "${idir}/" idir_length)
-    if(NOT ${idir_length} LESS ${devel_length})
-      string(SUBSTRING "${idir}/" 0 ${devel_length} idir_prefix)
-      if("${idir_prefix}" STREQUAL "${CATKIN_DEVEL_PREFIX}/")
-        # the value doesn't matter as long as it doesn't match IS_ABSOLUTE
-        set(idir "${CATKIN_GLOBAL_INCLUDE_DESTINATION}")
+    string_starts_with("${idir}/" "${CMAKE_CURRENT_SOURCE_DIR}/" _is_source_prefix)
+    string_starts_with("${idir}/" "${CMAKE_CURRENT_BINARY_DIR}/" _is_build_prefix)
+    string_starts_with("${idir}/" "${CATKIN_DEVEL_PREFIX}/" _is_devel_prefix)
```
dirk-thomas, line 402: Absolute include dirs in all three locations are considered to be "relative" which means an "include" entry is places in the generated CMake config file instead of the absolute path.

tfoote:
+1 looks like what we discussed.

Issue #304, opened 2013-01-13
catkin could advertise the buildspace


tkruse:
Based on this discussion: ros/catkin#294

Currently sourcing devel/setup.bash creates an env from which the develspace and source spaces can be derived. However, the buildspace cannot be derived from the environment.

This ticket suggests that catkin defines a way to create an environment that includes a pointed to the build space. This can be used for global build command invocation or quick navigation to the build space.

One possibility would be top have a file generated like:

```
workspace/build/setup.*sh
```

which sources the devel space setup.*sh, but also sets some Environment variable.

In the case of isolated build spaces, this setup is not too useful, but catkin_make_isolated could be extended to create a parent build-space for all isolated build spaces, into which such setup.*sh files would fit into.

Another alternative would be for catkin_make to generate setup.*sh files in the workspace root. However that still hides the detail of the build sace which can be useful for convenience tools.

dirk-thomas:
Please describe a specific use case which can currently not be realized but would be implemented. Also please describe how that should work in detail.

According to the description above the user would require to source the different setup from build space which seems to be not a good idea.

tkruse:
The use case is for the user to invoke build from anywhere on the filesystem easily without having to navigate to a specific folder via an absolute or relative path.

So at least one of these should be implemetable after this issue is fixed:

```
roscd; catkin_make
roscd; make
roscd foopkg; catkin_make
```

Currently no tool can implement this behavior based on catkin information alone, because sourcing the devel/setup.sh does not give a hint to the buildspace(s).

dirk-thomas:
Case 1 and 3 are not valid since catkin_make does not allow the invocation in the build folder.
Case 2 is not catkin related at all. So I don't think that catkin should address case 2.

tkruse:
What cakin_make allows today is irrelevant, what matters is what
catkin_make could allow in the future. It would have to be changed as well,
obviously, but that's for another ticket. And case2 is only possible if a
catkin offers a convenient way to locate a build space, so it is
catkin-related in my mind. no other too in the workflow can make case2
possible.

tkruse:
[...]

PR #603, opened 2014-03-03
Marker file for catkin_make (#304)

tkruse:
First commit only writes a marker file. Possibly that is all you want to merge.

The third commit also makes use of the marker file, in case no explicit target path argument was passed to catkin_make.

I was not sure for which part you wanted a PR, so you can cherry-pick what you want. I am dispassionate about the filename, and about file vs. folder, so you can change to your liking.

I recommend using yaml or a folder because a likely future extention is to store metadata such as preferences or build configurations, and it would be bloat to have to store them in separate files next to the marker file.

### DIFF
--- a/bin/catkin_make
+++ b/bin/catkin_make
@@ -19,6 +19,7 @@ from catkin.builder import handle_make_arguments
 from catkin.builder import print_command_banner
 from catkin.builder import run_command
 from catkin.builder import run_command_colorized
+from catkin_pkg.workspaces import ensure_workspace_marker
 from catkin_pkg.packages import find_packages
 
 
@@ -158,6 +159,8 @@ def main():
                 run_command_colorized(cmd, build_path)
         except subprocess.CalledProcessError:
             return fmt('@{rf}Invoking @{boldon}"make cmake_check_build_system"@{boldoff} failed')
+
+    ensure_workspace_marker(base_path)
 
     # invoke make
     cmd = ['make']
--- a/package.xml
+++ b/package.xml
@@ -19,7 +19,7 @@
   <buildtool_export_depend>cmake</buildtool_export_depend>
 
   <depend>python-argparse</depend>
-  <depend version_gte="0.2.1">python-catkin-pkg</depend>
+  <depend version_gte="0.2.2">python-catkin-pkg</depend>
 
   <build_depend>python-empy</build_depend>
 
--- a/python/catkin/builder.py
+++ b/python/catkin/builder.py
@@ -58,6 +58,7 @@ except ImportError as e:
 
 from catkin.cmake import get_cmake_path
 from catkin.terminal_color import ansi, disable_ANSI_colors, fmt, sanitize
+from catkin_pkg.workspaces import ensure_workspace_marker
 
 
 def split_arguments(args, splitter_name, default=None):
@@ -818,6 +819,8 @@ def build_workspace_isolated(
     if not force_cmake and cmake_input_changed(packages, buildspace, cmake_args=cmake_args, filename='catkin_make_isolated'):
         print('The packages or cmake arguments have changed, forcing cmake invocation')
         force_cmake = True
+
+    ensure_workspace_marker(workspace)
 
     # Build packages
     pkg_develspace = None


tkruse:
On second thought might be even better to write marker after configuration has succeeded, as make failure does not invalidate the workspace (while cmake failure might). Added commit.

tkruse:
reorganized commits a bit, now first commit just writes the marker file, next real commit provides a python function to find it, last commit invokes catkin_make by first searching in path ancestors for marker file.

wjwwood:
Thank you for opening the pull request.

I think the name should somehow contain 'workspace', as there are already `.catkin` files in the devel and install space. Perhaps `<workspace_root>/.catkin_workspace.yaml` or just `<workspace_root>/.workspace.yaml` would be good, I'm open to suggestion though.

tkruse:
I am happy with anything you suggest. I like a hint of workspace in there. Even ".workspace" would work for me.

dirk-thomas:
Sine `.catkin` already exists with different semantic and `.workspace` gives no good indication where it comes from I would suggest using `.catkin_workspace`. I would omit the file extension for now since it does not contain any yaml yet. Instead the Python API should also be exposed by a script (e.g. `catkin_find_workspace_root`) to make it accessible to other languages.

Since the marker file is not specified by any REP it should not be used directly but only through this API / script and therefore changing the extension later (when we have need for yaml content) should be no problem.

python/catkin/workspace.py
```diff
@@ -34,11 +34,12 @@
 import os
 
 CATKIN_MARKER_FILE = '.catkin'
+WORKSPACE_ROOT_MARKER_FILE = '.catkin.yaml'
```
dirk-thomas: I would suggest naming the constant `CATKIN_WORKSPACE_MARKER_FILE`.

python/catkin/workspace.py
```diff
@@ -50,6 +51,18 @@ def get_workspaces():
     return workspaces
 
 
+def get_containing_workspace_root(path):
+    """
+    Return path or first ancestor containing catkin workspace root makrerfile, or None
+    """
+    looppath = path
```
dirk-thomas: Please remove the `looppath` variable and just use `path` in the loop.

python/catkin/workspace.py
```diff
@@ -50,6 +51,18 @@ def get_workspaces():
     return workspaces
 
 
+def get_containing_workspace_root(path):
+    """
+    Return path or first ancestor containing catkin workspace root makrerfile, or None
```
dirk-thomas: Please fix couple of spell errors.

dirk-thomas:
I think the same changes as for `catkin_make` should be done for `catkin_make_isolated` since the users will expect similar behavior.

python/catkin/workspace.py
tkruse: renamed markerfile var and marker file name

python/catkin/workspace.py
tkruse: done

python/catkin/workspace.py
tkruse: extended-fixed spelling

tkruse:
- renamed marker var and file
- renamed looppath
- fixed/changed docstrings
- Made same change to c_m_i (before cmake invocation here)
- Also added commit to unify workspace sanity checks between catkin_make and c_m_i

Last commit could be a separate PR, tell me if you want it out.

tkruse:
wait... vulnerability detected...

tkruse:
ok, fixed. Good for review.

dirk-thomas:
As you suggested yourself please keep the last commit containing the refactoring separate since it is not related to the marker file.

tkruse:
- Removed refactor commit
- added unit tests
- added shell script as suggested in first comment

wjwwood:
Looking good, I think you need to add `bin/catkin_find_workspace_root` to the `setup.py`:

https://github.com/ros/catkin/blob/indigo-devel/setup.py#L9

tkruse:
added script to setup.py

wjwwood:
Cursory code review looks ok, but I haven't tested it locally yet.

jbohren:
I've tried it out, there doesn't seem to be any issues, though does the indigo-devel branch search more places for package.xml files? It found a template package.xml file in orogen which the current version of catkin has never found before. 

---

Meanwhile, this addition brings up a bit of nomenclature that I've always found confusing.

When someone runs `catkin_init_workspace`, it's to declare the "source space" of a catkin workspace while all the other language describes the workspace as the directory under which `build`, `devel`. `src`, etc. can be found.

Before the existence of a `.catkin_worksapce` file, the only thing that really made a catkin workspace "special" was a symlink to `toplevel.cmake`. Now, however, it seems like it makes more sense for `catkin_init_workspace` to create a `.catkin_workspace` file in the current directory instead of create the symlink to `toplevel.cmake` in the source space. 

---

It seems like the addition of contents to this file could dramatically change how catkin works underneath (specification of workspace chaining, noting whether or not a workspace was built in isolation, etc). It's also been the subject of the solution to issues with `catkin build` (https://github.com/catkin/catkin_tools/issues/11, https://github.com/catkin/catkin_tools/issues/10) 

With that in mind, I have three questions @dirk-thomas @wjwwood @tkruse:
1. Which tools should be allowed to modify the contents of the marker file? Should it only be modified through invocations of `catkin_make` and `catkin_make_isolated`, or could 3rd-party tools modify it for their own means. It seems like if the vanilla catkin scripts create it, only they should change its contents (or it can be edited by hand, similar to .git/config).
2. If only the vanilla catkin scripts can modify the marker file, are there currently the resources available to decide on what content it should contain? 
3. If anyone can modify the marker file, should they be limited to certain sections of it? Like if `catkin build` modifies it, should all of its metadata be listed under a `catkin_tools` section?

tkruse:
Hi Jon,
I think we should wait for #603 to be merged before deciding how to further use its contents.

It might be nice later to have a syntax and general read/write contract on the file, but so far nothing has been discussed, and as i said discussing those things is a huge pain if the people discussing have fundamentally incompatible views. 

Until further notice, any tool should read/write it's own file and place it relative to the marker file (e.g. in the same folder). That way we do not need to discuss at all, and there will not be conflicts should catkin_make once in the future decide to use the file for writing content.

wjwwood:
> Now, however, it seems like it makes more sense for `catkin_init_workspace` to create a `.catkin_workspace` file in the current directory instead of create the symlink to `toplevel.cmake` in the source space.

The purpose of `catkin_init_workspace` was to only create the `toplevel.cmake` symlink, in order to allow users to build a workspace without using `caktin_make` (`catkin_init_workspace` is no required with `caktin_make`). I think it makes sense to argue that `catkin_init_workspace` should be called `catkin_init_sourcespace` or something like that, but I'm not sure that it makes sense to extend its behavior, or to have it not symlink the `toplevel.cmake` in to the root of the source space as it is required for technical reasons, not just to mark the source space's location. I think that the tools which operate on the workspace can create the `.catkin_workspace` file. Now, there may be a new command verb for creating that file before building is started, but that depends ultimately on the functionality which relies on the that file's existence.

> Which tools should be allowed to modify the contents of the marker file? Should it only be modified through invocations of catkin_make and catkin_make_isolated, or could 3rd-party tools modify it for their own means. It seems like if the vanilla catkin scripts create it, only they should change its contents (or it can be edited by hand, similar to .git/config).

It seems like something that we should consider on a tool by tool basis, and it is not like we can enforce the behavior through anything other than recommendations.

> If only the vanilla catkin scripts can modify the marker file, are there currently the resources available to decide on what content it should contain?

For now, I think it should contain nothing. It would be conceivable for it to contain previous invocation arguments and settings, so that rerunning it is quicker, but I think that this needs to be considered carefully, because it becomes awkward to deal with some of the workflows. I don't think I have any time to spend on this, but I can give feedback on proposals and prototypes from others from time to time.

> If anyone can modify the marker file, should they be limited to certain sections of it? Like if catkin build modifies it, should all of its metadata be listed under a catkin_tools section?

I think you could argue that the different tools should not be allowed to operate on the same workspace, because workspaces are cheap to make, by creating a new folder and copying or symlinking the source. In general, however, it does make sense for tools to namespace themselves, either in the file or in separate files. There was mention of making `.catkin_workspace` a folder instead.

bin/catkin_find_workspace_root
```diff
@@ -0,0 +1,30 @@
...
+    return parser.parse_args(args=args)
+
+
+def main():
+    args = parse_args()
+    result = get_containing_workspace_root(os.path.abspath(args.path or '.'))
+    if result:
+        print(result)
+    else:
+        return errno.ENOENT
```
dirk-thomas: Please don't use this error code since `No such file or directory` is commonly used when a path passed in as an argument does not exist or when you can print an error message like: `No such file or directory: /does/not/eixst`. Therefore it kind of gives a wrong feedback to the user.

Please add an error message indicating that the marker file could not be found.

bin/catkin_find_workspace_root
```diff
@@ -0,0 +1,30 @@
...
+
+# find the import relatively if available to work before installing catkin or overlaying installed version
+if os.path.exists(os.path.join(os.path.dirname(__file__), '..', 'python', 'catkin', '__init__.py')):
+    sys.path.insert(0, os.path.join(os.path.dirname(__file__), '..', 'python'))
+from catkin.workspace import get_containing_workspace_root
+
+
+def parse_args(args=sys.argv[1:]):
+    parser = argparse.ArgumentParser(description='returns the workspace root folder for the passed argument path, or else for the current path')
+    parser.add_argument('path', nargs='?', help='Any path from where to search')
```
dirk-thomas: Please let argparse check that the argument is actually a valid directory (like here: https://github.com/ros-infrastructure/rosinstall_generator/blob/upstream_options/bin/rosinstall_generator#L64).

bin/catkin_find_workspace_root
```diff
@@ -0,0 +1,30 @@
...
+    return parser.parse_args(args=args)
+
+
+def main():
+    args = parse_args()
+    result = get_containing_workspace_root(os.path.abspath(args.path or '.'))
```
dirk-thomas: Please make `.` the default value in argparse so that `--help` can print the default value as well.

tkruse: rebased PR and changed according to comments:
- argparse validation of passed path argument
- return -1 instead or errno error code, and print error message
- argparse --help prints default value for path

In contrast to comment, new argparse validation checks for valid file or
folder, since I see no need to restrict the argument to folders only, i.e.
it is valid to ask for the workspace root for a given file within that
workspace.

bin/catkin_find_workspace_root
```diff
@@ -0,0 +1,40 @@
...
+    if result:
+        print(result)
+    else:
+        print("No workspace root could be found in the chain of ancestors of path '%s'" %args.path)
```
dirk-thomas: Please use spaces around the `%`.

Please print the error message to stderr.

bin/catkin_find_workspace_root
```diff
@@ -0,0 +1,40 @@
...
+def _existing_path(path):
+    if not os.path.exists(path):
+        raise argparse.ArgumentTypeError("'%s' is not an existing file or directory" % path)
+    return path
+
+def parse_args(args=sys.argv[1:]):
+    parser = argparse.ArgumentParser(description='returns the workspace root folder for the passed argument path, or else for the current path',
+                                     formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument('path', nargs='?',
+                        type=_existing_path,
```
dirk-thomas: If checking for an existing path only please use the API available in argparse (`type=argparse.FileType('r')`) rather then implement a custom check function.

tkruse:
Rebased and modified last commit printing message to stderr, and using argparse.Filetype('r')

dirk-thomas:
I tried the current state of the PR but the logic to walk up to find the workspace root (https://github.com/ros/catkin/pull/603/files#diff-73f511dee79a0650b15251b79faf10ffR44) fails for my workspaces. It looks like it does not handle use cases where some of the packages/repos in the source space are symlinked into the workspace. Can you please update the PR to also work in these scenarios?

tkruse:
Modified last commit to be more robust against symbolic links in common shell environments. 
Also reverted usage of `argparse.FileType`, since that does not work with directories.
Moved normalizing and abspath into library function so that other clients do not need to re-implement.

dirk-thomas:
It still does not work for me. E.g. I have the folder `genmsg` with the package `genmsg` in the sourcespace but it is actually a symlink to a different location (where I store the cloned repositories). Invoking `catkin_make` in this folder results in:

```
The specified base path "/home/dthomas/ros/indigo/github/ros/genmsg" contains a CMakeLists.txt but "catkin_make" must be invoked in the root of workspace
```

tkruse:
Modified last commit to also fix symlink behavior of catkin_make and catkin_make_isolated

jbohren:
Does this just need more user testing?

wjwwood:
It certainly could use that, personally, I just haven't had time to look into it yet.

jbohren:
> It certainly could use that, personally, I just haven't had time to look into it yet.

Whats the easiest way to try it out? Just install to `/usr/local`?

wjwwood:
No just clone @tkruse's fork for this pull request into a workspace and locally call `./src/catkin/bin/catkin_make` or `./src/catkin/bin/catkin_make_isolated`. No need to install this version of catkin anywhere.

jbohren:
> No just clone @tkruse's fork for this pull request into a workspace and locally call `./src/catkin/bin/catkin_make` or `./src/catkin/bin/catkin_make_isolated`. No need to install this version of catkin anywhere.

Ok, cool.

dirk-thomas:
It seems to still need more patching. I just tried invoking `catkin_make_isolated --install` on a workspace containing only catkin and one other package with the following stacktrace:

```
  File "./src/catkin/bin/catkin_make_isolated", line 136, in <module>
main()
  File "./src/catkin/bin/catkin_make_isolated", line 132, in main
destdir=destdir
  File "./src/catkin/bin/../python/catkin/builder.py", line 707, in build_workspace_isolated
ws_sourcespace = os.path.join(workspace, 'src')
  File "/usr/lib/python2.7/posixpath.py", line 77, in join
elif path == '' or path.endswith('/'):
AttributeError: 'NoneType' object has no attribute 'endswith'
```

Please fix this problem and perform some more checking on really trivial use cases. I know that there endless combinations which could be tested and might not all be feasible to try but at least the really simple ones should be verified to work.

tkruse:
Fixed AttributeError for both catkin_make and catkin_make Isolated. Rebased.

python/catkin/workspace.py
```diff
@@ -50,6 +51,43 @@ def get_workspaces():
...
+
+def ensure_workspace_marker(base_path):
+    """
+    creates workspace marker file at path if not existing
+
+    :param path: target folder
+    """
+    if not os.path.exists(os.path.join(base_path, CATKIN_WORKSPACE_MARKER_FILE)):
+        with open(os.path.join(base_path, CATKIN_WORKSPACE_MARKER_FILE), 'a') as fhand:
+            fhand.write('# This file currently only serves to mark the location of a catkin workspace for tool integration')
```
dirk-thomas: Please add a newline character at the end of the line.

dirk-thomas:
A problem related to symlink handling still exists.

I first invoked `cmi` with some argument like `--build mybuild`.
Then I change into `src/catkin` which is actually a symlinked folder.
When now invoking `cmi` again with the correct relative argument `--build ../../mybuild` the build space is resolved wrongly.

It first resolves the real path of the current folder (which is the symlink destination) and then goes relative from there. Instead it should first apply the relative path and afterwards get the real path.

tkruse:
- Added newline
- made build, devel and install space command line arguments robust against invocation inside symlink. Note that this PR did not introduce that vulnerability.

jbohren:
@wjwwood I wanted to test this with `catkin build` but I guess to do that it would actually need to be installed somewhere?

dirk-thomas:
The problem with the symlinks still exists when I use `--source`.

dirk-thomas:
Could you please add some test(s) to this PR in order to ensure that this functionality works with various different use cases.

tkruse:
Since I am now asked to fix further existing bugs in catkin to have my pull request merged, i reduced this PR to only create the marker file.

wjwwood:
@jbohren as I pointed out, `catkin_tools` should not use this implementation unless it is relocated because otherwise `catkin_tools` would need to depend on `catkin`, which would make it where it could not be installed from `pip` and I really would like to avoid that.

wjwwood:
The CI failure is related to this I think: https://github.com/ros/catkin/pull/615

NikolausDemmel:
@tkruse: Are the commits that fixed some of the discussed issues (not purely related to this PR) still available somewhere (for a seperate PR)?

tkruse:
old commit available at https://github.com/tkruse/catkin/tree/search_base_old
I would create a second pull requests after this one has been merged and the community informed about what goes on here.

wjwwood:
@tkruse that seems like a perfectly fine way to go about it.

I wanted to point out again that I think this functionality should be provided in `catkin_pkg` so that it can be reused outside of `catkin` without depending on `catkin`, specifically in `catkin_tools`. If we don't do that, I would be forced to reimplement some or all of this functionality in `catkin_tools`.

Specifically I think that snippet from `python/catkin/workspace.py` should be somewhere in `catkin_pkg`.

dirk-thomas:
Please update the PR to use the new version of `catkin_pkg` (ros-infrastructure/catkin_pkg#95) and rebase it to make sure that it is still mergable.

A new `catkin` release should happen very soon (next 1-2 days). If the PR is updated by then I will merge it for the release. Otherwise it will be released with the next iteration (for `catkin` that is usually a longer time window around 1 - 3 months).

tkruse:
updated and rebased PR, added versioned dependency.

wjwwood:
+1

dirk-thomas:
+1 both (`catkin_make` as well as `catkin_make_isolated`) work for me

jbohren:
Should this be backported into hydro?

dirk-thomas:
Since the functionality in catkin_pkg and (eventually soon) catkin_tools is cross distro and this only adds API I would backport it to Groovy/Hydro before their new patch releases.

jbohren:
> Since the functionality in catkin_pkg and (eventually soon) catkin_tools is cross distro and this only adds API I would backport it to Groovy/Hydro before their new patch releases.

Haha, when you say "I" do you mean you, or do you mean someone else should backport it?

dirk-thomas:
Commonly some time after I released a package into the latest ROS I go through all commits and consider to cherry pick whatever should be back ported and then (if "enough" are there) also release the package into the "older" distros. You can expect this to happen for catkin next week.

jbohren:
@dirk-thomas

Ahh that's good to know. Thanks!

Issue #545, opened 2013-10-25
Support arch-specific lib and cmake folders with catkin_make_isolated


scpeters:
In order to get into debian, they recommend using GNUInstallDirs for installing architecture-dependent libraries. For example, [sdformat ](https://bitbucket.org/osrf/sdformat/pull-request/64/support-for-gnuinstalldirs/diff) is in the process of this change, which on my machine causes libraries to be installed to `{prefix}/lib/x86_64-linux-gnu/` and the cmake config file to be installed to `{prefix}/lib/x86_64-linux-gnu/cmake/sdformat/`.

There was some discussion about whether GNUInstallDirs should be used only for packaging debians or for all source installs. In the end, sdformat chose to use it always by default, and I think other packages may do the same. It would greatly enhance the utility of catkin_make_isolated to support this behavior, both for setting the CMAKE_PREFIX_PATH and for setting the LD_LIBRARY_PATH in the setup.sh scripts.

I had been using `catkin_make_isolated` to build sdformat and gazebo together by manually adding my own package.xml files. gazebo already uses GNUInstallDirs, so I have had to manually edit the LD_LIBRARY_PATH in setup.sh so that the gazebo binaries can find the gazebo libraries. Now with the sdformat change, gazebo cannot `find_package` sdformat because the cmake folder is in a subfolder of lib rather than share.

One clue for implementing this would be to parse the contents of `/etc/ld.so.conf.d/`:

```
$ cat /etc/ld.so.conf.d/x86_64-linux-gnu.conf 
# Multiarch support
/lib/x86_64-linux-gnu
/usr/lib/x86_64-linux-gnu
```

scpeters:
An alternative to parsing `/etc/ld.so.conf.d/` is to include the GNUInstallDirs module in a sandbox to see if there are arch-specific lib folders for the given platform. An example CMakeLists.txt:

```
cmake_minimum_required(VERSION 2.8 FATAL_ERROR)
project (CMAKE_GNU_DIRS_TEST)
include(GNUInstallDirs)
message (STATUS "CMAKE_INSTALL_LIBDIR ${CMAKE_INSTALL_LIBDIR}")
```

and its output on Ubuntu:

```
...
-- CMAKE_INSTALL_LIBDIR lib/x86_64-linux-gnu
...
```

scpeters:
Note that this will be an issue for console_bridge when 0.2.5 is released (currently it's on 0.2.4).

[...]

PR #604, opened 2014-03-05
Handle arch-specific library folders #545 (Not ready for merging)

scpeters:
Issue #545 covers problems using `catkin_make_isolated` with projects that use GNUInstallDirs (such as console_bridge, sdformat, simbody, and gazebo). The problem on Ubuntu is that libraries get installed to `lib/x86_64-linux-gnu` and catkin can't find them. I've made a workaround in this pull request, though it hard-codes the suffix. To find the suffix automatically, one can invoke cmake with the following CMakeLists.txt file and parse the output:

```
cmake_minimum_required(VERSION 2.8 FATAL_ERROR)
project (CMAKE_GNU_DIRS_TEST)
include(GNUInstallDirs)
message (STATUS "CMAKE_INSTALL_LIBDIR ${CMAKE_INSTALL_LIBDIR}")
```

It generates the following output:

```
$ cmake .. | grep CMAKE_INSTALL_LIBDIR
-- CMAKE_INSTALL_PREFIX: lib/x86_64-linux-gnu
```

I've been using this code in my fork for over 6 months with no problems. It's even able to find cmake config files embedded in places like `lib/x86_64-linux-gnu/cmake/project` (tested with simbody). It's obviously hard-coded for Ubuntu, and I'd be willing to try to generalize it, but no one offered any advice in #545, so I figured making a pull request might spur some discussion.


scpeters:
@wjwwood and I discussed this the other day. Parsing the console output of the sample CMakeLists.txt file pasted above seems like an option. @wjwwood mentioned that there might be a command-line argument to print a specified cmake variable (similar to `-DCMAKE_VARIABLE=X`), but I couldn't find it in the documentation.

Relevant to:
https://github.com/ros-infrastructure/bloom/pull/268

wjwwood:
I was thinking of something like this:

``` cmake
include(GNUInstallDirs)

message(${CMAKE_INSTALL_LIBDIR})
```

```
% cmake -P test.cmake
lib
```

scpeters:
That looks like a great way to print the `CMAKE_INSTALL_LIBDIR` variable. I don't know where in the catkin code should cmake be invoked to evaluate this though.

thomas-moulard:
FYI:
- MultiArch documentation: https://wiki.debian.org/Multiarch
  - Tuples list: https://wiki.debian.org/Multiarch/Tuples#Used_solution

You can get the tuple programmatically using `dpkg-architecture`:

```
dpkg-architecture -qDEB_HOST_MULTIARCH
```

So basically:

```
LIBDIR=lib/`dpkg-architecture -qDEB_HOST_MULTIARCH 2> /dev/null`
```

If `dpkg-architecture` does not exist (as on non-Debian distributions) then it will fail quietly and the result remains correct (i.e. `lib/`).

wjwwood:
@thomas-moulard the reason I was wanting to use the cmake solution is that it will track any future changes to the `GNUInstallDirs` cmake module.

However, you solution may work as well, and depending on how it is implemented it may end up being called from Python where we have much more control over the circumstances in which is executed (check for existence first rather than relying on `2>`).

thomas-moulard:
Either way is fine for me too but IMHO following standards (i.e. the Debian policy) should be preferred over mimicking a particular implementation of the standard (i.e. GNUInstallDirs.cmake).

wjwwood:
> following standards (i.e. the Debian policy) should be preferred over mimicking a particular implementation of the standard (i.e. GNUInstallDirs.cmake).

That's a good point, I'm fine with using the `LIBDIR=lib/`dpkg-architecture -qDEB_HOST_MULTIARCH 2> /dev/null`` one liner.

scpeters:
I just fixed the install builds in cd42e10 by updating builder.py; I'm not exactly sure what's the most elegant way to fix devel builds, since that requires changes in `_setup_util.py.in`. It would be easy to invoke a subprocess each time the setup script is called, but it seems like it would be better to pass the multiarch string as a variable when that script is configured. Can anyone advise on how to do that?

wjwwood:
@scpeters I can give you some pointers. This is where the `_setup_util.py.in` file gets templated:

https://github.com/ros/catkin/blob/770135f0a6502cdfbbe74a8149a78791aa0df2c0/cmake/catkin_generate_environment.cmake#L14

You would just need to set a CMake variable with a list of the right lib directories (it should include `CATKIN_GLOBAL_LIB_DESTINATION` which should eval to `lib`) before that call. Then use that variable in the template here:

https://github.com/ros/catkin/blob/61a8f012450ae379051de7da35d2bd1fc8e1d204/cmake/templates/_setup_util.py.in#L55

Instead of just `CATKIN_GLOBAL_LIB_DESTINATION`, that list may also include architecture specific directories as well.

Since you are calling out to the system for this information, I'm not sure if it makes sense for it to happen in the `catkin_generate_environment` function (in the `catkin_generate_environment.cmake` file linked above) or somewhere else, @dirk-thomas might have a better idea. Either way the check should probably be guarded on at least `if(UNIX)` if not something more specific.

scpeters:
See #624 instead.

PR #607, opened 2014-03-20
"Notify" user of a cyclic package dependency

chris-blay:
It's too bad a lot of useful information was lost inside `topological_order.py` but still better than _just_ an `AttributeError`.

Still looking for other places that need to check for this...


dirk-thomas:
Thanks for reporting the problem. Your pull request won't be merged as-is due to the inappropriate commend and the lack of printing the actual package names involved in the circular dependency which makes it less value for the user.

PR #608, opened 2014-03-20
"Notify" user of a cyclic package dependency (part 2 of 2)

chris-blay:
See https://github.com/ros/catkin/pull/607

I think I found them all... Hopefully someone remembers to always test for this whenever using topological_order() in the future.


dirk-thomas:
Thanks for reporting the problem. Your pull request won't be merged as-is due to the inappropriate commend and the lack of printing the actual package names involved in the circular dependency which makes it less value for the user.

PR #609, opened 2014-03-20
support setting ${PROJECT_NAME}_LIBRARIES before catkin_package()

dirk-thomas:
Similar to what is already possible with ${PROJECT_NAME}_INCLUDE_DIRS, ${PROJECT_NAME}_CFG_EXTRAS, etc.


dirk-thomas:
@wjwwood Can you please give this a try with the prototype?

wjwwood:
+1 works for the prototype

PR #614, opened 2014-04-04
add travis file

dirk-thomas:


dirk-thomas:
The one failing unit test is actual what is currently expected. Thats why I am adding the travis check in the first place :wink:

PR #595, opened 2014-02-18
fix unicode error with japanese LANG (fix #578)

dirk-thomas:
@josegaert @nicolasWDC @saljam @wjwwood @yukota This PR supersedes #578.

Please test your use case against this branch of catkin and report if it works for you. If not please provide a detailed description how to reproduce your problem.

PR #578, opened 2014-01-09
Fix UnicodeEncodeError

yukota:
if fail to encode to utf8, unencoded charactor is written.

PR #615, opened 2014-04-04
improve unicode handling

dirk-thomas:
As a follow up to #578 and #595 I have another patch to improve unicode handling.

Added another unit test which check processing when output is encoded.

@josegaert @nicolasWDC @saljam @wjwwood @yukota Since you have all been part of the previous discussion I would like to ask you to try the changes of this branch with your specific scenario.

Since these unicode issues are highly specific to your environment, available locale and output of your tools the more people will validate that it works the more confident we can be that it does not break things (as it has happened before). Please comment on this ticket with either feedback ("it works for you" as well as "it doesn't (please with as much details as possible)").

### DIFF
--- a/python/catkin/builder.py
+++ b/python/catkin/builder.py
@@ -39,7 +39,7 @@ import platform
 import re
 import stat
 try:
-    from cStringIO import StringIO
+    from StringIO import StringIO
 except ImportError:
     from io import StringIO
 import subprocess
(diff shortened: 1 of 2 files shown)


saljam:
LGTM.

yukota:
LGTM!!

wjwwood:
+1

PR #616, opened 2014-04-06
fix bad usage of IMPORTED_IMPLIB by IMPORTED_LOCATION

vrabaud:
ok, that was NOT FUN. To test this, just use the tf2_geometry_msgs. It DEPENDS on orocos_kdl in catkin_package() but this library does not appear in the *.cmake files. My patch fixes that.

Info at: http://www.cmake.org/Wiki/CMake/Tutorials/Exporting_and_Importing_Targets and http://www.cmake.org/cmake/help/v2.8.12/cmake.html#prop_tgt:IMPORTED_LOCATION


vrabaud:
I don't understand why Travis is failing.
On the other hand, what I am solving is needed to prevent some failures on the farm:
http://jenkins.ros.org/job/ros-indigo-image-rotate_binarydeb_trusty_i386/2/

tfoote:
The failure is a unicode issue in 2.7 it passes in 3.3

dirk-thomas:
The failure on Python 2.7 is addressed in #615. That was the whole reason to add the travis file recently.

dirk-thomas:
@vrabaud I have ran a build of `orocos_kdl` and (a stripped down version of `tf2_geometry_msgs`. What ever version of catkin I use (the unpatched as well as the patched version) the _orocos-kdl_ library is never part of the downstream package CMake config file. How did you actually tested your patch? Can you please post some detailed command sequences and expected / actual output?

vrabaud:
@dirk-thomas , I just create a workspace with:

```
git clone https://github.com/ros/geometry_experimental.git
git clone https://github.com/vrabaud/catkin.git
git clone https://github.com/ros-perception/image_pipeline.git
```

(checkout all the indigo for each)
source /opt/ros/indigo/setup.sh and call:

```
./src/catkin/bin/catkin_make
```

and it works for me.

dirk-thomas:
I was able reproduce the problem and also verify that the same patch is required for Hydro and Groovy. Thank you for the PR.

PR #618, opened 2014-04-11
python: fix python path setting in plain cmake workspaces

smits:
In plain cmake workspaces the python install dir is set to
lib/python2/dist-packages instead of lib/python2.7/dist-packages

I checked the catkin/cmake/python.cmake file and in that file there is
an extra check for the version, so I assume that we should only use
the major version in case the major version is 3 or higher.

### DIFF
--- a/python/catkin/builder.py
+++ b/python/catkin/builder.py
@@ -240,9 +240,9 @@ def get_python_install_dir():
     python_install_dir = 'lib'
     python_use_debian_layout = os.path.exists('/etc/debian_version')
     if os.name != 'nt':
-        python_version_xdoty = str(sys.version_info[0])
-        if not python_use_debian_layout:
-            python_version_xdoty += '.' + str(sys.version_info[1])
+        python_version_xdoty = str(sys.version_info[0]) + '.' + str(sys.version_info[1])
+        if python_use_debian_layout and sys.version_info[0] == 3:
+            python_version_xdoty = str(sys.version_info[0])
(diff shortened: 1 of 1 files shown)


smits:
The Travis CI build also fails without my patch!

dirk-thomas:
Yes, the travis file was just added recently and a PR exists to fix the failing unit test.

The Python function to determine the folder name is obviously not in sync with the CMake equivalent: https://github.com/ros/catkin/blob/indigo-devel/cmake/python.cmake
I will add a small comment in the patch to make it exactly identical. If you could please update that (and squash your commits afterwards) I will merge this.

python/catkin/builder.py
```diff
@@ -241,7 +241,7 @@ def get_python_install_dir():
     python_use_debian_layout = os.path.exists('/etc/debian_version')
     if os.name != 'nt':
         python_version_xdoty = str(sys.version_info[0])
-        if not python_use_debian_layout:
+        if (not python_use_debian_layout) or (sys.version_info[0] < 3) :
             python_version_xdoty += '.' + str(sys.version_info[1])
```
dirk-thomas: For better readability and in order to not diverge from the CMake implementation please change the lines 243-245 to:

```
python_version_xdoty = str(sys.version_info[0]) + '.' + str(sys.version_info[1]
if python_use_debian_layout and sys.version_info[0] == 3:
    python_version_xdoty = str(sys.version_info[0])
```

smits:
Done.

dirk-thomas:
Thx!

PR #619, opened 2014-04-14
update package manifest to format 2

dirk-thomas:


tfoote:
It needs a run_depend on at least python since it has python scripts for use at runtime.

dirk-thomas:
`python` would be a `depend` since it is required for build time as well as run time. But we don't list `python` in any other ROS package which contains Python scripts so I am not convinced with this PR.

wjwwood:
`catkin_pkg` implicitly brings in the dependency on `Python`, but I don't think it would be wrong for `catkin` to depend on it, but if we want to be consistent with existing packages, then I wouldn't add it.

tfoote:
yeah, i guess it is in our prerequisites list and is implicitly dependend upon. +1

dirk-thomas:
We currently also don't have a `python` rosdep entry - only one for the dev package which is not really required for catkin. I will leave it out for now (as it is for all other packages).

PR #621, opened 2014-04-22
fix catkin_find to not return path with '/.' suffix

dirk-thomas:


wjwwood:
+1

python/catkin/find_in_workspaces.py
```diff
@@ -131,7 +131,9 @@ def find_in_workspaces(search_dirs=None, project=None, path=None, _workspaces=ge
                         packages = find_packages(source_path)
                         matching_packages = [p for p, pkg in packages.items() if pkg.name == project]
                         if matching_packages:
-                            p = os.path.join(source_path, matching_packages[0])
+                            p = source_path
+                            if matching_packages[0] != os.curdir:
+                                p = os.path.join(p, matching_packages[0])
```
tfoote, line 136: This would be easier to read with an if else instead of a reassignment

if matching...
 p = os.path.join(source_path, matched...
else
 p = source_path

And should these be elseifs?

dirk-thomas, line 136: Especially when the expressions get more complex I actually don't like the duplication in both branches. And in this case it is actually doing something and conditionally appending something where the single if logic makes sense.

Therefore I will keep it as is.

dirk-thomas:
Needs to be backported.

python/catkin/find_in_workspaces.py
wjwwood, line 136: I don't mind one style or the other.

Issue #626, opened 2014-04-30
Test that "python setup.py" supports "--install-layout" before using it


dirk-thomas:
Necessary due to this: travis-ci/travis-ci#2231

dirk-thomas:
As mentioned in the ticket this was worked around in the Travis configuration. Since it is unlikely that support for this check is needed I will close the ticket for now.

PR #628, opened 2014-04-30
work around travis using non-debian Python by default

dirk-thomas:


wjwwood:
I think this one is related: https://code.google.com/p/python-nose/issues/detail?id=481

wjwwood:
Well that is a workaround I guess, did you open a ticket against travis? Should you link to it from here?

dirk-thomas:
This is at least a workaround until #626 gets implemented.

dirk-thomas:
Travis won't change it - they will use clean upstream Python from now on - see other ticet for reference.

PR #629, opened 2014-05-03
Only set catkin_* variables when being invoked with find_package COMPONENTS

dirk-thomas:
This is a necessary change to support ros/rosdistro#3010.

@tfoote @wjwwood Please review and consider potential side effects. Is it possible / reasonable to backport this? Otherwise it will force a lot of packages to have a separate indigo branch (which they might not yet have).


tfoote:
+1 tested locally on geometry. 

re backporting: This only fixes c++ unit tests which want the ifdefed tests. @bulwahn has made a good number of pull requests into hydro which have been merged. Geometry appears to be the first one noticed to regress without this patch. The case that will break is if there are people misusing find package and find_package(foo) which is a catkin package and then using the unmatched variables like ${catkin_LIBRARIES}. As this is more of a bug taking advantage of side effects from under the hood it's probably worth backporting. This will allow users to have conditional testing on hydro and not require forking to have it in indigo.

wjwwood:
+1

PR #633, opened 2014-05-16
Add catkin_download functions for build data

trainman419:
Add two functions for downloading data at build time; catkin_download_build_data and catkin_download_extra_data. These are useful for download large data blobs such as binaries or configuration that users don't want to include in their repositories.

This is different from catkin_download_test_data because:
- It always builds, and thus always checks the checksum; re-downloading the file if it has changed
- It isn't hidden behind the CATKIN_ENABLE_TESTING flag
- It's optionally added to the all target so that it runs as part of the normal build
- It adds a REQUIRED option in cmake, which allows the download to fail with a warning, without causing the build to fail completely. This is useful for doing builds when there is no network connection.
- It adds the download_extra_data target, which depends on all of the download targets

Comments welcome.

### DIFF
--- a/cmake/all.cmake
+++ b/cmake/all.cmake
@@ -102,6 +102,7 @@ foreach(filename
     atomic_configure_file
     catkin_add_env_hooks
     catkin_destinations
+    catkin_download
     catkin_generate_environment
     catkin_install_python
     catkin_libraries
--- a/cmake/catkin_download.cmake
+++ b/cmake/catkin_download.cmake
@@ -1,0 +1,150 @@
+#
+# Download a file containing data from a URL.
+#
+# It is commonly used to download larger data files which should not be
+# stored in the repository.
+#
+# .. note:: The target will be registered as a dependency
+#   of the "download_extra_data" target.
+#
+# :param url: the url to download
+# :type url: string
+#
+# :param TARGET: the target name
+#   (default: ${PROJECT_NAME}_download_${the basename of the url})
+# :type TARGET: string
+# :param DESTINATION: the directory where the file is downloaded to
+#   (default: ${CATKIN_DEVEL_PREFIX}/${CATKIN_PACKAGE_SHARE_DESTINATION})
+# :type DESTINATION: string
+# :param FILENAME: the filename of the downloaded file
+#   (default: the basename of the url)
+# :type FILENAME: string
+# :param MD5: the expected md5 hash to compare against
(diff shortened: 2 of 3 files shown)


wjwwood:
We should probably make this against `indigo-devel` and then decide if it should be back ported. Rather than the other way around, but I guess that's up to @dirk-thomas.

dirk-thomas:
First of all it is **not** recommended to download data during the configure / make step of a package. Simply because this prevent building packages without network connectivity.

Regarding the PR: it currently duplicates a lot of code to introduce the two new functions. Instead the download part of the `catkin_download_test_data` function should be extracted and reused.

I have created a new branch and separated the download functionality into `catkin_download` (https://github.com/ros/catkin/tree/catkin_download). Please create a new pull request against that branch to:
- use argparse in the download_checkmd5 script
- add an `OPTIONAL` flag to the download function

The target name should not be optional but should remain required. Also the `EXCLUDE_FROM_ALL` should be rather handled by the user then in the `catkin_download` function.

trainman419:
Ok, I'll make my changes against the `catkin_download` branch.

Because I'm using add_custom_target instead of add_custom_command, I can't base this on the existing catkin_download function. I'm using add_custom_target because I want this to re-run the MD5 sum every time to detect and download new versions of a file, even if the filename hasn't changed.

Note that the rebuilding the download target ONLY requires an internet connection if the target file doesn't exist locally, or if the MD5 sums don't match. If the file exists and the MD5 sums are correct, it doesn't need a network connection.

I agree that for some use cases, having the build fail if it doesn't have a network connection and can't download data isn't desirable, this is what the REQUIRED flag is for. On the other hand, there are cases where the software in question is incomplete and won't run without the downloaded data, so it makes sense to fail the build if the required data isn't on disk and can't be downloaded.

trainman419:
Opened #637 for the chnages to the python script.

dirk-thomas:
Is this pull request still valid and will be updated based on the existing comments or should I close it since it is not required anymore?

trainman419:
I'll update this once #637 is merged. The fix for this will likely also fix #690.

dirk-thomas:
#637 has been merged by recently.

czalidis:
What is the state of this pull request? Is it going to get any updates soon?

Just letting you know that these are some nice new features! I I've been looking a long time for something similar. :+1:

dirk-thomas:
Superseded by #716.

PR #634, opened 2014-05-18
Update cakin Indigo how-to documentation for package.xml format 2

jack-oquin:
Almost all of these changes will also need to be applied to the `groovy-devel` branch used for Groovy and Hydro.

I want to get review approval before merging them there, so let's just focus on the `indigo-devel` branch for now.

For preliminary discussions, see: #440

Formatted version here: http://farnsworth.csres.utexas.edu/docs/catkin/html/


wjwwood:
Nothing jumps out at me after reading over this.

Thanks again for your work on this!

+1

jack-oquin:
There was a comment on ros-users that it would help to have an XML schema for `package.xml`.

That seems like a good idea to me.

mikeferguson:
On the migrating from format 1 line 78 (https://github.com/ros/catkin/pull/634/files#diff-bb1c244b9bba2ff70076ebeecf3f3cc6R78) we have "find_package(rostest REQUIRED)" however, this reset the catkin variables as pointed out by @dirk-thomas here (https://github.com/ros/rosdistro/issues/3010#issuecomment-42093192). Is this still the case? If so this should definitely be called out that you should build any test-related code BEFORE doing find_pacakge(rostest) and destroying your catkin variables (for instance, I had to fix a similar problem in navigation here https://github.com/ros-planning/navigation/commit/72310ffeb80aada04e3df736382e2a168f949fe6)

jack-oquin:
@mikeferguson: that is what I thought Dirk recommended. Someone please correct me if I am wrong about that.

The catkin variables would be reset if we used this, instead:

```
find_package(catkin REQUIRED COMPONENTS rostest)
```

dirk-thomas:
You are right, @jack-oquin.

mikeferguson:
Is package format 2 going to be allowed in hydro? Or is it only for indigo? Because the fix for not resetting catkin variables (https://github.com/ros/catkin/commit/d2c8d4c8dd45d57f6fffb0807effc7a8f00f5858) appears to only  be in the indigo branch of catkin

dirk-thomas:
Both format versions are allowed in all ROS distros since Groovy. The patch has not yet been back ported because there has not been a Hydro/Groovy psych release since it was committed.

doc/howto/format2/installing_python.rst
```diff
@@ -0,0 +1,85 @@
...
+
+Standard ROS practice is to place all your executable Python programs
+in a ``scripts/`` subdirectory.  To keep the user API clean,
+executable script names generally do not include a ``.py`` suffix.
+Your ``CMakeLists.txt`` should install all the scripts explictly
+using the special install function ``catkin_install_python``.
+This will make sure that shebang lines are updated to use the
+specific Python version used at configure time::
+
+  install(PROGRAMS scripts/your_node1 scripts/your_node2
```
vrabaud: that should be catkin_install_python

doc/howto/format2/installing_python.rst
```diff
@@ -0,0 +1,85 @@
...
+Modules
+:::::::
+
+Standard ROS practice is to place Python modules under the
+``src/your_package`` subdirectory, making the top-level module name
+the same as your package.  Python requires that directory to have an
+``__init__.py`` file, too.
+
+  **Note:** with rosbuild, it was possible to place Python modules
+  directly within ``src/``, but catkin does not allow that.  If you
```
vrabaud: catkin does not allow that to follow setup.py conventions.

doc/howto/format2/installing_python.rst
```diff
@@ -0,0 +1,85 @@
...
+This ``setup.py`` is only for use with catkin. Remember *not* to
+invoke it yourself.
+
+Put that script in the top directory of your package, and add this to
+your ``CMakeLists.txt``::
+
+  catkin_python_setup()
+
+That takes care of installing your Python modules.  Never use it to
+install executable scripts, use the ``install()`` command shown above.
```
vrabaud: catkin_python_install

doc/howto/format2/run_tests.rst
```diff
@@ -0,0 +1,44 @@
...
+tests is invoked before.
+
+
+Individual test targets
+:::::::::::::::::::::::
+
+Besides the global ``run_tests`` target each individual `gtest` /
+`nosetest` / `rostest` can be triggered via its own target, e.g.::
+
+  catkin_make run_tests_packagename_gtest_mytest
```
vrabaud, line 20: that can be done with pure make, no need for catkin_make

vrabaud:
made a few comments, looks really good and exhaustive. Only one thing I'd like to see (or that I missed) not to draw people away right up: mention that catkin provides minimal CMake intrusion: for common tasks, only one new macro is invoked (and another one for Python code), as well as some destination variables. Maybe at the top of doc/howto/index.rst ?

doc/howto/format2/installing_python.rst
jack-oquin: In the how-to pages we focus on what users need to do to make things work. 

Rationales and justifications can go somewhere else.

doc/howto/format2/installing_python.rst
jack-oquin: Good catch!

It looks like the syntax is the same, only substituting `catkin_install_python` for `install`. Right?

jack-oquin:
@vrabaud: How about adding those comments here?
- http://farnsworth.csres.utexas.edu/docs/catkin/html/howto/format2/catkin_overview.html#cmakelists-txt

doc/howto/format2/run_tests.rst
jack-oquin, line 20: Yes, but you would have to run `cmake`, `catkin_make` or `catkin build` to create the make files, first.

We don't need to tell people every option, just provide an example that works.

doc/howto/format2/installing_python.rst
dirk-thomas: Right. E.g. https://github.com/ros/ros_comm/blob/indigo-devel/tools/rostopic/CMakeLists.txt#L8

jack-oquin: Thanks, I'll make that change.

tkruse:
+1 in general to this PR. Is catkin_lint also ready for format 2? Might be good to notifiy Timo just in case.

vrabaud:
@jack-oquin  , that seems like the good spot indeed: maybe just something like:

```
Follow them, observing the common CMake commands in the usual order:
```

but then prepend 4,5,10 with "ROS specific: ". It's just that catkin tries hard to stay close to CMake to ease adaptability so it would be sad to not mention it at first sight imho. Thx !

jack-oquin:
@roehling recently added support for the new dependency tags to catkin_lint. It's in master, but not yet released.

jack-oquin:
Thanks to all the reviewers who have participated!

I have tried to address all the comments received so far. If successful, this pull request should be ready to merge. It is at least an improvement over the existing docs, which do not mention package.xml format 2.

@vrabaud, @tkruse, @tfoote, @wjwwood, @dirk-thomas, @mikeferguson, @roehling, and anyone else interested: please check it one more time and vote for or against closing this review and merging the changes.

wjwwood:
+1 Thanks again for helping with this!

dirk-thomas:
+1 Thanks!

roehling:
+1

jack-oquin:
No one has objected. 

I would prefer that this be merged before I start making the corresponding changes to the groovy-devel version, which covers both Groovy and Hydro.

dirk-thomas:
I think we are good to merge then.

@jack-oquin Do you want to squash some commit before the merge?

jack-oquin:
> Do you want to squash some commit before the merge?

I would prefer not to.

dirk-thomas:
Since you will do the merging / cherry-picking it is your call.

Issue #635, opened 2014-05-20
Incorrect PYTHONPATH on Ubuntu 10.04


smits:
Hi all,

I'm trying to use catkin on an older Ubuntu version due to legacy reasons. I've installed python2.7 on it from ppa:fkrull/deadsnakes and made sure I have CMake 2.8.3. I'm using virtualenv to make sure that catkin uses python2.7 since all catkin shebangs are without python version. Although it seems catkin is building ok the generated _setup_util.py exports the wrong PYTHONPATH: 

```
grep PYTHONPATH /opt/ros/hydro/_setup_util.py

        'PYTHONPATH': 'lib/python./site-packages
```

For some reason the MAJOR and MINOR versions are empty. I've skimmed through the catkin sources but cannot find where this CMake variables get set.

dirk-thomas:
The template used for the `_setup_util.py` file uses the CMake variable `PYTHON_INSTALL_DIR` (see https://github.com/ros/catkin/blob/9a5da56c7ecfc456c2cb73124432eaf2fcc482b2/cmake/templates/_setup_util.py.in#L58).

This CMake variable is computed in https://github.com/ros/catkin/blob/9a5da56c7ecfc456c2cb73124432eaf2fcc482b2/cmake/python.cmake#L53 but you should see a message during the configure step which states the Python version being used (https://github.com/ros/catkin/blob/9a5da56c7ecfc456c2cb73124432eaf2fcc482b2/cmake/python.cmake#L27).

There is a second code path for the tools catkin_make / catkin_make_isolated in Python. May be you can add some debug out there if the CMake part is not responsible: https://github.com/ros/catkin/blob/9a5da56c7ecfc456c2cb73124432eaf2fcc482b2/python/catkin/builder.py#L238

wyc:
Please see #636

smits:
The patch provided in #636 fixes my issue!

dirk-thomas:
@smits Could you please provide some information after trying it on your machine why the Python version was not extracted for you? I think it would still be valuable to understand _why_ it failed in the first place - e.g. the proposed patch relies on invoking Python which is performance wise always something to reconsider.

dirk-thomas:
@smits See previous comment.

smits:
Update: I was using indigo-devel instead of groovy-devel. With the original groovy-devel my issue is resolved even without #636

dirk-thomas:
Thank you for clarifying the reason why this happened for you.

Issue #632, opened 2014-05-14
Incorrect PYTHONPATH on gentoo


Antagonistic:
Recently gone though the ROS from source install process for gentoo.  Biggest issue I've hit (so far) is a pathing problem for the installed python scripts.

Running the command:

```
./src/catkin/bin/catkin_make_isolated --install -DCMAKE_BUILD_TYPE=Release --install-space=/opt/ros/hydro
```

installs the catkin (and actually all future) python scripts into:

```
/opt/ros/hydro/lib64/python2.7/site-packages
```

while _setup_util.py adds a path to:

```
lib/python2.7/site-packages
```

which results in a lot of python module not found errors.

Worked around in my case by running:

```
PYTHONPATH="/opt/ros/hydro/lib64/python2.7/site-packages/" ./src/catkin/bin/catkin_make_isolated --install -DCMAKE_BUILD_TYPE=Release --install-space=/opt/ros/hydro
```

As an aside, there are a lot of .so files being installed into /opt/ros/hydro/lib, including opencv, so this seems only a python installation issue.

wjwwood:
Probably a similar to #604 and #545.

wyc:
Narrowed down the issue to these two lines:
https://github.com/ros/catkin/blob/indigo-devel/python/catkin/builder.py#L273
https://github.com/ros/catkin/blob/indigo-devel/cmake/python.cmake#L35

There needs to be a platform-agnostic way of specifying lib vs. lib64. Could we use something like this in both scripts?

```
# gentoo
 $ python -c 'import site; print("/".join(site.getsitepackages()[0].split("/")[-3:]));'
lib64/python2.7/site-packages

# debian
$ python -c 'import site; print("/".join(site.getsitepackages()[0].split("/")[-3:]));'
lib/python2.7/dist-packages
```

The script would have to be Unix-only or amended to support windows too.

Thanks to @wjwwood for suggesting to ask Python.

wyc:
Please see #636

spaghetti-:
still a broken on current kinetic. I can verify that #636 works on Gentoo (`Portage 2.2.28 (python 2.7.10-final-0, default/linux/amd64/13.0, gcc-4.9.3, glibc-2.22-r4, 4.4.6-gentoo x86_64)`).

dirk-thomas:
Since #636 has been closed I just want to mentioned the follow up PR #831 here.

dirk-thomas:
Closing due to inactivity.

[...]

PR #636, opened 2014-05-22
updated PYTHON_INSTALL_DIR generation

wyc:
In response to #632 #635

This patch uses Python to report PYTHON_INSTALL_DIR, and it therefore allows cmake/python.cmake and python/catkin/builder.py to share most of the same code. It eliminates the debian-specific directive and works with python2 and python3. It has been tested on Gentoo, Debian, and Windows.

More testing is in order to confirm its complete functionality.


wyc:
unfortunately the Travis CI build seems to fail due to a shortcoming of python's virtualenv:
https://github.com/pypa/virtualenv/issues/228

wyc:
looks like Travis CI was okay with that one. requesting human review now.

wjwwood:
Thanks for the pull request. As you can probably imagine, we'll have to scrutinize this one pretty hard since it affects all Python packages in ROS.

I'll give some feedback soon.

smits:
I can confirm that the provided patch fixes #635. Thanks!!

spaghetti-:
I would like to mention that this patch did not work for me on gentoo

`Portage 2.2.28 (python 2.7.10-final-0, default/linux/amd64/13.0, gcc-5.4.0, glibc-2.22-r4, 4.4.26-gentoo x86_64)` on kinetic (manual patch)

dirk-thomas:
@spaghetti- Can you please also try the follow up PR #831. And if it doesn't work provide some more information what the problem is, what the various paths are in your case in order to allow iterating on the patch.

spaghetti-:
Yes of course. However @dirk-thomas is the right approach having `lib`/`lib64` both? On my OS X install I only have `lib`.

dirk-thomas:
There can't be more than one location for a single package. Should it needs to be all in either `lib` or `lib64`. Each package can pick a different location though as long the Python path is set to include that path.

dirk-thomas:
I will close this as a duplicate of #831. Please continue the discussion there.

PR #637, opened 2014-05-22
Add flag to download_checkmd5 to ignore network errors

trainman419:
Convert catkin_checkmd5 to argparse and add a flag to make it exit normally if the network connection isn't available.

### DIFF
--- a/cmake/test/download_checkmd5.py
+++ b/cmake/test/download_checkmd5.py
@@ -131,30 +131,39 @@ def main(argv=sys.argv[1:]):
     parser.add_argument("uri")
     parser.add_argument("dest")
     parser.add_argument("md5sum", nargs='?')
+    parser.add_argument("--ignore-error", action="store_true", help="Ignore download errors")
     args = parser.parse_args(argv)
 
     uri = args.uri
-    dest = args.dest
-    md5sum = args.md5sum
 
     if '://' not in uri:
         uri = 'file://' + uri
 
     fresh = False
-    if not os.path.exists(dest):
-        download_md5(uri, dest)
+    if not os.path.exists(args.dest):
+        try:
+            download_md5(uri, args.dest)
+        except Exception:
+            if args.ignore_error:
+                return 0
+            raise
         fresh = True
 
-    if md5sum:
-        result, hexdigest = checkmd5(dest, md5sum)
+    if args.md5sum:
+        result, hexdigest = checkmd5(args.dest, args.md5sum)
         if result is False and fresh is False:
-            print('WARNING: md5sum mismatch (%s != %s); re-downloading file %s' % (hexdigest, md5sum, dest))
-            os.remove(dest)
-            download_md5(uri, dest)
-            result, hexdigest = checkmd5(dest, md5sum)
+            print('WARNING: md5sum mismatch (%s != %s); re-downloading file %s' % (hexdigest, args.md5sum, args.dest))
+            os.remove(args.dest)
+            try:
+                download_md5(uri, args.dest)
+            except Exception:
+                if args.ignore_error:
+                    return 0
+                raise
+            result, hexdigest = checkmd5(args.dest, args.md5sum)
         if result is False:
-            sys.exit('ERROR: md5sum mismatch (%s != %s) on %s; aborting' % (hexdigest, md5sum, dest))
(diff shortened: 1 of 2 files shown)


wjwwood:
+1

cmake/test/download_checkmd5.py
```diff
@@ -127,24 +127,30 @@ def main(argv=sys.argv[1:]):
...
-        uri, dest = args
-    elif len(args) == 3:
-        uri, dest, md5sum = args
-    else:
-        parser.error("wrong number of arguments")
+    parser = ArgumentParser(description="Dowloads URI to file dest. If md5sum is given, checks md5sum. If file existed and mismatch, downloads and checks again")
+    parser.add_argument("uri")
+    parser.add_argument("dest")
+    parser.add_argument("md5sum", nargs='?')
+    parser.add_argument("--non-essential", action="store_true")
```
wjwwood: Should there be a `default=False` parameter here? Maybe the default `default` will work, I'm not sure.

trainman419: The default here is automatically false: https://docs.python.org/2/library/argparse.html#action

wjwwood: Neat, I always thought the default was _always_ `None`.

dirk-thomas: Please add a help text for this argument.

May be additionally a different name would also make the functionality more obvious - may be: `--ignore-error`

cmake/test/download_checkmd5.py
```diff
@@ -127,24 +127,30 @@ def main(argv=sys.argv[1:]):
...
+    args = parser.parse_args(argv)
+
+    uri = args.uri
+    dest = args.dest
+    md5sum = args.md5sum
+    essential = not args.non_essential
```
dirk-thomas: Please avoid these four lines und just use `args.xxx` later in the code.

cmake/test/download_checkmd5.py
```diff
@@ -127,24 +127,30 @@ def main(argv=sys.argv[1:]):
...
     fresh = False
     if not os.path.exists(dest):
-        download_md5(uri, dest)
+        try:
+            download_md5(uri, dest)
+        except Exception:
+            if essential:
+                raise
+            else:
+                sys.exit(0)
```
dirk-thomas: Please avoid the extra `else`. Since the boolean logic is inverted (`non_essential` / `essential`) I would also flip the if / else parts like this:

```
if args.non_essential:
    sys.exit(0)
raise
```

The same several lines below.

dirk-thomas:
Is this pull request still valid and will be updated based on the existing comments or should I close it since it is not required anymore?

trainman419:
I've been busy. I'll update this to follow the proper coding conventions.

trainman419:
I've addressed most of the comments.

I've opted to keep the `uri` variable because it gets modified.

dirk-thomas:
Since the main functions has a lot of different return codes now can you please make one last change. When invoking `main()` please wrap this into `sys.exit()`:

```
sys.exit(main())
```

And in the main function convert all `sys.exit` calls to simple returns.

Please also squash your changes afterwards before merging. Thanks!

trainman419:
k

dirk-thomas:
Please look into the Travis failure.

trainman419:
It looks like the test is failing because it expects main() to call sys.exit()

I guess I'll update the test

trainman419:
Test updated. I'm going to wait for the tests to pass before I squash this.

trainman419:
I thought this way was more pythonic. See https://github.com/ros-visualization/rqt_common_plugins/pull/292

wjwwood:
I would suggest `assertNotEqual(main(...), 0)` since you have an expected value or even `assertEqual(main(...), <some ret value you expect>)`.

I believe it is more pythonic to evaluate a list or tuple without `len(...)`. When evaulating the logical value of a list something specific is expected, but in this case is `main(...)` supposed to return a boolean, an integer, a string (empty or not), a list (empty or not), None, all of these, some subset? It's not clear what you're testing for. That being said, either is probably fine, and pretty easy to understand with some context.

trainman419:
Done. Again.

I've also updated main so that it always returns 0 on success.

wjwwood:
+1

dirk-thomas:
Thank you. Squashed and merged in: 8033824d340d801eb7a9b2a94504b501173aaff9

Issue #638, opened 2014-05-25
catkin_make and friends fail to correctly resolve paths when invoked inside symlinked folders


tkruse:
To reproduce

```
# in basefolder, e.g. /tmp
mkdir origin
mkdir hydro-ws
cd hydro-ws/
ln -s ../origin/ src
cd src
catkin_make -C ..
```

Result:
The specified source space "<basefolder>/src" does not exist

Should be:
Building in basefolder/hydro-ws

The issue was originally discovered by @dirk-thomas in https://github.com/ros/catkin/pull/603 and used as a reason to reject new features. Not sure why it was not ticketed and/or fixed since then, maybe keeping this bug can help reject more undesired pull requests in the future. 
A function used to fix the buggy behavior was also originally given in PR #603, and would be merged in catkin by now if the maintainers had cared.

dirk-thomas:
Thank you for filling the ticket.

The problem is that Python is not aware that it is invoked in `/tmp/hydro-ws/src` as far as I can see. If you run the following command in that folder:

```
python -c "import os; print(os.getcwd())"
```

it outputs `/tmp/origin` as the already resolved symlinked location. I am not aware of any method how Python could get the information that it was actually in `/tmp/hydro-ws/src`. But I think this piece of information would be necessary to resolve the scenario you describe.

If you do know a way how Python can get the required information please share it and I will be happy to resolve the the described bug in the `catkin_make` and `catkin_make_isolated` tools.

tkruse:
As I have said before, I had already solved this in #603 using 

```
python -c "import os; print(os.getenv('PWD'))"
```

https://github.com/tkruse/catkin/commit/de98d064b32eea82b023a2056f25dba4a70edb3a

In catkin/python/catkin/workspace.py

``` Python
def get_cwd_rel_path(rel_or_abspath):
    '''returns the path if absolute, cwd if None, else cwd joined with the given relative path'''
    if rel_or_abspath and os.path.isabs(rel_or_abspath):
        return rel_or_abspath
    # Many, but not all shells, support this var that contains the path preserving symbolic link structures
    pwdpath = os.getenv('PWD')
    if pwdpath is None:
        # this function normalizes the path, does not follow symbolic links
        pwdpath = os.getcwd()
    if rel_or_abspath is not None:
        pwdpath = os.path.join(pwdpath, rel_or_abspath)
    return pwdpath
```

[...]

PR #639, opened 2014-05-25
fix problem when cm / cmi is invoked in a symlinked folder

dirk-thomas:

### DIFF
--- a/bin/catkin_make
+++ b/bin/catkin_make
@@ -35,9 +35,10 @@ def main():
         disable_ANSI_colors()
 
     # verify that the base path is known
-    base_path = os.path.abspath('.')
+    # use PWD in order to work when bein invoked in a symlinked location
+    base_path = os.path.abspath(os.getenv('PWD', '.'))
     if args.directory:
-        base_path = os.path.abspath(args.directory)
+        base_path = os.path.abspath(os.path.join(base_path, args.directory))
     if not os.path.exists(base_path):
         return fmt('@{rf}The specified base path @{boldon}"%s"@{boldoff} does not exist' % base_path)
 
--- a/bin/catkin_make_isolated
+++ b/bin/catkin_make_isolated
@@ -110,10 +110,15 @@ def main():
     if not sys.stdout.isatty():
         opts.no_color = True
 
+    # use PWD in order to work when bein invoked in a symlinked location
+    workspace = os.path.abspath(os.getenv('PWD', '.'))
+    if opts.workspace:
+        workspace = os.path.abspath(os.path.join(workspace, opts.workspace))
+
     destdir = os.environ['DESTDIR'] if 'DESTDIR' in os.environ else None
 
     build_workspace_isolated(
-        workspace=opts.workspace or '.',
+        workspace=workspace,
         sourcespace=opts.source,
         buildspace=opts.build,
         develspace=opts.devel,


tfoote:
+1

tkruse:
Obviously, this issue does not only affect the `-C` argument, but all of:
`catkin_make --build ... --source ... -DCATKIN_DEVEL_PREFIX ... -DCMAKE_INSTALL_PREFIX= ...`
The same arguments where you found this defect in #603

Also I do not understand why this patch does not include a fallback to os.getcwd() for shells not supporting os.getenv('PWD'), as I had explained in the code comment in #638. I have not tried, but this change may simply break on Windows because of that.

Also I would recommend refactoring the codebase such that catkin_make and catkin_make_isolated use the same functions to do the same things, unless you consider code duplication a good practice. I had also already done this in #603 until you asked me to create a separate PR for that, so maybe you prefer to take the time to implement this yourself, or maintain duplicate code.

dirk-thomas:
Can you please describe how you expect the following line of the patch to fail on Windows?

```
os.getenv('PWD', '.')
```

tkruse:
I don't know. Possibly it will work. I would use os.getcwd().

wjwwood:
You could invoke `cd` on Windows: http://stackoverflow.com/questions/921741/windows-equivalent-to-unix-pwd

> Also I would recommend refactoring the codebase such that catkin_make and catkin_make_isolated use the same functions to do the same things, unless you consider code duplication a good practice. I had also already done this in #603 until you asked me to create a separate PR for that, so maybe you prefer to take the time to implement this yourself, or maintain duplicate code.

Please, quit being provocative. "I would recommend refactoring ... to prevent code duplication, like I had in #603 before." would have been sufficient.

tkruse:
Do you need a new github issue for the same problem occurring for
``catkin_make --build ... --source ... -DCATKIN_DEVEL_PREFIX ... -DCMAKE_INSTALL_PREFIX= ...`, or would you maybe reconsider reopening #638?

I am not provocative, I am reporting facts that make me angry. It is not sufficient to be factual, if the problem that needs to be solved extends to OSRF's communication policies. If you think that #603 was handled for the best of the ROS community, say so. If not, who is taking responsibility for it?

tfoote:
@tkruse Please take a look at the fix proposed in #642 

Venting about unrelated topics in issues is not appropriate and is often provocative. Please avoid writing emails and messages while angry as it is not conducive to an open and productive conversation.

tfoote:
Note: '.' will work on POSIX and Windows: https://docs.python.org/2/library/os.html#os.curdir

tkruse:
#642 seems ok

@wjwwood attempted to judge and lecture me here in public. I agree with @tfoote such an action by @wjwwood are not appropriate and often provocative. Just look at where that got you.

Consider taking your own advices more seriously yourselves.

If you feel like you have violated your own code of what is appropriate here and what not, consider taking responsibility.

Also, when you are done lecturing me about how I should communicate, maybe take the time to actually read up on that topic:
- http://hear.ceoblognation.com/2013/02/25/tips-for-dealing-with-the-bad-and-ugly-of-customer-feedback-from-entrepreneurs-and-business-owners/
- http://www.searchenginejournal.com/how-not-to-respond-to-negative-reviews-online-aka-how-to-make-people-hate-your-brand/59903/
- http://mashable.com/2010/02/21/deal-with-negative-feedback/

Feel free to try and respond to my initial feedback in accordance to what professionals suggest. It's a valuable excercise.

dirk-thomas:
@tkruse On repositories which I maintain I will remove any further posts from you (or others) with content which is not strictly on topic (e.g. like your previous comment with these links) or any personally insulting content (as in recent other tickets) since they do harm the ROS ecosystem.

Any person working on an open source project should watch this https://www.youtube.com/watch?v=Q52kFL8zVoM and remember it when interacting with the community.

No further comments will be accepted on this ticket (except of course when being about the pull request itself).

dirk-thomas:
Since I am not aware of any use case which would not work with the proposed patch (even on Windows) I won't take any further actions on this already merged pull request. If there are cases not yet covered please comment with a detailed description of the scenario and a reproducible command sequence.

PR #641, opened 2014-05-27
fix rollback of environment when workspace has been deleted

dirk-thomas:
@jbohren Please verify that this patch address the issue you were able to reproduce in https://github.com/catkin/catkin_tools/issues/47#issuecomment-44228618


wjwwood:
+1

jbohren:
Testing this now.

jbohren:
What is the easiest way to test this without uninstalling catkin?

dirk-thomas:
Just clone catkin into a workspace and build that ontop of what is already installed. It will overlay the installed version of catkin.

Then you can use the same sequence ontop which you used to reproduce the problem.

jbohren:
Yeah it looks like this fixes it!

jbohren:
+1 good to merge

PR #642, opened 2014-05-29
fix argument handling when cm / cmi is invoked in a symlinked folder

dirk-thomas:
@tfoote @wjwwood Please review and try the cm / cmi commands for doubling checking that it doesn't break anything.

### DIFF
--- a/bin/catkin_make
+++ b/bin/catkin_make
@@ -36,7 +36,7 @@ def main():
         disable_ANSI_colors()
 
     # use PWD in order to work when being invoked in a symlinked location
-    cwd = os.getenv('PWD', '.')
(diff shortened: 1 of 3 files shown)


bin/catkin_make
```diff
@@ -34,56 +35,59 @@ def main():
     if args.no_color:
         disable_ANSI_colors()
 
+    # use PWD in order to work when being invoked in a symlinked location
+    cwd = os.getenv('PWD', '.')
```
tfoote: minor: os.curdir would be slightly more generica than '.'

wjwwood:
+1 code review, I haven't run it yet.

wjwwood:
Looks like travis hung, I poked it.

PR #646, opened 2014-06-03
use PYTHON_EXECUTABLE for _setup_util.py

dirk-thomas:
Similar to #645. Keep using the Python executable throughout all locations.

@wjwwood Please review.


wjwwood:
+1

PR #651, opened 2014-06-10
Add --no-push to catkin_prepare_release

po1:
The `catkin_prepare_release` command is useful to format the changelogs and create the tags, but will
fail if no remote repository is configured for the current branch.

Since I could not see a reaason for this limitation, I propose the`--no-push` option, that will do everything locally and not push anything.

This was done for my own convenience. I submit it as a pull request just in case it would be useful to someone else.

### DIFF
--- a/bin/catkin_prepare_release
+++ b/bin/catkin_prepare_release
@@ -345,6 +345,13 @@ def main():
             push_changes(base_path, vcs_type)
         else:
             print(fmt('@{gf}Not pushing because --no-push was given'))
+            try:
+                commands = push_changes(base_path, vcs_type, dry_run=True)
+                print(fmt('@{gf}You can manually use the following commands to push the changes to the remote repository:'))
+                for cmd in commands:
+                    print(fmt('  @{bf}@{boldon}%s@{boldoff}' % ' '.join(cmd)))
+            except RuntimeError as e:
+                print(fmt('@{yf}Warning: manual commands could not be displayed. This can happen if no remote is configured for the current branch.'))
 
     print(fmt("@{gf}The source repository has been released successfully. The next step will be '@{boldon}bloom-release@{boldoff}'."))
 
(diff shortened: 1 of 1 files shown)


bin/catkin_prepare_release
```diff
@@ -332,14 +334,17 @@ def main():
...
+            # confirm commands to push to remote repository
+            commands = push_changes(base_path, vcs_type, dry_run=True)
+            print(fmt('@{gf}The following commands will be executed to push the changes and tag to the remote repository:'))
+            for cmd in commands:
+                print(fmt('  @{bf}@{boldon}%s@{boldoff}' % ' '.join(cmd)))
+            if not args.non_interactive and not prompt_continue('Push the local commits and tags to the remote repository', default=True):
+                raise RuntimeError(fmt('@{rf}Skipping push, to finish the release execute the commands manually.'))
+            push_changes(base_path, vcs_type)
+        else:
+            print(fmt('@{gf}Not pushing because --no-push was given'))
```
dirk-thomas, line 347: Wouldn't it make sense to also print the commands when `--no-push` was passed so that the user just has to copy'n'paste?

po1, line 347: Well if the current branch is not tracking any remote, which happens to be my use case, then I don't see what commands to generate. What I could do is try to run `push_changes(..., dry_run=True)`, silently catch the exception if there is one and display the commands otherwise.

dirk-thomas, line 347: I think it would be valuable to show the command if possible. For your case it could just state that you have to push manually (without and command suggestion).

po1:
@dirk-thomas I added a commit to display the commands to be typed manually when possible.

dirk-thomas:
Can you please refactor the code a bit more to avoid duplicating logic / code?

po1:
I'm not sure what you want me to factorize. I can make the `for` loop that displays commands into a 2-line function if you want, but other than that I can't much see much duplicated code.

dirk-thomas:
Please try #657 and confirm that it fits your use case.

po1:
Works for me, thanks for the help.

PR #654, opened 2014-06-11
ignore cd path echo when using CDPATH, which breaks catkin usage

rhaschke:
cd'ing echos the reached directory to the console when the CDPATH environment variable is used
This breaks the extraction of _CATKIN_SETUP_DIR, because now the directory is listed twice!
see http://superuser.com/questions/90535/how-do-i-turn-of-auto-echo-in-bash-when-i-cd

### DIFF
--- a/cmake/templates/env.sh.in
+++ b/cmake/templates/env.sh.in
@@ -8,6 +8,6 @@ if [ $# -eq 0 ] ; then
 fi
 
 # source @SETUP_FILENAME@.sh from same directory as this file
-_CATKIN_SETUP_DIR=$(cd "`dirname "$0"`" && pwd)
+_CATKIN_SETUP_DIR=$(cd "`dirname "$0"`" > /dev/null && pwd)
 . "$_CATKIN_SETUP_DIR/@SETUP_FILENAME@.sh"
 exec "$@"
--- a/cmake/templates/setup.bash.in
+++ b/cmake/templates/setup.bash.in
@@ -4,5 +4,5 @@
 CATKIN_SHELL=bash
 
 # source setup.sh from same directory as this file
-_CATKIN_SETUP_DIR=$(builtin cd "`dirname "${BASH_SOURCE[0]}"`" && pwd)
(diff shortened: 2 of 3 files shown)


dirk-thomas:
Can you please provide some more information for this?

Whenever I invoke `cd` with an absolute path it won't print anything since it does not use the `CDPATH` variable. Is that behavior different on your system?

Also if your CDPATH actually matches the `cd` folder used in the setup files do you actually have moved the workspace somewhere else and use CDPATH to keep it working?

rhaschke:
Dear Dirk,

you are right, that cd with an absolute path won't print anything. 
However, if you use a relative path, it will print the reached directory.
This proved to be a problem for building some packages (though, not all) 
as well for sourcing the ROS environment, e.g. using source 
install/setup.bash (without an absolute path).

For example, building groovy desktop from source, I get the following 
error when building std_msgs:

```
catkin_generated/env_cached.sh: 12: .: Can't open
*/vol/ros/groovy/build_isolated/std_msgs/catkin_generated**
**/vol/ros/groovy/build_isolated/std_msgs/catkin_generated/setup_cached.sh*
make[1]: *** [CMakeFiles/std_msgs_generate_messages_cpp.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

This error originates in the double output of the path in env.sh, which 
is marked bold above.

On 12.06.2014 03:09, Dirk Thomas wrote:

> Can you please provide some more information for this?
> 
> Whenever I invoke |cd| with an absolute path it won't print anything 
> since it does not use the |CDPATH| variable. Is that behavior 
> different on your system?
> 
> Also if your CDPATH actually matches the |cd| folder used in the setup 
> files do you actually have moved the workspace somewhere else and use 
> CDPATH to keep it working?
> 
> I use CDPATH indepentendly from ROS. I didn't move the workspace. Plain 
> usage of CDPATH to ease hopping to certain (non-ROS) directories.

Kind regards, Robert

---

Dr. Robert Haschke
Faculty of Technology, Neuroinformatics Group, CITEC-3.309
Bielefeld University, 33615 Bielefeld, Germany
phone:  +49-521-106 12122
fax:    +49-521-106 6011
mobile: +49-157-817 365 13
WWW:    www.neuroinformatik.de

PR #656, opened 2014-06-14
Update cakin Groovy and Hydro how-to documentation for package.xml format 2

jack-oquin:
This merges the latest Indigo documentation, appropriately modified, into the branch used for Groovy and Hydro. Most of the changes are needed to support package.xml format 2, but there were some additional updates for errors found in the documentation review. 

previous discussions: #440 and #634

@dirk-thomas, @wjwwood, @tfoote: I doubt another full doc review is needed, but please look this over in case I missed something. Formatted copy is here: http://farnsworth.csres.utexas.edu/docs/catkin/groovy/html/

The only additional commit is 02a4248.


jack-oquin:
I just noticed that the python_version module was removed for Indigo, because it was causing Sphinx errors.

Is it still needed for Groovy and Hydro?

wjwwood:
I'm not sure about the `python_version` module question, but otherwise the docs look good to me.

+1 and thanks again for doing this for us.

dirk-thomas:
Since there is no `python_version` module I think it was safe to be removed from the docs.

If there are no further comments I would go ahead and merge it?

jack-oquin:
If `python_version` does not exist in groovy-devel either, then it _should_ be removed. Otherwise, sphinx would issue an error.

Yes, please merge now.

PR #658, opened 2014-06-15
flush stdin before asking for input

dirk-thomas:
Same as ros-infrastructure/bloom#249.

@wjwwood Please review.

### DIFF
--- a/bin/catkin_prepare_release
+++ b/bin/catkin_prepare_release
@@ -37,6 +37,7 @@ def prompt_continue(msg, default):
         msg += fmt(' @{yf}[y/N]@{reset}?')
 
     while True:
+        _flush_stdin()
         response = raw_input(msg)
         if not response:
             response = 'y' if default else 'n'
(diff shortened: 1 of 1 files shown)


wjwwood:
So I noticed in bloom that sometimes the messages which are printed to stdout before one of these prompts never gets displayed. I'm not sure I can reproduce it, but it has definitely happened.

dirk-thomas:
But do you think it is related to the stdin flushing?

wjwwood:
I do actually, this is an example of what I am talking about:

```
==> Generating pull request to distro file located at 'https://raw.github.com/ros/rosdistro/master/hydro/distribution.yaml'
Would you like to add source information for this repository? [Y/n]? n
Would you like to add a maintenance status for this repository? [Y/n]?
Please enter a maintenance status.
Valid maintenance statuses:
- developed: active development is in progress
Status:
```

There should be two other `-` prefixed options before the `Status:` line.

Maybe we should flush stdout before truncating the stdin? I can try that locally and see if that make a difference.

bin/catkin_prepare_release
```diff
@@ -49,6 +50,15 @@ def prompt_continue(msg, default):
         print(fmt("@{yf}Response '@{boldon}%s@{boldoff}' was not recognized, please use one of the following options: %s" % (response, ', '.join([('@{boldon}%s@{boldoff}' % x) for x in ['y', 'Y', 'n', 'N']]))), file=sys.stderr)
 
 
+def _flush_stdin():
+    try:
+        from termios import tcflush, TCIOFLUSH
+        tcflush(sys.stdin, TCIOFLUSH)
```
dirk-thomas: Actually shouldn't this line be:

```
tcflush(sys.stdin, TCIFLUSH)
```

dirk-thomas:
@wjwwood Can you please give it a try with my previous comment on changing `TCIOFLUSH` to `TCIFLUSH`?

wjwwood:
I'll give it a try now.

wjwwood:
+1

PR #662, opened 2014-06-17
make nosetests --xunit-file argument an absolute path to work around nose bug 779

### LINKED ISSUE #659: catkin_add_nosetests fail to run on Trusty in certain configurations due to nosetest bug
This is actually an upstream bug introduced into nosetest 1.3.1 (which ships with Trusty), see https://github.com/nose-devs/nose/issues/779 and the fix in https://github.com/nose-devs/nose/commit/ac9cc08e1a95e5614c8f9863091014186edecbdd. It appears this doesn't break the "catkin_make run_tests" use case, but it does appear to break https://github.com/ros-infrastructure/jenkins_scripts/blob/master/test_repositories.py#L182 (I'm not sure, but I think this is still how devel jobs are run, right?). Basically, the failure looks like this:

```
-- run_tests.py: execute commands
  /usr/bin/cmake -E make_directory ../test_results/ubr_drivers
  /usr/bin/nosetests-2.7 -P --process-timeout=60 --where=/home/fergs/tmp/fix-tests/src/test_package/test/python --with-xunit --xunit-file=../test_results/test_package/nosetests-python.xml
.....
  File "/usr/bin/nosetests-2.7", line 9, in <module>
    load_entry_point('nose==1.3.1', 'console_scripts', 'nosetests-2.7')()
  File "/usr/lib/python2.7/dist-packages/nose/core.py", line 121, in __init__
    **extra_args)
  File "/usr/lib/python2.7/unittest/main.py", line 95, in __init__
    self.runTests()
  File "/usr/lib/python2.7/dist-packages/nose/core.py", line 207, in runTests
    result = self.testRunner.run(self.test)
  File "/usr/lib/python2.7/dist-packages/nose/core.py", line 66, in run
    result.printErrors()
  File "/usr/lib/python2.7/dist-packages/nose/result.py", line 110, in printErrors
    self.config.plugins.report(self.stream)
  File "/usr/lib/python2.7/dist-packages/nose/plugins/manager.py", line 99, in __call__
    return self.call(*arg, **kw)
  File "/usr/lib/python2.7/dist-packages/nose/plugins/manager.py", line 167, in simple
    result = meth(*arg, **kw)
  File "/usr/lib/python2.7/dist-packages/nose/plugins/xunit.py", line 199, in report
    self.encoding, 'replace')
  File "/usr/lib/python2.7/codecs.py", line 878, in open
    file = __builtin__.open(filename, mode, buffering)
IOError: [Errno 2] No such file or directory: '../test_results/test_package/nosetests-python.xml'
```

Hopefully we can actually get this patched upstream in Ubuntu, but I wanted to open a ticket because it took me a while to track this down. Also, I have a somewhat hacky patch that replaces the xunit-file path with an absolute one (https://github.com/mikeferguson/catkin/commit/62e5f68a5def68f6ca5dbdb4267abbffa1513531) that we could use if an upstream fix isn't possible. This doesn't change the catkin_make use case results at all from what I can see, but does make the test job style use case work.

dirk-thomas:
Resolves #659.

### DIFF
--- a/cmake/test/nosetests.cmake
+++ b/cmake/test/nosetests.cmake
@@ -70,6 +70,8 @@ function(catkin_add_nosetests path)
   string(REPLACE ":" "." output_file_name ${output_file_name})
 
   set(output_path ${CATKIN_TEST_RESULTS_DIR}/${PROJECT_NAME})
+  # make --xunit-file argument an absolute path (https://github.com/nose-devs/nose/issues/779)
+  get_filename_component(output_path "${output_path}" ABSOLUTE)
   set(cmd "${CMAKE_COMMAND} -E make_directory ${output_path}")
   if(IS_DIRECTORY ${_path_name})
     set(tests "--where=${_path_name}")


dirk-thomas:
Confirmed to work by @mikeferguson

PR #663, opened 2014-06-20
add imported library targets for find_package'ed gtest

v4hn:
This makes catkin gtests work again with gtest 1.5 as well as 1.6+ on Arch
Linux and others.


dirk-thomas:
Can you please describe why the proposed change would change the behavior for a different gtest version? It looks like it should be unrelated to that - may be the reason is something else?

v4hn:
Sure,

gtest 1.5 and older officially supported installed libraries.
This changed with 1.6 and distributions took up different stances on that.

Debian/Ubuntu decided to resolve this by installing the gtest sources to
/usr/include/gtest and /usr/src/gtest. They provide distribution-specific
CMakeLists.txt to allow projects to use these sources.

ArchLinux and likely others still install the libraries and provide
a FindGTest.cmake with cmake (it's in cmake's official Modules section).
This cmake is basically broken in Ubuntu because there is no library
left to find.

Therefore find_package(GTest QUIET) only succeeds on systems with gtest 1.5
or older and systems taking the second stance. Whereas the first branch of the
if(not gtest_found)-block is used on Ubuntu (this one defines the gtest and
gtest_main targets), other systems use the respective else-block.
In this block, however, the two targets are never defined.

Because the catkin_add_gtest-function expects the gtest and gtest_main targets
to be around and other modules (e.g. laser_filters) directly link against
them, a standard indigo workspace currently produces lots of warnings
and failures in my case.

The proposed patch imports these targets from the installed libraries
and therefore makes them available in the same way as they are available
to catkin-modules on Ubuntu.

v4hn

v4hn:
A quick follow-up on the explanation:
It would be great to see this fixed in the upcoming official indigo-release.
This is the last custom fix I have to apply over here to successfully compile indigo's desktop_full package on Lunar-Linux.

Issue #665, opened 2014-06-25
catkin_make_isolated crashes if first package doesn't create devel dir


tfoote:
```
tecting C compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Configuring done
-- Generating done
-- Build files have been written to: /home/tfoote/work/indigo/workspaces/rosconsole/build_isolated/console_bridge/devel
==> make -j8 -l8 in '/home/tfoote/work/indigo/workspaces/rosconsole/build_isolated/console_bridge/devel'
==> make install in '/home/tfoote/work/indigo/workspaces/rosconsole/build_isolated/console_bridge/devel'
Install the project...
-- Install configuration: ""
Phony install target
==> Generating an env.sh
Unhandled exception of type 'OSError':
Traceback (most recent call last):
  File "/opt/ros/indigo/lib/python2.7/dist-packages/catkin/builder.py", line 875, in build_workspace_isolated
    number=index + 1, of=len(ordered_packages)
  File "/opt/ros/indigo/lib/python2.7/dist-packages/catkin/builder.py", line 641, in build_package
    destdir=destdir
  File "/opt/ros/indigo/lib/python2.7/dist-packages/catkin/builder.py", line 521, in build_cmake_package
    os.mkdir(os.path.dirname(new_env_path))
OSError: [Errno 2] No such file or directory: '/home/tfoote/work/indigo/workspaces/rosconsole/devel_isolated/console_bridge'
<== Failed to process package 'console_bridge': 
  [Errno 2] No such file or directory: '/home/tfoote/work/indigo/workspaces/rosconsole/devel_isolated/console_bridge'
Command failed, exiting.
tfoote@yeti:~/work/indigo/workspaces/rosconsole$ ls 
build_isolated  src
tfoote@yeti:~/work/indigo/workspaces/rosconsole$ mkdir -p devel_isolated/console_bridge
tfoote@yeti:~/work/indigo/workspaces/rosconsole$ catkin_make_isolated 
Base path: /home/tfoote/work/indigo/workspaces/rosconsole
Source space: /home/tfoote/work/indigo/workspaces/rosconsole/src
Build space: /home/tfoote/work/indigo/workspaces/rosconsole/build_isolated
Devel space: /home/tfoote/work/indigo/workspaces/rosconsole/devel_isolated
[...]

PR #666, opened 2014-06-25
fix handling of cmake packages which do not install any files

dirk-thomas:
@tfoote Please verify that it fixes your problem reported in #665.


wjwwood:
+1

tfoote:
+1 I confirmed it fixed the workspace from #665

Issue #667, opened 2014-06-27
fix quoting of error messages in catkin/cmake/em/order_packages.cmake.em


rhaschke:
I get the following order_packages.cmake:
`# generated from catkin/cmake/em/order_packages.cmake.em
message(FATAL_ERROR "Invalid package manifest "/homes/rhaschke/src/ros/rtt/install_isolated/lib/ruby/1.9.1/x86_64-linux/orogen/templates/typekit/package.xml": The manifest contains invalid XML:
not well-formed (invalid token): line 2, column 9")
`

which triggers the cmake error:

`Argument not separated from preceding token by whitespace.`

tfoote:
Note i believe this is an orogen package.xml which is different from a ros one.

rhaschke:
I know, that the package.xml is wrong. However, catkin should better 
deal with catkin-generated exception that contains double-quotes.
These shouldn't be written literally into the order_packages.cmake file.

dirk-thomas:
Could it be that the message you are referring to ("Argument not separated from preceding token by whitespace.") is only a CMake warning and that you are using CMake 2.8.12 (where this warning has been introduced)?

Can you please provide more information to reproduce the problem? What exactly is in your workspace and what command are you invoking?

rhaschke:
Indeed, I'm using cmake 2.8.12.2 on Ubuntu Trusty. The message is only a 
warning - but annoying. It originates from a non-ROS package.xml (attached).
I'm suggesting to fix the error handling routine in catkin, that finally 
generated the error message indicating the malformed package.xml.

Kind regards, Robert

On 29.06.2014 02:43, Dirk Thomas wrote:

> Could it be that the message you are referring to ("Argument not 
> separated from preceding token by whitespace.") is only a CMake 
> warning and that you are using CMake 2.8.12 (where this warning has 
> been introduced)?
> 
> Can you please provide more information to reproduce the problem? What 
> exactly is in your workspace and what command are you invoking?
> 
> —
> Reply to this email directly or view it on GitHub 
> https://github.com/ros/catkin/issues/667#issuecomment-47442615.

## 

Robert Haschke

---

Dr. Robert Haschke
Faculty of Technology, Neuroinformatics Group, CITEC-3.309
Bielefeld University, 33615 Bielefeld, Germany
phone:  +49-521-106 12122
fax:    +49-521-106 6011
mobile: +49-157-817 365 13
WWW:    www.neuroinformatik.de

[...]

PR #668, opened 2014-06-29
escape messages to avoid CMake warning

dirk-thomas:
Related to #667 but I can't reproduce that one since it fails much earlier in catkin_make when trying to parse the invalid xml.

Anyway the error message should be escaped to avoid a CMake warning (with 2.8.12 and newer).

@tfoote @wjwwood Please review


tfoote:
Is str(e) unicode safe?

wjwwood:
@tfoote pretty much the only thing safe to do when the thing your are templating in may or may not be unicode is to use `from __future__ import unicode_literals` and use formatting like `"%s" % <thing>` always, never use `str()` or `.__str__()` to convert the thing into a string.

dirk-thomas:
Actually there is at least once case where the exception might contain unicode characters: when it raises an invalid email where the message includes the email address as well as the name of the person.

Anyway I have updated the PR to avoid using `str()` but full unicode support is still left to be done (if someone cares).

wjwwood:
+1

PR #669, opened 2014-06-30
use local _argn instead of ARGN for efficiency

aleeper:
It seems in a macro doing something like

```
message("${ARGN}")
```

is **_much**_ slower than

```
set(_argn ${ARGN})
message("${_argn}")
```

I don't know why, but this fixes it in the pack/unpack macros that get called a lot.

**This one change took average configuration time on a certain project from ~115 seconds to 29.**


dirk-thomas:
Please provide an example to reproduce your measurement.

aleeper:
I can't share the project that resulted in the above measurement, though the project makes heavy use of catkin_simple which likely magnifies the problem since there are more libraries that end up in the giant string.

Anyway, here is a easily reproducible case that everyone has access to, and still shows a 17% time savings:
https://docs.google.com/spreadsheets/d/1ZRMTk5L5l5dzhrAwHspe-JMGvWQWsSrZYRS3A4q6Ks4/edit?usp=sharing

dirk-thomas:
I have done some more testing for this.

The MoveIt example greatly benefits from the patch. But in other CMake files I tried the introduction of a separate variable for `ARGN` should either no or sometimes negative impact.

I think we can achieve the major part of your improvement by either:
- disabling the `debug_message()` lines since they implode the `ARGN` list (which is pretty large in the MoveIt case) into a string which is then (if not configured with a custom log level) not being used anyway or
- use the local `_argn` variable for those `debug_message()` calls (while it remains unclear why that actually performs better - I was not able to reproduce that effect in any synthetic CMake file I tried to construct)

dirk-thomas:
I have created a patch for your PR: https://github.com/hiDOF-forks/catkin/pull/2

If that looks good to you too please merge it and squash the commits. Then I will merge this one.

dirk-thomas:
I did the merge / squash / cherry-pick manually in e4c113be2708a8936c5defd7a715c7b9cb5f7595.

Thanks for the patch.

aleeper:
Awesome, thanks!

Issue #670, opened 2014-07-02
".catkin" marker file can be overwritten if two catkin instances are running with a shared develspace


jbohren:
For example, when using `catkin build` from [catkin_tools](http://github.com/catkin/catkin_tools), there is a chance that some package source directories will not be listed in the `.catkin` file. If one package A writes it after another package B has read it, but before B has written to it, then A's modifications will be overwritten when B writes to it. This, in turn, means that they will not enter into `$ROS_PACKAGE_PATH`.

This was formerly discovered and discussed in catkin/catkin_tools#77.

dirk-thomas:
This requires to make the modification to file an atomic operation. Please consider to provide a pull request for this.

As a side note: https://github.com/ros/catkin/blob/indigo-devel/cmake/atomic_configure_file.cmake does something similar but since it does only overwrite the file (rather then extending it) the approach will not work for this case.

dirk-thomas:
I will close this with #671 merged and also being released in all distros.

The tickets (#674, #675) track separate enhancements.

PR #671, opened 2014-07-02
use atomic APPEND write mode for writing source spaces to .catkin marker file

jbohren:
There is no significance to the order of the packages written to the `.catkin` file, so instead of reading/inserting/writing, this patch uses a single `file(APPEND ...)` command to modify the file. While not bullet-proof, this significantly decreases the probability that the problem described in #670 / catkin/catkin_tools#77 occurs.

### DIFF
--- a/cmake/all.cmake
+++ b/cmake/all.cmake
@@ -16,19 +16,23 @@ else()
 endif()
 message(STATUS "Using CATKIN_DEVEL_PREFIX: ${CATKIN_DEVEL_PREFIX}")
 
-# create workspace marker
-set(_sourcespaces "${CMAKE_SOURCE_DIR}")
-if(EXISTS "${CATKIN_DEVEL_PREFIX}/.catkin")
-  # prepend to existing list of sourcespaces
-  file(READ "${CATKIN_DEVEL_PREFIX}/.catkin" _existing_sourcespaces)
-  list(FIND _existing_sourcespaces "${CMAKE_SOURCE_DIR}" _index)
-  if(_index EQUAL -1)
-    list(INSERT _existing_sourcespaces 0 ${CMAKE_SOURCE_DIR})
-  endif()
-  set(_sourcespaces ${_existing_sourcespaces})
-endif()
-file(WRITE "${CATKIN_DEVEL_PREFIX}/.catkin" "${_sourcespaces}")
+# update develspace marker file with a reference to this sourcespace
+set(_catkin_marker_file "${CATKIN_DEVEL_PREFIX}/.catkin")
 
+# check if the develspace marker file exists yet
+if(EXISTS ${_catkin_marker_file})
+  # append to existing list of sourcespaces 
+  file(READ ${_catkin_marker_file} _existing_sourcespaces)
+  list(FIND _existing_sourcespaces "${CMAKE_SOURCE_DIR}" _existing_sourcespace_index)
+  if(_existing_sourcespace_index EQUAL -1)
+    file(APPEND ${_catkin_marker_file} ";${CMAKE_SOURCE_DIR}")
+  endif()
+else()
+  # create a new develspace marker file
+  # NOTE: extra care must be taken when running multiple catkin jobs in parallel 
+  #       so that this does not overwrite the result of a similar call in another package
+  file(WRITE ${_catkin_marker_file} "${CMAKE_SOURCE_DIR}")
+endif()
 
 # use either CMAKE_PREFIX_PATH explicitly passed to CMake as a command line argument
 # or CMAKE_PREFIX_PATH from the environment


jbohren:
Additionally, this avoids having to perform the string concatenation for the file path while reading/writing the file. This might explain why this was easier to reproduce with long paths than it was in /tmp with a minimal example.

wjwwood:
Awesome, I didn't know that `file(APPEND ...)` was atomic, is that in the CMake docs? I was momentarily distraught because I thought we were going to have to move to some sort of file based locking.

One case I'm not sure about is what if two instances check if the file exists and then both decide to create it for the first time?

dirk-thomas:
You can use `file(APPEND ...)` to create the file in the first place. But I would also be curious why you think it is an atomic operation?

wjwwood:
Ok, so that addresses the race condition (with a small code change). But I am still concerned about the atomic nature of `file(APPEND ...)`.

jbohren:
> Ok, so that addresses the race condition (with a small code change). But I am still concerned about the atomic nature of `file(APPEND ...)`.

Well `APPEND` causes CMake to do this ( [cmFileCommand.cxx#L232](https://github.com/Kitware/CMake/blob/master/Source/cmFileCommand.cxx#L232) )

``` cpp
  cmsys::ofstream file(fileName.c_str(), std::ios::app);
```

And prevailing consensus on the interscape is that at least on UNIX this sort of operation is atomic as long as you're writing less than 4k: http://stackoverflow.com/questions/1154446/is-file-append-atomic-in-unix

wjwwood:
Ok, well that sounds good enough to me. At any rate it should be much better than before.

We should still remove the conditional logic for first time writing, so that it uses `file(APPEND ...)` always.

Thanks for digging into this @jbohren, it's a nice catch.

jbohren:
> We should still remove the conditional logic for first time writing, so that it uses file(APPEND ...) always.

Well `file(APPEND...)` won't create the file, I'm trying to use `execute_process(COMMAND cmake -E touch ...)` but it's not liking that for some reason.

wjwwood:
It creates a file for me:

```
% ls
test.cmake

% cat test.cmake
file(APPEND test.out ";TEST")

% cmake -P test.cmake

% ls
test.cmake test.out

% cat test.out
;TEST%
```

jbohren:
> It creates a file for me:

Yeah no issues there, but it appears the `touch` command wasn't creating something that CMake wanted to read. Regardless, I've just updated it so there's only a single `APPEND` call.

wjwwood:
Looks good to me.

dirk-thomas:
To be backported to _groovy-devel_.

dirk-thomas:
`file(APPEND...)` does actually create the file for me. What CMake version are you using?

wjwwood:
I think we all agree that works now, the pull request uses it.

jbohren:
`file (APPEND)` makes the file for me too, it was the `cmake -E touch` command that wasn't working.

dirk-thomas:
Sorry for the noise - I was reading the comments wrong.

One more question though - are we sure that all code reading this file is fine with the leading semicolon introduced by this patch?

jbohren:
> I think we all agree that works now, the pull request uses it.

Right.

jbohren:
> One more question though - are we sure that all code reading this file is fine with the leading semicolon introduced by this patch?

Worst case is that a split function returns an array with a leading empty element. As far as I can tell, the only other code that reads this file is the roslib env hook. The effect there is that it puts a leading colon into `ROS_PACKAGE_PATH`. 

I don't see that causing any problems, but we could update the env hook to remove that.

https://github.com/ros/ros/blob/indigo-devel/core/roslib/env-hooks/10.ros.sh.em

dirk-thomas:
I think we need to collect all code locations in this ticket which use this file. E.g. the roslisp environment hook also uses it.

jbohren:
> I think we need to collect all code locations in this ticket which use this file. E.g. the roslisp environment hook also uses it.

We meaning us, me, you, or someone else?

dirk-thomas:
It would be great if someone else can put together this list. I can for sure do that but it will take some time until I will find time for that.

Since this is kind of an API change (since other code parts rely on the file content structure). So it should be covered in the PR if all (known) "caller" are fine with the change.

jbohren:
Is there anyone other than roslib and roslisp actually reading this file?  There's no other mention of it in all of my packages installed via the debs.

dirk-thomas:
I actually don't know - the only way to find out might be search for it in the sources.

jbohren:
> I actually don't know - the only way to find out might be search for it in the sources.

Yeah, searching for ".catkin" on github is equivalent to searching for "catkin". It comes up in a few places.

jbohren:
So the leading or trailing separator in a PATH-like variable can sometimes be interpreted as the current working directory. This is not what we want.

I propose that we keep the formatting of the file the same by writing just the name of the source-space if the file doesn't exist yet. i.e.:

``` cmake
# update develspace marker file with a reference to this package
set(_catkin_marker_file "${CATKIN_DEVEL_PREFIX}/.catkin")

# append to existing list of sourcespaces if the develspace marker file exists
if(EXISTS ${_catkin_marker_file})
  file(READ ${_catkin_marker_file} _existing_sourcespaces)
  list(FIND _existing_sourcespaces "${CMAKE_SOURCE_DIR}" _existing_sourcespace_index)
  if(_existing_sourcespace_index EQUAL -1)
    file(APPEND ${_catkin_marker_file} ";${CMAKE_SOURCE_DIR}")
  endif()
else()
  # create a new develspace marker file
  # NOTE: extra care must be taken when running multiple catkin jobs in parallel 
  #       so that this does not overwrite the result of a similar call in another package
  file(WRITE ${_catkin_marker_file} "${CMAKE_SOURCE_DIR}")
endif()
```

This will still dramatically reduce the space for a race condition. Then over in catkin/catkin_tools/issues/77 we add a check at the end of the execution of `catkin build` to make sure that the number of packages in the `.catkin` file is correct. If it is not correct, then all the user needs to do is build again. The race condition can only occur if the catkin marker file doesn't exist.

The `catkin build` command could also attempt to repair this file if necessary.

dirk-thomas:
Sounds like a good compromise to me.

wjwwood:
I guess that's ok, I would prefer a solution where failure is not statistically possible, but I don't see another way of doing it which isn't really complicated.

jbohren:
> I guess that's ok, I would prefer a solution where failure is not statistically possible, but I don't see another way of doing it which isn't really complicated.

Going forward, we could change the format so that it's separated by line breaks, or alternatively, having each package write to a separate file.

jbohren:
Let me tweak one last thing before this gets merged.

dirk-thomas:
The "resource index" (https://github.com/wjwwood/package_resource_indexer) proposed for ROS 2 does exactly that - separate the information in separate files. But I think that might be out-of-scope to apply to the current catkin version.

wjwwood:
Yeah, I was thinking of suggesting that a CMake option could make each package write to a `.catkin_<package_name>` and then only invoke that when running in parallel, then afterwards the tool running the parallel jobs (`catkin build`) could create one `.catkin` file from the aggregate.

jbohren:
> Yeah, I was thinking of suggesting that a CMake option could make each package write to a .catkin_<package_name> and then only invoke that when running in parallel, then afterwards the tool running the parallel jobs (catkin build) could create one .catkin file from the aggregate.

Yeah, or even make a `.catkin_source_spaces` directory in the develspace with a metadata file for each source space.

jbohren:
> Yeah, I was thinking of suggesting that a CMake option 

Sorry I missed that part. Yeah, I think that would be great, like a "CATKIN_THREAD_SAFE" or something.

jbohren:
This should be good to merge.

cmake/all.cmake
```diff
@@ -16,19 +16,23 @@ else()
...
 
-# create workspace marker
-set(_sourcespaces "${CMAKE_SOURCE_DIR}")
-if(EXISTS "${CATKIN_DEVEL_PREFIX}/.catkin")
-  # prepend to existing list of sourcespaces
-  file(READ "${CATKIN_DEVEL_PREFIX}/.catkin" _existing_sourcespaces)
-  list(FIND _existing_sourcespaces "${CMAKE_SOURCE_DIR}" _index)
-  if(_index EQUAL -1)
-    list(INSERT _existing_sourcespaces 0 ${CMAKE_SOURCE_DIR})
+# update develspace marker file with a reference to this package
```
dirk-thomas: It is not "to this package" but "to the source space".

wjwwood:
+1 from me, we can think about more sophisticated solutions separately.

cmake/all.cmake
jbohren: Changed to "this package's sourcespace"

dirk-thomas: This is not the package level. It is the source space of the workspace which might contain multiple packages. Therefore it should not mention "package" in here.

jbohren: Ahhh right. Sorry, I've been in `catkin build` land for too long. Fixed.

dirk-thomas:
Thanks for iterating on this so fast.

cmansley:
See ros/catkin#752. 

After this discussion, all writes should be appends. Leaving this note for future explorers.

PR #672, opened 2014-07-02
Use atomic APPEND write mode for writing source spaces to .catkin marker file

### LINKED ISSUE #670: ".catkin" marker file can be overwritten if two catkin instances are running with a shared develspace

jbohren:
Same as #671 but fixes #670 for groovy/hydro.

### DIFF
--- a/cmake/all.cmake
+++ b/cmake/all.cmake
@@ -16,19 +16,18 @@ else()
 endif()
 message(STATUS "Using CATKIN_DEVEL_PREFIX: ${CATKIN_DEVEL_PREFIX}")
 
-# create workspace marker
-set(_sourcespaces "${CMAKE_SOURCE_DIR}")
-if(EXISTS "${CATKIN_DEVEL_PREFIX}/.catkin")
-  # prepend to existing list of sourcespaces
-  file(READ "${CATKIN_DEVEL_PREFIX}/.catkin" _existing_sourcespaces)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I am closing this since any change going to the _groovy-devel_ branch will be cherry-picked from the _indigo-devel_ branch whenever possible.

jbohren:
Sounds good.

jbohren:
> I am closing this since any change going to the groovy-devel branch will be cherry-picked from the indigo-devel branch whenever possible.

This never got cherry-picked, and people are running into issues with it using `catkin build`, any chance it could get pulled into groovy/hydro?

dirk-thomas:
It looks to me the patch has been cherry-picked to the _groovy-devel_ branch in https://github.com/ros/catkin/commit/f23396d11e6d459937ebfaf68eba47832bcf9f6f.

Also new version of catkin (0.5.89) has been released into _Groovy_ and _Hydro_ since then.

PR #676, opened 2014-07-09
catkin_generate_environment: don't overwrite existing .catkin file

jbohren:
The Orocos RTT package, a pure CMake package, currently doesn't _use_ catkin for building, but it does call `find_package(catkin)` in order to determine if catkin is installed on the system, and if so, it calls `catkin_add_env_hooks()` to add an env hook. This works fine if you're installing it to an isolated workspace or somewhere else, but when installing to a develspace (like with `catkin_tools`), it causes the `.catkin` file to be overwritten and you lose the list of all previously-built packages.

This patch replaces the normal CMake `install(FILES ...)` command with an `install(CODE ...)` command which just checks if the file already exists or not.

### DIFF
--- a/cmake/catkin_generate_environment.cmake
+++ b/cmake/catkin_generate_environment.cmake
@@ -75,11 +75,12 @@ function(catkin_generate_environment)
   set(SETUP_DIR ${CMAKE_INSTALL_PREFIX})
 
   if(NOT CATKIN_BUILD_BINARY_PACKAGE)
-    # generate and install workspace marker
-    file(WRITE ${CMAKE_BINARY_DIR}/catkin_generated/installspace/.catkin "")
-    install(FILES
-      ${CMAKE_BINARY_DIR}/catkin_generated/installspace/.catkin
-      DESTINATION ${CMAKE_INSTALL_PREFIX})
+    # install empty workspace marker if it doesn't already exist
+    install(CODE " 
+      if (NOT EXISTS \"\${CMAKE_INSTALL_PREFIX}/.catkin\") 
+        file(WRITE \"\${CMAKE_INSTALL_PREFIX}/.catkin\" \"\") 
+      endif()") 
+
     # generate and install Python setup util
     configure_file(${catkin_EXTRAS_DIR}/templates/_setup_util.py.in
       ${CMAKE_BINARY_DIR}/catkin_generated/installspace/_setup_util.py
--- a/cmake/all.cmake
+++ b/cmake/all.cmake
@@ -21,11 +21,16 @@ set(_catkin_marker_file "${CATKIN_DEVEL_PREFIX}/.catkin")
 
 # check if the develspace marker file exists yet
 if(EXISTS ${_catkin_marker_file})
-  # append to existing list of sourcespaces 
   file(READ ${_catkin_marker_file} _existing_sourcespaces)
-  list(FIND _existing_sourcespaces "${CMAKE_SOURCE_DIR}" _existing_sourcespace_index)
-  if(_existing_sourcespace_index EQUAL -1)
-    file(APPEND ${_catkin_marker_file} ";${CMAKE_SOURCE_DIR}")
+  if(_existing_sourcespaces STREQUAL "")
+    # write this sourcespace to the marker file
+    file(WRITE ${_catkin_marker_file} "${CMAKE_SOURCE_DIR}")
+  else()
+    # append to existing list of sourcespaces if it's not in the list
+    list(FIND _existing_sourcespaces "${CMAKE_SOURCE_DIR}" _existing_sourcespace_index)
+    if(_existing_sourcespace_index EQUAL -1)
+      file(APPEND ${_catkin_marker_file} ";${CMAKE_SOURCE_DIR}")
+    endif()
   endif()
 else()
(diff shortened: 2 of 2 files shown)


dirk-thomas:
Can you please update the PR to only write an empty file within the `install(CODE ...)` block? It shouldn't do an `install` in there and it doesn't require the temporary in build space anymore.

jbohren:
> Can you please update the PR to only write an empty file within the install(CODE ...) block? It shouldn't do an install in there and it doesn't require the temporary in build space anymore.

Yeah that makes more sense. Done.

cmake/catkin_generate_environment.cmake
```diff
@@ -76,10 +76,14 @@ function(catkin_generate_environment)
 
   if(NOT CATKIN_BUILD_BINARY_PACKAGE)
     # generate and install workspace marker
-    file(WRITE ${CMAKE_BINARY_DIR}/catkin_generated/installspace/.catkin "")
-    install(FILES
-      ${CMAKE_BINARY_DIR}/catkin_generated/installspace/.catkin
-      DESTINATION ${CMAKE_INSTALL_PREFIX})
+    set(_catkin_marker_file_installed "${CMAKE_INSTALL_PREFIX}/.catkin")
```
dirk-thomas: I am not sure that it works but have you tried to move the variable into the `install(CODE ..)` block and actually keep `CMAKE_INSTALL_PREFIX` as a variable (escaping the $) to expand it at install time?

jbohren:
Putting the escaped `${CMAKE_INSTALL_PREFIX}` to be evaluated at install time works.

I also modified the develspace marker file generation in the case that an empty `.catkin` file has been created in the develspace. As described in the log message, this should have a negligible impact on race condition sensitivity.

cmake/catkin_generate_environment.cmake
```diff
@@ -76,10 +76,14 @@ function(catkin_generate_environment)
 
   if(NOT CATKIN_BUILD_BINARY_PACKAGE)
     # generate and install workspace marker
-    file(WRITE ${CMAKE_BINARY_DIR}/catkin_generated/installspace/.catkin "")
-    install(FILES
-      ${CMAKE_BINARY_DIR}/catkin_generated/installspace/.catkin
-      DESTINATION ${CMAKE_INSTALL_PREFIX})
+    set(_catkin_marker_file_installed "")
```
dirk-thomas: This variable seems to be not used anymore?

jbohren: Sorry, thought I got that.

jbohren: Fixed in 2f23fd8

cmake/catkin_generate_environment.cmake
```diff
@@ -75,11 +75,12 @@ function(catkin_generate_environment)
...
-    # generate and install workspace marker
-    file(WRITE ${CMAKE_BINARY_DIR}/catkin_generated/installspace/.catkin "")
-    install(FILES
-      ${CMAKE_BINARY_DIR}/catkin_generated/installspace/.catkin
-      DESTINATION ${CMAKE_INSTALL_PREFIX})
+    # install empty workspace marker if it doesn't already exist
+    install(CODE " 
+    if (NOT EXISTS \"\${CMAKE_INSTALL_PREFIX}/.catkin\") 
+      file(WRITE \"\${CMAKE_INSTALL_PREFIX}/.catkin\" \"\") 
+    endif()")
```
dirk-thomas: (This should be really the last one...) Can you please indent lines 80-82 by two space to make it clear that they are not on the same level as the `install()` call?

jbohren: Done.

dirk-thomas:
Thanks for keeping up with all my nitpicking.

jbohren:
> Thanks for keeping up with all my nitpicking.

Thank you for responding so quickly to these PRs!

cmake/catkin_generate_environment.cmake
```diff
@@ -75,11 +75,12 @@ function(catkin_generate_environment)
...
   if(NOT CATKIN_BUILD_BINARY_PACKAGE)
-    # generate and install workspace marker
-    file(WRITE ${CMAKE_BINARY_DIR}/catkin_generated/installspace/.catkin "")
-    install(FILES
-      ${CMAKE_BINARY_DIR}/catkin_generated/installspace/.catkin
-      DESTINATION ${CMAKE_INSTALL_PREFIX})
+    # install empty workspace marker if it doesn't already exist
+    install(CODE " 
+      if (NOT EXISTS \"\${CMAKE_INSTALL_PREFIX}/.catkin\") 
+        file(WRITE \"\${CMAKE_INSTALL_PREFIX}/.catkin\" \"\")
```
dirk-thomas, line 81: This code fails when the folder `CMAKE_INSTALL_PREFIX` does not exist: http://jenkins.ros.org/job/ros-indigo-catkin_binarydeb_trusty_amd64/11/console

I hope the fix in f2efd9e88de383dbfda1cc2015668d5b792c1f15 works...

Actually that won't be sufficient since it also needs to handle `DESTDIR` correctly: 149a2412fc37379dd72f0cf7ba60cb7460a5ab44

PR #678, opened 2014-08-21
refactor code from run_tests.py into Python module

dirk-thomas:


python/catkin/test_results.py
```diff
@@ -31,8 +31,68 @@
...
+
+
+def ensure_junit_result_exist(filename):
+    if os.path.exists(filename):
+        # if result file exists ensure that it contains valid xml
+        tree = None
+        try:
+            tree = ElementTree(None, filename)
+        except ParseError:
+            # print('Invalid XML in result file "%s"' % filename)
```
wjwwood, line 59: Should this be removed?

wjwwood:
+1

tfoote:
+1

python/catkin/test_results.py
dirk-thomas, line 59: Good point. But sometimes I use those commented print statements to trace issues so I will keep it for now.

PR #682, opened 2014-09-16
Fix catkin_make working directory issue

lgerardSRI:
The proper way of removing symlink from a path seems to be os.path.realpath()
Using the PWD environment variable is uncertain, there is no guaranty that it is free of symlinks and moreover nothing ensure that it is properly set, unless you assume some particular shell and way of calling catkin_make. 
One example of this issue is the call of catkin_make as a python subprocess, hitting issues like: see http://bugs.python.org/issue4057

### DIFF
--- a/bin/catkin_make
+++ b/bin/catkin_make
@@ -35,8 +35,8 @@ def main():
     if args.no_color:
         disable_ANSI_colors()
 
-    # use PWD in order to work when being invoked in a symlinked location
-    cwd = os.getenv('PWD', os.curdir)
+    # remove possible symlinks from the working directory
+    cwd = os.path.realpath(os.curdir)
 
     # verify that the base path is known
     base_path = os.path.abspath(os.path.join(cwd, args.directory))


wjwwood:
I think this is actually the whole point, we want the symbolic directory not the real path. Consider this:
- You have your `src` in `~/sources/github/src`
- You create a workspace `~/my_ws` and symbolically link `ln -s ~/sources/github/src ~/my_ws/src`
- You run `catkin_make -C ~/my_ws` in `~/my_ws/src/foo`

The default build space is calculated as relative to your current location (`cwd`) as something like `../../build`, which currently would resolve to `~/my_ws/src/foo`, but in your patch would resolve to `~/sources/github/src/foo`. Therefore you would get `~/sources/github/build` as the build space with your patch. In fact we are counting on the fact that `PWD` gives us the symbolic directory name and not the real one.

Now I didn't work on this part and I am working from memory, so I might have part of this wrong, but I think this was the issue with `PWD` vs `os.path.realpath` vs `os.getcwd`.

This is the original commit: https://github.com/ros/catkin/commit/246674cf34eafa1dfe04492538b776cef76f945e

lgerardSRI:
Ok, I totally misunderstood the goal of this hack. Your example doesn't have any issue, the issue arise if an argument file is given with parent lookup:
let's consider the following:

```
cd /tmp
mkdir real
mkdir other
mkdir src
cd other
ln -s ../real sym
cd sym
```

At this point in the shell, `pwd` will be `/tmp/other/sym`
the user will then call `catkin_make -C ..` which you want to resolve to `/tmp/other` and not `/tmp` ...

wjwwood:
Yeah, what is already there probably isn't a perfect solution, I'll let @dirk-thomas comment, I think he had more intimate knowledge of the problem.

lgerardSRI:
This is a related question http://stackoverflow.com/questions/1542803/is-there-a-version-of-os-getcwd-that-doesnt-dereference-symlinks?rq=1

From what I understand of pwd handling, it is quite fragile but coherent with the interactive user shell experience.

Finally, the fact that you assume pwd is correct if set seems reasonable, I don't see any work around.

PR #684, opened 2014-10-12
add description on CATKIN_BLACKLIST_PACKAGES

k-okada:
because I always forget how to do and google to find http://answers.ros.org/question/54181/how-to-exclude-one-package-from-the-catkin_make-build/


dirk-thomas:
I merged a modified version: 01dc3c37b2cd7a79c53407c86e1be55c8ade4d3d
Thanks!

PR #687, opened 2014-10-20
set terminal title only when isatty

dirk-thomas:
This prevents the ascii characters to be printed when e.g. running on the farm without a terminal.


wjwwood:
+1

Issue #686, opened 2014-10-20
sh emulation in setup.zsh breaks my zsh config


de-vri-es:
The `emulate sh` in setup.zsh breaks something in my zsh config.  I think it's something in my prompt since I get an error message at every prompt. I haven't pinpointed the problem though because I have a rather large configuration, and removing the `emulate sh` fixed the problem for me.

Is there a reason for the `emulate sh`? If there is I can dig and find out what exactly is conflicting, but it seems likely than most sh code will work without modification in zsh, without having zsh emulate sh.

dirk-thomas:
`emulate sh` is currently required since otherwise the called `sh` code will not work correctly in `zsh`. E.g. `zsh` does not honor the `IFS` which is used in _for_ loops.

Since it reset the emulate mode afterwards (a30121998599424367fd19a591e17c904691b903) I don't see how that would conflict with your zsh configuration. It might be worth to track down what exactly conflicts for you since it can't be changed in `catkin` easily.

de-vri-es:
Alright, I will dig to find the exact problem and try to provide a minimal broken zshrc.

de-vri-es:
I found the problem. Either `emulate sh` or `emulate zsh` enables the nomatch option which I specifically disabled in my zshrc. Although I can plate around this specific issue, it is not nice to change options someone set in their zshrc or to require that setup.zsh is sourced before you set your own options.

`emulate -L sh` inside a function is truly temporary and doesn't affect global options. So this would be much nicer:

```
_catkin_source_setup() {
    emulate -L sh
    source "$_CATKIN_SETUP_DIR/setup.sh"
}
_catkin_source_setup
```

dirk-thomas:
It looks like we were depending on `emulate sh` before but as of a0d699a08f10424adad38eee25673c9a66120ec4 we are not using the `IFS` anymore. So removing the emulate lines should work fine.

I created a new PR for it: #692
@de-vri-es Thank you for your effort - you were right in the first place - just removing the emulate lines still works.

de-vri-es:
Awesome, thanks.

Ram-Z:
This breaks it for my setup. I have the zsh option `noclobber` set which doesn't allow truncating a file when redirecting to it.

```
/opt/ros/indigo/setup.sh:52: file exists: /tmp/setup.sh.ksO4Uv1fen
Failed to run '"/opt/ros/indigo/_setup_util.py" ': return code 1
```

PR following.

PR #688, opened 2014-10-20
use emulate -L in setup.zsh to keep options intact

de-vri-es:
Should resolve issue #686. The original overwrites options explicitly set by the user in their zshrc.

### DIFF
--- a/cmake/templates/setup.zsh.in
+++ b/cmake/templates/setup.zsh.in
@@ -3,6 +3,10 @@
 
 CATKIN_SHELL=zsh
 _CATKIN_SETUP_DIR=$(builtin cd -q "`dirname "$0"`" > /dev/null && pwd)
-emulate sh # emulate POSIX
-. "$_CATKIN_SETUP_DIR/setup.sh"
-emulate zsh # back to zsh mode
+
+function _catkin_source_setup {
+  emulate -L sh # emulate POSIX
+  source "$_CATKIN_SETUP_DIR/setup.sh"
+}
+
+_catkin_source_setup


de-vri-es:
I forgot to actually call `_catkin_source_setup` but that's fixed now.

dirk-thomas:
@wjwwood Can you please verify that this works well on Mac OS. E.g. with a small workspace like ros_core and making sure that the environment hooks are sourced correctly.

de-vri-es:
@dirk-thomas Do you know something that would fail without emulation? Because like I said it worked fine for me without emulation as well, I'm not sure what to test. I don't notice any difference with this PR (besides my option not being overwritten any more), but I also didn't notice any difference with/without emulation.

wjwwood:
It originally comes from this: https://github.com/ros/catkin/pull/182

But we have seen zsh not behave like sh (dash) in many cases, especially when dealing with IFS.

de-vri-es:
I don't see the behaviour described in #182, with or without emulation (any method of enabling/disabling it). My guess is it got fixed in a different manner.

By the way, it seems that `emulate sh` does not propagate to functions defined while the emulation mode was active. Using `emulate -L` or the original method does not matter. I believe the proper way to do something like that in zsh is this:

```
emulate sh -c '_catkin_source_setup() {
  source "$_CATKIN_SETUP_DIR/setup.sh"
}
_caktin_source_setup
```

This enables sticky emulation which is inherited by all functions defined by that function. However, doing that I do notice a difference, namely that autocompletion doesn't work any more at all (at least for rosservice).

For now I still believe this PR is an improvement since it seems to preserve behaviour (as far as I can tell, but please double-check) but it doesn't overwrite options set by the user any more.

dirk-thomas:
Closing in favor of #692. Please see #686 for more information.

PR #689, opened 2014-10-20
use local scope for zsh emulating sh

dirk-thomas:
Addressing #686

@wjwwood Can you please verify that this works well on Mac OS. E.g. with a small workspace like `ros_core` and making sure that the environment hooks are sourced correctly.


dirk-thomas:
Too slow - I just saw #688.

PR #692, opened 2014-10-21
remove 'emulate sh' from setup.zsh

dirk-thomas:
Addresses #686. I checked it with zsh on Ubuntu.

@wjwwood Please verify that it works on OS X too.


wjwwood:
+1

PR #693, opened 2014-10-22
add --use-ninja option to catkin_make(_isolated) to use ninja instead of make

dirk-thomas:

### DIFF
--- a/bin/catkin_make
+++ b/bin/catkin_make
@@ -133,11 +133,22 @@ def main():
             print(fmt('@{yf}Packages @{boldon}"%s"@{boldoff} not found in the workspace - ignoring them' % ', '.join(sorted(unknown_packages))), file=sys.stderr)
             args.pkg = [name for name in args.pkg if name in packages_by_name]
 
+    if not [arg for arg in cmake_args if arg.startswith('-G')]:
+        if not args.use_ninja:
+            cmake_args += ['-G', 'Unix Makefiles']
+        else:
+            cmake_args += ['-G', 'Ninja']
+    elif args.use_ninja:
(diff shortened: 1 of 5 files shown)


tfoote:
+1 there are two or three repeated if clauses for the makefile name and command line args, but at two instances the function method will not save lines, but if there's ever to be more, it might make sense to make them a function.

bin/catkin_make
```diff
@@ -133,11 +133,20 @@ def main():
             print(fmt('@{yf}Packages @{boldon}"%s"@{boldoff} not found in the workspace - ignoring them' % ', '.join(sorted(unknown_packages))), file=sys.stderr)
             args.pkg = [name for name in args.pkg if name in packages_by_name]
 
+    if '-G' not in cmake_args:
```
wjwwood: Is it valid for an input to be `['-GFoo']`, which would cause this to fail? It seems to be possible:

```
% cmake -GXcode ..
-- The C compiler identification is AppleClang 6.0.0.6000054
-- The CXX compiler identification is AppleClang 6.0.0.6000054
-- Check for working C compiler using: Xcode
-- Check for working C compiler using: Xcode -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working CXX compiler using: Xcode
^C
```

Would it not be an error if you specify both a `-G` argument in the `cmake_args` and `--use-ninja`?

dirk-thomas: Both good points. Addressed in second commit.

wjwwood:
+1

dirk-thomas:
Squashed.

PR #700, opened 2014-12-18
check changes to -D args when considering to reinvoke cmake

dirk-thomas:
- `CATKIN_DEVEL_PREFIX`
- `CMAKE_INSTALL_PREFIX`

Came up here: http://answers.ros.org/question/199740/catkin-dcmake_install_prefix-flag/


wjwwood:
+1

PR #704, opened 2014-12-31
Add argument to find_in_workspaces in order to cache result of catkin_pkg.packages.find_packages

garaemon:
patch for https://github.com/ros/ros_comm/issues/541


dirk-thomas:
Adding this new argument as well as an internal cache seems to be not necessary. The function already has an optional argument `source_path_to_packages` to pass a dictionary containing these cached information.

dirk-thomas:
I will close this PR since as mentioned before the existing argument `source_path_to_packages` already provides the necessary API to implement the requested functionality externally.

garaemon:
Sure, I will update PR!

PR #705, opened 2015-01-05
add verbose option to catkin_test_results

k-okada:


dirk-thomas:
The new test fails with Python 3.

k-okada:
ok, fixed for python3

bin/catkin_test_results
```diff
@@ -18,6 +18,7 @@ def main():
     parser = argparse.ArgumentParser(description='Outputs a summary of the test results. If there are any test errors or failures the scripts return code is 1.')
     parser.add_argument('test_results_dir', nargs='?' if test_results_dir else None, default=test_results_dir, help='The path to the test results (necessary when the environment variable "%s" is not defined)' % env_name)
     parser.add_argument('--all', action='store_true', default=False, help='Show all test results even the ones without errors/failures')
+    parser.add_argument('--detail', action='store_true', default=False, help='Show all errored/failureed tets')
```
dirk-thomas: Please name the command line option `--verbose` instead.

Also correct the spelling errors in the help text.

python/catkin/test_results.py
```diff
@@ -134,6 +134,15 @@ def test_results(test_results_dir):
                 print('Skipping "%s": %s' % (name, str(e)))
                 continue
             results[name] = (num_tests, num_errors, num_failures)
+            if show_detail and (num_errors + num_failures > 0) :
+                print("Test %s was failed" % (name))
+                print("-------------------------------------------------")
+                detail=open(filename_abs,"rb").read()
+                if sys.version_info[0] >= 3:
+                    detail=detail.decode('utf-8')
+                detail=detail.replace('\n',"\n").replace('&gt;',">").replace('&lt;',"<")
```
dirk-thomas: What is `replace('\n',"\n")` supposed to do?

Regarding the other replaces: should all XML entities be unescaped?

python/catkin/test_results.py
```diff
@@ -134,6 +134,15 @@ def test_results(test_results_dir):
...
+                print(detail)
```
dirk-thomas: I don't think that it is a good idea to print the whole result file. E.g. a result file can contain many tests (and even test suites) and if only one of them failed / contains a failed test print the whole file seems to be overly verbose.

python/catkin/test_results.py
```diff
@@ -134,6 +134,15 @@ def test_results(test_results_dir):
...
+                print("-------------------------------------------------")
```
dirk-thomas: This code block does not follow PEP8.

test/unit_tests/test_test_results.py
```diff
@@ -44,6 +44,26 @@ def test_test_results(self):
         finally:
             shutil.rmtree(rootdir)
 
+    def test_test_results_detail(self):
+        results = {'test1.xml': (5, 1, 3), 'test2.xml': (7, 2, 4)}
```
dirk-thomas: This line looks unnecessary.

test/unit_tests/test_test_results.py
```diff
@@ -44,6 +44,26 @@ def test_test_results(self):
...
+        try:
+            oldstdout = sys.stdout
+            sys.stdout = StringIO()
+            rootdir = tempfile.mkdtemp()
+            for filename in ['test1.xml', 'test2.xml', 'foo.bar']:
```
dirk-thomas: Why should this test use more then one file? It is about the detailed output and not about correct discovery or parsing of result files.

test/unit_tests/test_test_results.py
```diff
@@ -44,6 +44,26 @@ def test_test_results(self):
...
+                result_file = os.path.join(rootdir, filename)
+                with open(result_file, 'w') as fhand:
+                    fhand.write('<testsuites tests="5" failures="3" errors="1" time="35" name="AllTests"></testsuites>')
+            results = catkin_test_results.test_results(rootdir, True)
```
dirk-thomas: The second argument should use the keyword.

test/unit_tests/test_test_results.py
```diff
@@ -44,6 +44,26 @@ def test_test_results(self):
...
+            self.assertEqual({'test1.xml': (5, 1, 3), 'test2.xml': (5, 1, 3)}, results)
+            summary = sys.stdout.getvalue()
+            self.assertTrue('test1.xml' in summary, summary)
+            self.assertTrue('test2.xml' in summary, summary)
+            self.assertTrue('<testsuites tests="5" failures="3" errors="1" time="35" name="AllTests"></testsuites>' in summary, summary)
```
dirk-thomas: Instead of having the same string again spelled out it should come from a variable.

python/catkin/test_results.py
```diff
@@ -134,6 +134,15 @@ def test_results(test_results_dir):
                 print('Skipping "%s": %s' % (name, str(e)))
                 continue
             results[name] = (num_tests, num_errors, num_failures)
+            if show_detail and (num_errors + num_failures > 0) :
+                print("Test %s was failed" % (name))
+                print("-------------------------------------------------")
+                detail=open(filename_abs,"rb").read()
```
dirk-thomas: Why read the file in binary mode if you decode it later anyway?

bin/catkin_test_results
k-okada: fixed in 00f10c5

python/catkin/test_results.py
k-okada: it does not have to be binary, fixed in e8818d7

python/catkin/test_results.py
k-okada: because "verbose file" (xml output from the rostest) is escaped, so we want to unescape the file.
Do you think it is ok to import some of xml library and unescape the text?

python/catkin/test_results.py
k-okada: fixed in 5749122

python/catkin/test_results.py
k-okada: If it's ok to use xml library, it may easyer

test/unit_tests/test_test_results.py
k-okada: fixed in 7c85298

test/unit_tests/test_test_results.py
k-okada: it does not have to be several files, fixed 64611c7

test/unit_tests/test_test_results.py
k-okada: fixed in 03036ba

test/unit_tests/test_test_results.py
k-okada: fixed 0504fa2, 

Thanks for review and comments

python/catkin/test_results.py
dirk-thomas: I assume you want to replace "\\n" with "\n"? And not the same string with itself as it is currently?

bin/catkin_test_results
```diff
@@ -18,6 +18,7 @@ def main():
     parser = argparse.ArgumentParser(description='Outputs a summary of the test results. If there are any test errors or failures the scripts return code is 1.')
     parser.add_argument('test_results_dir', nargs='?' if test_results_dir else None, default=test_results_dir, help='The path to the test results (necessary when the environment variable "%s" is not defined)' % env_name)
     parser.add_argument('--all', action='store_true', default=False, help='Show all test results even the ones without errors/failures')
+    parser.add_argument('--verbose', action='store_true', default=False, help='Show all errored/failureed tests')
```
dirk-thomas: Still spelling errors / invalid words in the help text: `errored/failureed`.

Suggestion:

```
Show all tests which have errors or failed
```

python/catkin/test_results.py
```diff
@@ -113,12 +113,13 @@ def read_junit(filename):
...
 
 
-def test_results(test_results_dir):
+def test_results(test_results_dir, show_verbose=False):
     '''
     Collects test results by parsing all xml files in given path,
     attempting to interpret them as junit results.
 
     :param test_results_dir: str foldername
+    :param show_verbose: bool show test results
```
dirk-thomas: Suggestion:

```
show output for tests which had errors or failed
```

python/catkin/test_results.py
```diff
@@ -134,6 +135,13 @@ def test_results(test_results_dir):
                 print('Skipping "%s": %s' % (name, str(e)))
                 continue
             results[name] = (num_tests, num_errors, num_failures)
+            if show_verbose and (num_errors + num_failures > 0):
+                print("Test %s was failed" % (name))
+                print("-------------------------------------------------")
+                verbose = open(filename_abs, "r").read()
+                verbose = verbose.replace('\n', "\n").replace('&gt;', ">").replace('&lt;', "<")
```
dirk-thomas: The same comment still applies here: https://github.com/ros/catkin/pull/705#discussion_r22666329

k-okada: Thanks I found this is no longer required.

k-okada:
@dirk-thomas thanks for comment, I fixed again.

python/catkin/test_results.py
```diff
@@ -134,6 +135,12 @@ def test_results(test_results_dir):
                 print('Skipping "%s": %s' % (name, str(e)))
                 continue
             results[name] = (num_tests, num_errors, num_failures)
+            if show_verbose and (num_errors + num_failures > 0):
+                print("Test %s was failed" % (name))
```
dirk-thomas: "Test %s was failed" does not sound good. Also the test name should be surrounded by single quotes. Suggestion: "Full test results for test '%s'".

python/catkin/test_results.py
```diff
@@ -134,6 +135,12 @@ def test_results(test_results_dir):
...
+                print("-------------------------------------------------")
+                verbose = open(filename_abs, "r").read()
+                print(verbose)
+                print("-------------------------------------------------")
```
dirk-thomas: Please use single quotes instead of double quotes.

python/catkin/test_results.py
```diff
@@ -134,6 +135,12 @@ def test_results(test_results_dir):
                 print('Skipping "%s": %s' % (name, str(e)))
                 continue
             results[name] = (num_tests, num_errors, num_failures)
+            if show_verbose and (num_errors + num_failures > 0):
+                print("Test %s was failed" % (name))
+                print("-------------------------------------------------")
+                verbose = open(filename_abs, "r").read()
```
dirk-thomas: Have you tried this with a test result file which contains non-ascii characters? May be adding a test for this would be good.

python/catkin/test_results.py
k-okada: thanks, fixed in https://github.com/k-okada/catkin/commit/cce91c1ee2f53ca8b40d3038781234d1d3e3b9f7

python/catkin/test_results.py
k-okada: is https://github.com/k-okada/catkin/commit/c8bbaa9f84c01c2dca14f4d4a40598a2cf12e4d2 what you suggested?

python/catkin/test_results.py
```diff
@@ -134,6 +135,12 @@ def test_results(test_results_dir):
                 print('Skipping "%s": %s' % (name, str(e)))
                 continue
             results[name] = (num_tests, num_errors, num_failures)
+            if show_verbose and (num_errors + num_failures > 0):
+                print("Full test results for '%s'" % (name))
+                print('-------------------------------------------------')
+                verbose = open(filename_abs, 'r').read()
```
dirk-thomas: This keeps the file handle open longer then it needs. Please read the content inside a `with` block instead.

test/unit_tests/test_test_results.py
```diff
@@ -44,6 +45,45 @@ def test_test_results(self):
...
+                fhand.write(test_suites)
+            results = catkin_test_results.test_results(rootdir, show_verbose=True)
+            self.assertEqual({test_xml: (5, 1, 3)}, results)
+            summary = sys.stdout.getvalue()
+            self.assertTrue(test_xml in summary, summary)
+            self.assertTrue(test_suites in summary, summary)
+        finally:
+            shutil.rmtree(rootdir)
+            sys.stdout = oldstdout
+            print summary
```
dirk-thomas: This is invalid in Python 3. Please also ensure that the tests pass with Python 3.

k-okada:
ok, fixed

CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ install(DIRECTORY cmake
   PATTERN "catkinConfig.cmake.in" EXCLUDE)
 
 catkin_python_setup()
+execute_process(COMMAND sleep 10)
```
dirk-thomas: ???

k-okada: Oh, sorry , that's a debug code. removed .

## 

◉ Kei Okada

dirk-thomas:
Thanks, can you please squash your commits.

k-okada:
ok, squash into one commit

dirk-thomas:
LGTM

@wjwwood Can you please take a short look at this?

wjwwood:
+1, so this just prints the XML files one after another? I don't understand the use case, but the change looks ok.

k-okada:
we need this when we want to know why the test fails.

## 

◉ Kei Okada

Issue #706, opened 2015-01-12
How to disable --only-pkg-with-deps?


mkoval:
I used `--only-pkg-with-deps` to rebuild a single package in a Catkin workspace. After doing so, future invocations of `catkin_make` (with no options) do not force a configure. As a result, they build the subset of packages that were selected by the flag.

The only solution I've found to work is to delete the entire build directory. Surprisingly, the `--force-cmake` flag also doesn't force a reconfigure (this seems like a bug?).

This definitely happens in version 0.6.11 on Ubuntu 12.04 and 14.04.

dirk-thomas:
`--only-pkg-with-deps` equates to set the CMake variable `CATKIN_WHITELIST_PACKAGES`. Once that is set it will be in the `CMakeCache.txt` file until being redefined.

Passing `--force-make` _does_ invoke the CMake configuration step again. But it will keep using the variable from the cache.

The script has no knowledge if the user had passed `-CATKIN_WHITELIST_PACKAGES=...` before (and it should keep using it) or if the user passed `--only-pkg-with-deps` before and now intends it to be "reset". Therefore I don't see a good approach to change the current behavior.

wjwwood:
I can confirm the behavior you describe is happening, but it is not because CMake is not running. I did this:

```
$ mkdir -p /tmp/test_catkin_ws/src

$ cd /tmp/test_catkin_ws/src

$ git clone https://github.com/ros/catkin.git
...

$ cat catkin/package.xml | grep version
<?xml version="1.0"?>
  <version>0.6.11</version>
  <depend version_gte="0.2.2">python-catkin-pkg</depend>

$ catkin_create_pkg --rosdistro indigo foo
...

$ catkin_create_pkg --rosdistro indigo bar foo
...

$ cd ..

$ ./src/catkin/bin/catkin_make
Base path: /tmp/test_catkin_ws
Source space: /tmp/test_catkin_ws/src
Build space: /tmp/test_catkin_ws/build
Devel space: /tmp/test_catkin_ws/devel
Install space: /tmp/test_catkin_ws/install
Creating symlink "/tmp/test_catkin_ws/src/CMakeLists.txt" pointing to "catkin/cmake/toplevel.cmake"
####
[...]

PR #707, opened 2015-01-13
Improved help text for --only-pkg-with-deps

mkoval:
I changed the description of the `--only-pkg-with-deps` option to address #706. I left the description for the option in `catkin_make_isolated` unchanged. From what @wjwwood and @dirk-thomas described, the option should perform as expected on that command (since each package is built in isolation).


dirk-thomas:
I think the improved help text should mention the exact CMake variable and may even suggest the user how to reset it (beside mentioning that the option is persistent).

Can you please update the patch with this information?

mkoval:
@dirk-thomas Done in b2aa505a067d0bd8a07f6a20d630248cc7dcfffc.

wjwwood:
+1

dirk-thomas:
Thanks, lgtm. Can you please squash the commits before the merge?

mkoval:
@dirk-thomas Done.

PR #711, opened 2015-01-22
source in emulation mode  (really fix #686)

### LINKED ISSUE #686: sh emulation in setup.zsh breaks my zsh config

Ram-Z:
#686 (links do not get made in the subject line)

### DIFF
--- a/cmake/templates/setup.zsh.in
+++ b/cmake/templates/setup.zsh.in
@@ -5,4 +5,4 @@ CATKIN_SHELL=zsh
 
 # source setup.sh from same directory as this file
 _CATKIN_SETUP_DIR=$(builtin cd -q "`dirname "$0"`" > /dev/null && pwd)
-source "$_CATKIN_SETUP_DIR/setup.sh"
+emulate -R zsh -c 'source "$_CATKIN_SETUP_DIR/setup.sh"'


dirk-thomas:
The use of `emulate` was removed recently for a good reason. Please see the related tickets to why it was necessary.

The issues fixed with the removal would again be a problem if you patch would be applied. Without any further information I don't think we can just readd it. Please provide much more context and also why you think your patch does not result in the previous issues to reappear.

Ram-Z:
Please read the patch. This would emulate only the source command and not touch the current environment or options.

The removal of emulate breaks the setup for me, see my comment on #686.

dirk-thomas:
I have read your patch. It reverts a recent change (namely removing the usage of emulate) which was made for a reason.

Using emulate when sourcing the setup.sh file implies that also all zsh environment hooks are sourced in emulate mode. This is not desired since zsh environment hooks should be sourced in a "normal" zsh environment.

May be there is a different way to work around the issue you are having.

Ram-Z:
I'll be a bit more precise. The previous line read:

```
emulate sh
source [..]
emulate zsh
```

This results in resetting all options in the parent shell. Calling `emulate sh -c 'source [..]'` only emulates sh for the command specified after `-c` without touching the parent shell environment.

This PR is a better way to fix #686 and should not reset any options for the calling shell. It is the same way that zsh will source `/etc/profile` in sh mode by default (at least on Arch).

And here is a quote from `man zsh`:

```
emulate [ -LR ] [ {zsh|sh|ksh|csh} [ flags ... ] ]
[..]
 If -c arg appears in flags, arg is evaluated while the requested
       emulation  is temporarily in effect.  In this case the emulation
       mode and all options  are  restored  to  their  previous  values
       before  emulate  returns.
```

Ram-Z:
> This is not desired since zsh environment hooks should be sourced in a "normal" zsh environment.

The linked issue doesn't mention that. Why is this necessary?  `setup.sh` doesn't define any functions so this should be irrelevant.

Now that I think of it, sourcing it in a reset zsh mode will probably also work, i.e.:

```
emulate -R zsh 'source [..]'
```

edit: tested and it works

dirk-thomas:
It is not "necessary" but it is what a developer expects when he implements a zsh environment hook.

`emulate -R zsh ...` sounds like an interesting option. Can you update the PR with this? I guess we would need  several people to try this change before it can be merged.

Ram-Z:
> It is not "necessary" but it is what a developer expects when he implements a zsh environment hook.

But as `setup.sh` doesn't define any functions this should not matter, and if there were any, they should probably be run in emulation mode to avoid errors. Anyway no need to discuss this further, the  `emulate -R zsh` is pushed and that should do the trick.

dirk-thomas:
`setup.sh` sources environment hooks and those can perform arbitrary actions. Especially the hooks providing auto completion features do use functions and other stuff.

dirk-thomas:
@de-vri-es @wjwwood Can you please test this change (which adds back `emulate` but in a different manner) in your zsh environments and confirm that it works (especially check that all environment hooks are run / all environment variables are set after sourcing)?

wjwwood:
@Ram-Z the exact use case for sourcing the `setup.sh` with `zsh` and not `sh` emulation is so that zsh autocompletion will work. I'm concerned that this will nullify changes to the environment (like the autocompletion functions defined by the environment hooks), but maybe I misunderstand the `-R` option.

I'll give it a try.

de-vri-es:
The problems I experienced were indeed because options from my own zshrc were reset and broke some of my zsh configuration. This PR doesn't reintroduce that problem. I don't know what effect it would have on other hooks or auto-completion.

wjwwood:
Sourcing works and tab completion works in my test setup. Seems ok to me.

Ram-Z:
What tab completion? For `catkin_make`? I can't make that work even without any emulate. Only directly sourcing `etc/catkin/profile.d/05.catkin_make.bash` after setting bashcompinit makes those completion work. (it never worked for me, only just now discovered that they existed)

Any other completion is unaffected by any emulate (zsh,sh,-R) or lack thereof, in my tests.

wjwwood:
@Ram-Z I was testing `rosrun` from `rosbash`:

https://github.com/ros/ros/blob/indigo-devel/tools/rosbash/env-hooks/15.rosbash.zsh.em

dirk-thomas:
@Ram-Z `catkin` does not provide any auto completion for zsh. Simply because nobody implemented it yet. If your are interested please consider to provide a PR adding zsh support.

Since it looks like this patch doesn't break anything I will merge it.

PR #712, opened 2015-01-22
documentation: updates for rostest

czalidis:
Modify documentation to reflect the new API for rostest, see ros/ros_comm#555.

### DIFF
--- a/doc/howto/format1/rostest_configuration.rst
+++ b/doc/howto/format1/rostest_configuration.rst
@@ -38,6 +38,29 @@ Finally, declare your rostest launch scripts::
     add_rostest(tests/your_second_rostest.test)
   endif()
 
+In case you have a test that accepts arguments, you can pass them like
+this::
+
+  add_rostest(tests/your_rostest.test ARGS arg1:=true arg2:=false)
+
+If your rostest needs extra data in order to run, you can use the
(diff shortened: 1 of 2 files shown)


wjwwood:
I guess this patch would also apply to the "format1" version of this:

https://github.com/czalidis/catkin/blob/documentation_rostest/doc/howto/format1/rostest_configuration.rst

It's a little tedious to keep them in sync, but we should try to. Can you please update the "format1" version with your changes as well?

wjwwood:
Otherwise it looks good, thanks @czalidis. +1

czalidis:
Ok, I'll do. I completely forgot about "format1"...

dirk-thomas:
Can you please add a newlines before the start of a new sentence (in the paragraphs you added)?

czalidis:
You mean that each sentence should start in a new line? (because thats not the case for the rest of the document)

dirk-thomas:
Yes, this is currently our recommended style since it makes diffs easier to read.
Otherwise changes can propagate through the whole paragraph due to re-wrapping everything.

Don't worry about the existing paragraph - they need to be cleaned up at some point.

dirk-thomas:
Thanks!

PR #713, opened 2015-02-02
Fix setting CATKIN_GLOBAL_LIBEXEC_DESTINATION to libexec.

aballier:
I tried to set CATKIN_GLOBAL_LIBEXEC_DESTINATION to libexec, as described as possible in doc/user_guide/variables.rst, since I find this location more natural for nodes & co.
However, catkin_find could not find any node after that. This patch makes it possible to find nodes there, while maintaining the translation libexec -> lib.

### DIFF
--- a/doc/user_guide/variables.rst
+++ b/doc/user_guide/variables.rst
@@ -71,6 +71,9 @@ They only contain relative paths and are supposed to be relative to the ``${CMAK
 
    This is set to ``lib``.
    On non-Debian distributions it could be set to ``libexec``.
+   Note that setting this variable to anything but ``lib`` or ``libexec`` will
+   not work as expected since tools like ``catkin_find`` will only consider
+   these two locations.
    This variable should not be used directly, use ``CATKIN_PACKAGE_BIN_DESTINATION`` instead.
 
 .. cmake:data:: CATKIN_GLOBAL_PYTHON_DESTINATION
--- a/python/catkin/find_in_workspaces.py
+++ b/python/catkin/find_in_workspaces.py
@@ -118,9 +118,11 @@ def find_in_workspaces(search_dirs=None, project=None, path=None, _workspaces=ge
     existing_paths = []
     try:
         for workspace in (_workspaces or []):
+            if 'libexec' in search_dirs:
+                search_dirs.insert(search_dirs.index('libexec'), 'lib')
             for sub in search_dirs:
                 # search in workspace
-                p = os.path.join(workspace, sub if sub != 'libexec' else 'lib')
+                p = os.path.join(workspace, sub)
                 if project:
                     p = os.path.join(p, project)
                 if path:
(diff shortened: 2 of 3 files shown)


dirk-thomas:
The order of `search_dirs` determines where resources are looked up first. Appending `search_dirs_extra` at the end would change the order and can therefore not be done as it is in this patch.

Also changing the variable `CATKIN_GLOBAL_LIBEXEC_DESTINATION` to anything else would not work since the Python logic does not know about that.

I would recommend to not override the libexec destination since it is currently not well supported. May be a note should be added to the documentation to make this clear.

aballier:
`The order of search_dirs determines where resources are looked up first. Appending search_dirs_extra at the end would change the order and can therefore not be done as it is in this patch.`

Right, didn't think about that.
search_dirs.insert(search_dirs.index('libexec'), 'lib') should do it then (and be simpler)

`Also changing the variable CATKIN_GLOBAL_LIBEXEC_DESTINATION to anything else would not work since the Python logic does not know about that.`

Could you please be more specific ? I've been using this for months without a single issue.

dirk-thomas:
If you would set `CATKIN_GLOBAL_LIBEXEC_DESTINATION` to e.g. `foo` the script `catkin_find` would still not work since it doesn't know about that path.

aballier:
Ah, yes indeed. Thanks.
Will update the PR with the above fix + a doc patch.

aballier:
Updated

doc/user_guide/variables.rst
```diff
@@ -71,6 +71,8 @@ They only contain relative paths and are supposed to be relative to the ``${CMAK
 
    This is set to ``lib``.
    On non-Debian distributions it could be set to ``libexec``.
+   Note that setting this variable to anything but ``lib`` or ``libexec`` will
+   not work as expected since the logic in the code expects one of these values.
```
dirk-thomas: Instead of:

```
since the logic in the code expects one of these values.
```

I would write it like:

```
since tools like ``catkin_find`` will only consider these two locations.
```

aballier: agreed, updating it

dirk-thomas:
Thank you. Squashed and merged in a4b3a23576dba8bf1cece8432044f0c334860ac8.

dirk-thomas:
ros/ros#83 indicates a regression due to this PR. Extending `search_dirs` should not be done in the workspaces loop but outside.

The duplicated return in the updated test should have been a strong indicator of that: https://github.com/ros/catkin/blob/37576a5d5afe2599bcfec21831ea757ac3b77430/test/unit_tests/test_find_in_workspace.py#L63-L64

I have created #739 to address it.

aballier:
oh, indeed, sorry & thanks for fixing it

i assumed the duplicated return was due to that search_dirs already contained 'lib', guess i was wrong

PR #714, opened 2015-02-02
output whitelist/blacklist if set

dirk-thomas:
@wjwwood Please review.


wjwwood:
Does this cover the `if(NOT DEFIEND ...)` case too, i.e. what happens if `CATKIN_WHITELIST_PACKAGES` is not defined? Or is it always defined.

dirk-thomas:
If it is not defined the variable evaluates to an empty string and therefore doesn't output a message.

But with it being defined as an `option` it is always defined (defaults to an empty string).

wjwwood:
+1

PR #716, opened 2015-02-06
New implementation of #633

### LINKED ISSUE #690: catkin_download_test_data doesn't update files when MD5 sum changes
If the MD5 sum on a file changes or the file is corrupted locally, catkin_download_test_data does not run the download_checkmd5.py script to check it and download a clean copy.

To reproduce (in a workspace with ros_comm):

```
catkin_make tests
md5sum ./build/ros_comm/test/test_rosbag/test/test_rosbag_latched_pub.bag
echo 1 > ./build/ros_comm/test/test_rosbag/test/test_rosbag_latched_pub.bag
catkin_make tests
md5sum ./build/ros_comm/test/test_rosbag/test/test_rosbag_latched_pub.bag
```

bricerebsamen:
This is a new implementation of #633. I did it on top of recent changes in indigo-devel, with the changes from the catkin_download branch. I also modified @trainman419 's implementation a bit to take into account @dirk-thomas requests.

Hopefully this will be finally merged and will fix #690.

See discussion for #633 and #690 for info

### DIFF
--- a/cmake/catkin_download.cmake
+++ b/cmake/catkin_download.cmake
@@ -1,0 +1,40 @@
+#
+# Download a file from a URL.
+#
+# .. note:: It is not recommended to rely on downloaded data during
+#   a configure / make cycle since this prevents building the package
+#   when no network connectivity is available.
+#
+# :param target: the target name
+# :type target: string
+# :param url: the url to download
+# :type url: string
+
+# :param DESTINATION: the directory where the file is downloaded to
+#   (default: ${PROJECT_BINARY_DIR})
+# :type DESTINATION: string
+# :param FILENAME: the filename of the downloaded file
+#   (default: the basename of the url)
+# :type FILENAME: string
+# :param MD5: the expected md5 hash to compare against
+#   (default: empty, skipping the check)
+# :type MD5: string
+#
+# @public
+function(catkin_download target url)
+  cmake_parse_arguments(ARG "" "DESTINATION;FILENAME;MD5" "" ${ARGN})
+  if(ARG_UNPARSED_ARGUMENTS)
+    message(FATAL_ERROR "catkin_download() called with unused arguments: ${ARG_UNPARSED_ARGUMENTS}")
+  endif()
+  if(NOT ARG_DESTINATION)
+    set(ARG_DESTINATION ${PROJECT_BINARY_DIR})
+  endif()
+  if(NOT ARG_FILENAME)
+    get_filename_component(ARG_FILENAME ${url} NAME)
+  endif()
+  set(output "${ARG_DESTINATION}/${ARG_FILENAME}")
+  add_custom_command(OUTPUT ${output}
+    COMMAND ${PYTHON_EXECUTABLE} ${catkin_EXTRAS_DIR}/test/download_checkmd5.py ${url} ${output} ${ARG_MD5}
+    VERBATIM)
+  add_custom_target(${target} DEPENDS ${output})
+endfunction()
--- a/cmake/catkin_download.cmake
+++ b/cmake/catkin_download.cmake
@@ -17,7 +17,7 @@
 # :type url: string
 #
 # :param DESTINATION: the directory where the file is downloaded to
-#   (default: ${CATKIN_DEVEL_PREFIX}/${CATKIN_PACKAGE_SHARE_DESTINATION})
+#   (default: ${PROJECT_BINARY_DIR})
 # :type DESTINATION: string
 # :param FILENAME: the filename of the downloaded file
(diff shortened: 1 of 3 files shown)


cmake/catkin_download.cmake
```diff
@@ -0,0 +1,146 @@
...
+# .. note:: The target will be registered as a dependency
+#   of the "download_extra_data" target.
+#
+# :param target: the target name
+# :type target: string
+# :param url: the url to download
+# :type url: string
+#
+# :param DESTINATION: the directory where the file is downloaded to
+#   (default: ${CATKIN_DEVEL_PREFIX}/${CATKIN_PACKAGE_SHARE_DESTINATION})
```
dirk-thomas: The default value in the code seems to be different: https://github.com/ros/catkin/pull/716/files#diff-f6d46f8e02eefd6f88ad660967bc8182R44

cmake/catkin_download.cmake
```diff
@@ -0,0 +1,146 @@
...
+# :type DESTINATION: string
+# :param FILENAME: the filename of the downloaded file
+#   (default: the basename of the url)
+# :type FILENAME: string
+# :param MD5: the expected md5 hash to compare against
+#   (default: empty, skipping the check)
+# :type MD5: string
+# :param EXCLUDE_FROM_ALL: exclude this download from the 'all' build target
+#   (default: true, excluding it from the all target)
+# :type EXCLUDE_FROM_ALL: bool
```
dirk-thomas: In CMake `EXCLUDE_FROM_ALL` is usually used as an option. It might be confusing to use it as a boolean argument.

cmake/catkin_download.cmake
```diff
@@ -0,0 +1,146 @@
...
+# :param REQUIRED: require the file to be available for the build to succeed
+#   (default: true, requiring the file)
+# :type REQUIRED: bool
```
dirk-thomas: Same as `EXCLUDE_FROM_ALL`. CMake uses `REQUIRED` usually as an option.

cmake/catkin_download.cmake
```diff
@@ -0,0 +1,146 @@
...
+#   (default: empty, skipping the check)
+# :type MD5: string
+# :param REQUIRED: require the file to be available for the build to succeed
+#   (default: true, requiring the file)
+# :type REQUIRED: bool
+#
+# @public
+function(catkin_download_extra_data)
+  catkin_download(${ARGN} EXCLUDE_FROM_ALL true)
+endfunction()
```
dirk-thomas: Do we really need these two functions (`catkin_download_build_data` and `catkin_download_extra_data`) since they only pass a single argument to the download function? I would rather avoid adding those in favor of documenting the uses when how to invoke `catkin_download` for different use cases.

cmake/catkin_download.cmake
```diff
@@ -0,0 +1,146 @@
...
+
+if(NOT TARGET download_extra_data)
+  add_custom_target(download_extra_data)
```
dirk-thomas, line 72: What is this target being used for?

cmake/catkin_download.cmake
bricerebsamen: can someone provide some guidance or examples on how to do this properly. I spent the last 30 minutes searching but found nothing.

cmake/catkin_download.cmake
bricerebsamen, line 72: it's referenced at line 70. All the downloads will be added to this target. So that if you want to download the data you can call this target. This is useful if the download targets were added with EXCLUDE_FROM_ALL. This way you can build all first, and the call download_extra_data instead of calling all the download targets one by one.

bricerebsamen:
is there something stopping this PR from being merged? anything I can do?

wjwwood:
We probably didn't notice it had been updated. We only get emails when you comment, not when you push a commit. If that's not the case, then we just haven't had time to give it a second review since the most recent changes.

I'll give it a review now.

wjwwood:
I reviewed the changes in ec239b6, but not the whole request, and they look reasonable and in line with @dirk-thomas's comments.

dirk-thomas:
Code review looks good. I will try to run and merge it sometime next week.

trainman419:
Bump. Has anyone had a chance to test and validate this?

dirk-thomas:
Not yet, it is scheduled for the "ROS bug / ticket squashing day" next Thursday.

dirk-thomas:
Thank you for the PR!

Squashed and merged with only minor changes: 09ce538e47e179dbd433d9c6a2f04211c609a44b

PR #720, opened 2015-03-06
display elapsed time during running catkin

k-okada:
I'm not sure how to pass argument to enable/disable dis feature, should we add new function aruguments and fix catkin_make commands?

### DIFF
--- a/python/catkin/builder.py
+++ b/python/catkin/builder.py
@@ -38,6 +38,7 @@ import os
 import platform
 import re
 import stat
+from datetime import datetime, timedelta
 try:
     from StringIO import StringIO
 except ImportError:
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Yes, this feature should be optional and disabled by default. You can add command line options for both `catkin_make` and `catkin_make_isolated` for that.

Please also avoid having a global start time in the builder module.

dirk-thomas:
Please consider to update the PR based on the above comment in order to be merged.

dirk-thomas:
It might be more flexible to use an existing tool for that feature. Basically the functionality would apply to any tool and I would rather avoid to add it everywhere.

Please take a look at `ts -r` from http://joeyh.name/code/moreutils/

I will close this for now. If you think the external tool is not sufficient for whatever reason please comment.

PR #721, opened 2015-03-08
improve generality of generate_cmake_rst.py

pwm1234:
- use an optional json config file to exclude listed from the auto-generated api file
- extend generate_rst with optional arguments that allow skipping of private and undocumented objects
- use '~' rather than '-' so the individual macro/function objects are one level in the outline below the section headers for Public, Non-public, and Not documented.
- use :local: option to the contents directive to avoid pulling in the title page header

I built and deployed the catkin docs to my fork's gh-pages branch. The page http://pwm1234.github.io/catkin/doc/build/html/dev_guide/generated_cmake_api.html shows the changes/improvements to the extracted api page.

### DIFF
--- a/doc/generate_cmake_rst.py
+++ b/doc/generate_cmake_rst.py
@@ -37,11 +37,12 @@ import argparse
 import os
 import re
 import sys
+import json
 
 '''Simple superficial API doc generator for .cmake files'''
 
 
-def crawl_for_cmake(path):
+def crawl_for_cmake(path, excluded_files=[]):
     '''
     Crawls over path, looking for files named *.cmake,
     returns tuple of full and relative path.
@@ -49,7 +50,8 @@ def crawl_for_cmake(path):
     cmake_files = []
     for (parentdir, _, files) in os.walk(path):
         for filename in files:
-            if not filename.endswith('.cmake'):
+            if not filename.endswith('.cmake') or \
+                  filename in excluded_files:
                 continue
             fullpath = os.path.join(parentdir, filename)
             relpath = os.path.relpath(fullpath, path)
@@ -57,7 +59,7 @@ def crawl_for_cmake(path):
     return cmake_files
 
 
-def generate_rst(files):
+def generate_rst(files, skip_private=False, skip_undocumented=False):
     '''
     Each of the CMake files is traversed line by line, looking for
     lines like function(...) or macro(...).  For each of these,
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thank you for the PR. I will add a couple of inline comments. It would be great if you can apply them.

doc/generate_cmake_rst.py
```diff
@@ -37,11 +37,12 @@
 import os
 import re
 import sys
+import json
```
dirk-thomas: Please move the import some lines up to be in alphabetical order (PEP 8).

doc/generate_cmake_rst.py
```diff
@@ -37,11 +37,12 @@
...
 
 '''Simple superficial API doc generator for .cmake files'''
 
 
-def crawl_for_cmake(path):
+def crawl_for_cmake(path, excluded_files=[]):
```
dirk-thomas: The default value should not be `[]` due to side effects in multiple invocations. Please use `None` instead.

doc/generate_cmake_rst.py
```diff
@@ -49,7 +50,8 @@ def crawl_for_cmake(path):
     cmake_files = []
     for (parentdir, _, files) in os.walk(path):
         for filename in files:
-            if not filename.endswith('.cmake'):
+            if not filename.endswith('.cmake') or \
+                  filename in excluded_files:
```
dirk-thomas: The wrapped line needs to be indented 4 spaces.

doc/generate_cmake_rst.py
```diff
@@ -155,10 +161,26 @@ def generate_rst(files):
     parser = argparse.ArgumentParser(description='Crawls a path for .cmake files and extract documentation of functions and macros into reStructured text.')
     parser.add_argument('path', nargs='?', default='.', help='The path to be crawled')
     parser.add_argument('-o', '--output', help='The name of the generated rst file')
+    parser.add_argument('--skip_private', action="store_true", help='skip documented items not marked with @public')
+    parser.add_argument('--skip_undocumented', action="store_true", help='skip items without documentation')
```
dirk-thomas: Please start the help with an upper case character to match the other help texts.

doc/generate_cmake_rst.py
```diff
@@ -155,10 +161,26 @@ def generate_rst(files):
...
+
     args = parser.parse_args()
 
-    cmake_files = crawl_for_cmake(args.path)
-    lines = generate_rst(cmake_files)
+    exclusions = '{}/.sphinx_exclusions.json'.format(args.path)
+    excluded_files = []
+    try:
+        if os.path.exists(exclusions):
```
dirk-thomas: Please move the condition outside of the try / catch block because it shouldn't raise.

doc/generate_cmake_rst.py
```diff
@@ -155,10 +161,26 @@ def generate_rst(files):
...
+            # print('exclusions={}'.format(exclusions))
```
dirk-thomas: Please remove the comment.

doc/generate_cmake_rst.py
```diff
@@ -155,10 +161,26 @@ def generate_rst(files):
...
+            with open(exclusions) as f:
```
dirk-thomas: Please specify the second arg `'r'` to be explicit.

doc/generate_cmake_rst.py
```diff
@@ -155,10 +161,26 @@ def generate_rst(files):
...
+                excluded_files = json.load(f)
+    except:
```
dirk-thomas: Please only catch very specific exceptions.

scpeters:
:+1:

dirk-thomas:
I have generated the update API with this - it looks much better. Thanks.

Squashed and merged in a84a31227d70f83bbaff85c94a274d559fb7603a

PR #723, opened 2015-03-30
Fixes regex for library config types

stuglaser:
The old regex, `^debug|optimized|general$`, matches "debug_foo", "foo_optimized_bar", and "foo_general" because `^` and `$` have precedence over `|`.  This breaks certain library names, for example, `lib_kalman_general`.

The new regex fixes the order of operations.


dirk-thomas:
Great catch. Thank you for the patch!

PR #724, opened 2015-04-02
Fix python error when working with non-ascii characters in path

tmoinel:
This is a recurrent error when working in non-english local system. Non-ascii charaters in the path to the catkin workspace make the cmake execution fail :

```
File "/home/moinel/éàç/src/visual_actimetry_msgs/build/devel/_setup_util.py", line 262
SyntaxError: Non-ASCII character '\xc3' in file /home/moinel/éàç/src/visual_actimetry_msgs/build/devel/_setup_util.py on line 262, but no encoding declared; see http://www.python.org/peps/pep-0263.html for details
```

It is maybe worth specifying utf-8 encoding to every python script.

### DIFF
--- a/cmake/templates/__init__.py.in
+++ b/cmake/templates/__init__.py.in
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 # generated from catkin/cmake/template/__init__.py.in
 # keep symbol table as clean as possible by deleting all unnecessary symbols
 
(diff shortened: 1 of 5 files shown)


wjwwood:
+1

> It is maybe worth specifying utf-8 encoding to every python script.

I would agree, but further scope the statement to "every _templated_ python script".

tmoinel:
I add utf-8 encoding to 3 others templated python script.

I don't know catkin and python well enough to figure out if the others templated python scripts need encoding too. I can do the change if someone confirms.

scpeters:
:+1:

dirk-thomas:
Thank you for the PR. Squashed and merged in 1b90e571b6683bb2175236b91b3d823debc72161

PR #725, opened 2015-04-02
setup.sh.in: honor $TMPDIR if it exists

### LINKED ISSUE #710: setup.sh hardcodes /tmp, should honor $TMPDIR
Hello,

in our environment (Ubuntu Snappy), packaged apps cannot currently access `/tmp/`, but the environment sets `$TMPDIR` (and `$TEMPDIR`) to an appropriate place, see https://en.wikipedia.org/wiki/TMPDIR. But catkin's setup.sh.in doesn't honor that, but does

```
 _SETUP_TMP=`mktemp /tmp/setup.sh.XXXXXXXXXX`
```

It should call `mktemp --tmpdir setup.sh.XXXXXXXXXX` instead to honor those.

This is at https://github.com/ros/catkin/blob/indigo-devel/cmake/templates/setup.sh.in#L47 at the time of this writing.

scpeters:
Fixes #710.

### DIFF
--- a/cmake/templates/setup.sh.in
+++ b/cmake/templates/setup.sh.in
@@ -44,12 +44,20 @@ if [ -z "$CATKIN_SHELL" ]; then
 fi
 
 # invoke Python script to generate necessary exports of environment variables
-_SETUP_TMP=`mktemp /tmp/setup.sh.XXXXXXXXXX`
+if [ -n "${TMPDIR}" ]; then
+  # honor TMPDIR if it exists
+  _TMPDIR="${TMPDIR}"
+else
+  # otherwise use /tmp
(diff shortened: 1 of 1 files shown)


scpeters:
I unset the variable and also added quotes in case `TMPDIR` has spaces, using the following to test:

```
TMPDIR=$HOME/tmp\ spaces ./src/catkin/bin/catkin_make_isolated
```

dirk-thomas:
Thanks! I have merged a slightly modified version: 3c91601d769984839c510fb5f448879db3e59eaf

It uses `-d` for the check (instead of `-n`) to ensure that the TMPDIR actually exists.

scpeters:
good call on `-d`

Issue #694, opened 2014-11-13
ld-2.24.90 does not support -l:/full/path/to/library.so


tho-:
pkg-config files are generated with items like -l:/full/path/to/library.so.
(See https://github.com/ros/catkin/blob/indigo-devel/cmake/catkin_package.cmake#L264 )

This does not seem to work with ld-2.24.90. Only -l:library.so works (also, -L/ -l:/full/path/ seems to work, but it's not clear if this is an intended behaviour or not). The ld manual only talks about -l:library.so.

I could not find yet it this is a regression in ld or a deliberate change.

One option could be to change catkin so that it generates proper -L/full/path -lrary in the .pc to link with /full/path/library.so.
Another option is to add an explicit -L/ in front of the -l: items.

Any comment?

dirk-thomas:
Using `-l:/path/to/lib` is necessary to prevent `pkg-config` from reordering libraries.

This was working well until now. What system are you experiencing this?

Using `-L` and `-l` is not really applicable since it changes resolution of libraries and would break workspace chaining. I don't know a different way to maintain the current behavior.

tho-:
On Thursday, at 10:13, Dirk Thomas wrote:
| Using -l/path/to/lib is necessary to prevent pkg-config from reordering libraries.

You mean -l:/path/to/lib? (-l/path/to/lib has never worked).

AFAIK, pkg-config does not reorder libraries (this would be a bug). But yes, I
understand your point that -L and -l are not useable.

| This was working well until now. What system are you experiencing this?

ld-2.24.90, from ubuntu-14.10

% ld --version
GNU ld (GNU Binutils for Ubuntu) 2.24.90.20141014
% gcc -xc /dev/null -l:/usr/lib/x86_64-linux-gnu/libc.so
/usr/bin/ld: cannot find -l:/usr/lib/x86_64-linux-gnu/libc.so
collect2: error: ld returned 1 exit status

dirk-thomas:
Yes, `-l:/path/to/lib`.

It sadly _does_ reorder libraries - please see more info in #300.

If that effects _Utopic_ it will be "problematic" to build any dry packages which depend on pkg-config.

tho-:
On Thursday, at 11:13, Dirk Thomas wrote:
| It sadly does reorder libraries - please see more info in #300.

I do not see any reordering issue in #300 ?
#300 is about fixing -l/path/to/lib by using -l:/path/to/lib

[...]

Issue #300, opened 2013-01-10
Link flag error


hsu:
Some of the catkin added link flags in the package installed `.pc` files appears incorrect.  For example, `ros-groovy-console-bridge` installs:

`/opt/ros/groovy/lib/pkgconfig/console_bridge.pc`

which contains a line:

```
Libs: -L/opt/ros/groovy/lib -lconsole_bridge -l/usr/lib/libboost_thread-mt.so -lpthread
```

But the `libboost_thread_mt` link flag is wrong, it should be

```
Libs: -L/opt/ros/groovy/lib -lconsole_bridge /usr/lib/libboost_thread-mt.so -lpthread
```

The extra `-l` triggers a compile error:

```
Linking CXX shared library libgazebo_sdf_interface.so
/usr/bin/ld: cannot find -l/usr/lib/libboost_thread-mt.so
```

It seems `catkin_depends` might be adding the extra boost library link flags incorrectly?

dirk-thomas:
Please provide a description what you do to reproduce the failure.

hsu:
checkout [simulator_gazebo from trunk](https://code.ros.org/svn/ros-pkg/stacks/simulator_gazebo/trunk/), edit [gazebo/CMakelists.txt](https://code.ros.org/svn/ros-pkg/stacks/simulator_gazebo/trunk/), modify the line:

```
set(USE_HG false)
```

to

```
set(USE_HG true)
```

then

```
rosmake gazebo
```

this will checkout gazebo into `gazebo/build/gazebo-hg`.  When compile is finished, edit [gazebo/build/gazebo-hg/cmake/SearchForStuff.cmake](https://bitbucket.org/osrf/gazebo/src/d9edb000a87db61d30b85c7f14885be451dda876/cmake/SearchForStuff.cmake?at=deprecated_parser_sdf_1.2_support)  and modify [line 268](https://bitbucket.org/osrf/gazebo/src/d9edb000a87db61d30b85c7f14885be451dda876/cmake/SearchForStuff.cmake?at=deprecated_parser_sdf_1.2_support#cl-268), change `find_package` to `pkg_check_modules`:

```
pkg_check_modules(console_bridge console_bridge)
```

Next,

```
rm -rf gazebo/build/gazebo-hg/build
```

then, re-compile, and you should be able to see the error.

```
rosmake gazebo
```

PR #727, opened 2015-04-02
Replace l:$absolute_path in pkgconfig by just the absolute path

j-rivero:
Fix for issue #694.

The syntax `-l:$library_absolute_path` seems to be unsupported by upstream ld and it is failing in some versions of the linker.

This patch workaround about the problem by generating pkg-config files which does not set the `-l` (only works with relative paths) and just provide the absolute path to the library. The compiler will link the library as a shared lib (if `-static` was not supplied).

### DIFF
--- a/cmake/catkin_package.cmake
+++ b/cmake/catkin_package.cmake
@@ -261,7 +261,7 @@ function(_catkin_package)
     if(IS_ABSOLUTE ${library})
       get_filename_component(suffix ${library} EXT)
       if(NOT "${suffix}" STREQUAL "${CMAKE_STATIC_LIBRARY_SUFFIX}")
-        # We are passing the absolute library as it, since the l:<absolute_path> syntax was tricky and deprecated
+        # We are passing the absolute library as it, since the -l:<absolute_path> syntax was tricky and deprecated
         # in some version of ld. See https://github.com/ros/catkin/issues/694 for more information.
         set(library "${library}")
       endif()
--- a/cmake/catkin_package.cmake
+++ b/cmake/catkin_package.cmake
@@ -261,7 +261,9 @@ function(_catkin_package)
     if(IS_ABSOLUTE ${library})
       get_filename_component(suffix ${library} EXT)
       if(NOT "${suffix}" STREQUAL "${CMAKE_STATIC_LIBRARY_SUFFIX}")
-        set(library "-l:${library}")
+        # We are passing the absolute library as it, since the l:<absolute_path> syntax was tricky and deprecated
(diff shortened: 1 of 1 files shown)


cmake/catkin_package.cmake
```diff
@@ -261,7 +261,9 @@ function(_catkin_package)
     if(IS_ABSOLUTE ${library})
       get_filename_component(suffix ${library} EXT)
       if(NOT "${suffix}" STREQUAL "${CMAKE_STATIC_LIBRARY_SUFFIX}")
-        set(library "-l:${library}")
+        # We are passing the absolute library as it, since the l:<absolute_path> syntax was tricky and deprecated
```
wjwwood: Isn't it `-l:<absolute_path>` and not just `l:<absolute_path>`?

wjwwood:
I _think_ this doesn't address the original issue (the reason why we started to do `-l:`). Let me take a stab at explaining. Consider this is the correctly ordered list of link arguments:

```
-l:/path/to/anotherlib
-lsomelib
```

When sorted the above results in the same thing (which is what we want):

```
-l:/path/to/anotherlib
-lsomelib
```

However, if you drop the `-l:`, you start with this:

```
/path/to/anotherlib
-lsomelib
```

But after `pkg-config` reorders things you end up with this:

```
-lsomelib
/path/to/anotherlib
```

Which is not the correct order anymore. I'm a bit foggy on whether or not this was the whole problem, but I remember the library order being paramount. We were certainly abusing the `-l:` of `ld`, but unless we can get `pkg-config` to not reorder the link flags alphabetically then we're gonna have a bad time.

I do remember that this was actually breaking `rosbuild` packages in some cases, but the scenario in which this occurred was pretty extreme.

It might be worth digging through the issue history (this may have even been on code.ros.org) to try to figure the scenario that we were addressing and retesting it now. Honestly it should have been made into a test, but I suppose that's a bit of technical debt we have to live with.

cmake/catkin_package.cmake
j-rivero: Yep, the whole syntax is -l:<absolute_path>. I've fixed it.

dirk-thomas:
I agree with @wjwwood comment. In ros/catkin#300 we specifically prefixed absolute paths with `-l:` to prevent the reordering problem (https://github.com/ros/catkin/commit/b6e6239a291c6d61cb99f181f5a6aaa1e3ffa612).

j-rivero:
I see. Thanks for the detailed explanation. Closing this PR.

Issue #719, opened 2015-02-28
run_tests writes to wrong workspace


bricerebsamen:
I have a chain of workspaces: third, stable, exp, in this order, i.e. stable sources third and exp sources stable.

if I run `catkin_make run_tests` in the stable workspace, the xml result file is written to `exp/build/test_results` instead of `stable/build/test_results`. 

```
-- run_tests.py: execute commands
  /opt/ros/indigo/share/rostest/cmake/../../../bin/rostest --pkgdir=/home/brice/ws/stable/src/velodyne_localizer --package=velodyne_localizer --results-filename tests_short_loop_4-11-05-2014_12-01-05.xml /home/brice/ws/stable/src/velodyne_localizer/tests/short_loop_4-11-05-2014_12-01-05.launch 
... logging to /home/brice/.ros/log/rostest-boox-19288.log
[ROSUNIT] Outputting test results to /home/brice/ws/exp/build/test_results/velodyne_localizer/rostest-tests_short_loop_4-11-05-2014_12-01-05.xml
testtest_localizer_short_loop_4_11_05_2014 ... ok

[ROSTEST]-----------------------------------------------------------------------

[velodyne_localizer.rosunit-test_localizer_short_loop_4_11_05_2014/VelodyneLocalizer][passed]

SUMMARY
 * RESULT: SUCCESS
 * TESTS: 1
 * ERRORS: 0
 * FAILURES: 0

rostest log file is in /home/brice/.ros/log/rostest-boox-19288.log
-- run_tests.py: verify result "/home/brice/ws/stable/build/test_results/velodyne_localizer/rostest-tests_short_loop_4-11-05-2014_12-01-05.xml"
Cannot find results, writing failure results to '/home/brice/ws/stable/build/test_results/velodyne_localizer/MISSING-rostest-tests_short_loop_4-11-05-2014_12-01-05.xml'
```

```
$ ls /home/brice/ws/exp/build/test_results/velodyne_localizer
rostest-tests_short_loop_4-11-05-2014_12-01-05.xml
rosunit-test_localizer_short_loop_4_11_05_2014.xml
$ ls /home/brice/ws/stable/build/test_results/velodyne_localizer/
gtest-test_laser_map.xml
MISSING-rostest-tests_short_loop_4-11-05-2014_12-01-05.xml
```

note that in this examples there are 2 tests in the `velodyne_localizer` package: 
- `test_laser_map` which is a unit test added using `catkin_add_gtest`. The result is written to the appropriate location
[...]

PR #728, opened 2015-04-03
remove CATKIN_TEST_RESULTS_DIR environment variable

dirk-thomas:
This patch removes the CATKIN_TEST_RESULTS_DIR environment variable. That should prevent any kind of "cross talk" between packages build in parallel since the location is only depending on configure-time information.

The user can still override the location by setting the CMake variable via `-DCATKIN_TEST_RESULTS_DIR=...`.

Addresses #719.

### DIFF
--- a/bin/catkin_test_results
+++ b/bin/catkin_test_results
@@ -12,11 +12,8 @@ from catkin.test_results import aggregate_results, print_summary, test_results
 
 
 def main():
-    env_name = 'CATKIN_TEST_RESULTS_DIR'
-    test_results_dir = os.environ[env_name] if env_name in os.environ else False
-
     parser = argparse.ArgumentParser(description='Outputs a summary of the test results. If there are any test errors or failures the scripts return code is 1.')
-    parser.add_argument('test_results_dir', nargs='?' if test_results_dir else None, default=test_results_dir, help='The path to the test results (necessary when the environment variable "%s" is not defined)' % env_name)
+    parser.add_argument('test_results_dir', nargs='?', default=os.curdir, help='The path to the test results')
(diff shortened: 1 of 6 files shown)


tfoote:
lgtm

esteve:
+1

bricerebsamen:
that fixed #719 for me.

It prints a lot of lines about skipping packages such as the one below, but it finally reports the correct number of tests, success, etc.

```
Skipping "src/driving_public/carviz_plugins_imagery/plugin_description.xml": 'tests'
```

wjwwood:
lgtm, however, I am concerned about how we roll this out since it is a pretty serious change to remove an env variable like this at such a low level in the system. I talked with @dirk-thomas about it and we agreed that the right thing to do was to go forward with it, but to be proactive about communicating this to the users and giving them a chance to comment before releasing this change.

dirk-thomas:
@bricerebsamen The script searches recursively for `.xml` - by default starting in the current folder. You might want to pass the test results folder to narrow the search and prevent non-unit test results to show. For a non-isolated build with default paths that would be:

```
catkin_test_results build/test_results
```

**Update:** The script can not use an environment variable from the devel space for the exact location since the devel space is commonly not sourced when running tests. Also the naming scheme of the workspace can be custom and the script can not infer the correct location of the build space.

mikepurvis:
Change LGTM. None of our tooling relies on setting `CATKIN_TEST_RESULTS_DIR`.

aleeper:
+1.

dirk-thomas:
Waiting for feedback after announcement: https://groups.google.com/forum/#!topic/ros-sig-buildsystem/suBvWzoGODU

floweisshardt:
+1. None of our tooling uses 'CATKIN_TEST_RESULTS_DIR'.

PR #731, opened 2015-04-13
avoid warning with CMake 3.1 and newer

dirk-thomas:
Different approach to address #729.


scpeters:
It eliminates the warnings I referenced in #729, and is a more conservative approach. :+1:

wjwwood:
+1

Issue #718, opened 2015-02-13
file COPY cannot set permissions for 05.catkin_make.bash


k-okada:
we sometimes see following error when we compile multiple packages in parallel, I still can not find a way to reproduce this issue perfectly, but it occurs sometimes. It seems we have to not to overwrite on 
https://github.com/ros/catkin/blob/indigo-devel/cmake/catkin_add_env_hooks.cmake#L91, or find any other way to write atomic copy.

```
[jsk_recognition_msgs] CMake Error at /opt/ros/indigo/share/catkin/cmake/catkin_add_env_hooks.cmake:91 (file): 
[jsk_recognition_msgs]   file COPY cannot set permissions on 
[jsk_recognition_msgs]   "/workspace/ros/ws_jsk_recognition/devel/etc/catkin/profile.d/05.catkin_make.bash" 
[jsk_recognition_msgs] Call Stack (most recent call first): 
[jsk_recognition_msgs]   /opt/ros/indigo/share/catkin/cmake/all.cmake:198 (catkin_add_env_hooks) 
[jsk_recognition_msgs]   /opt/ros/indigo/share/catkin/cmake/catkinConfig.cmake:20 (include) 
[jsk_recognition_msgs]   CMakeLists.txt:3 (find_package) 
[jsk_recognition_msgs]  
```

dirk-thomas:
Please provide an example command sequence for which it at least happens sometimes.

k-okada:
ok, I finally able to find the way to reproduce this issue, see https://travis-ci.org/k-okada/test_catkin_parallel/jobs/50891940#L2704

wjwwood:
So the file in question (based on your travis-ci job) being installed which presumably collides is `05.catkin_make_isolated.bash`, but my question is why is this file getting installed for every package? It is the tab completion for `catkin_make_isolated`, that should only get installed by catkin right?

If you build multiple packages in parallel which create the same file then you can get a collision like this. The question is why is this file being installed by more than one package.

In the original post it is the tab completion for `catkin_make` which collides. The installation of these files seems to be done in `all.cmake` in `catkin`:

https://github.com/ros/catkin/blob/2599eb973762c2a23f103fdc8bd0ef40e779f2ed/cmake/all.cmake#L197-L203

@dirk-thomas why is that being done?

k-okada:
[...]

PR #732, opened 2015-04-13
only install environment hooks for catkin_make(_isolated) completion in the catkin package

dirk-thomas:
This will address #718.


wjwwood:
+1

PR #733, opened 2015-04-16
show skipped test results file only when using --all

dirk-thomas:
With #728 being merged the script `catkin_test_results` will not be able to crawl the test results folder by default. Instead it uses the current directory (commonly the workspace root). This usually results in more xml files to be found which are not actually test result files.

Therefore this patch only outputs the skipped result files when the `--all` option is passed to reduce the verbosity of the tool.


tfoote:
+1

esteve:
+1

Issue #737, opened 2015-05-22
cakin_make_isolated with custom test results dir


dirk-thomas:
Moved from ros-infrastructure/ros_buildfarm#77.

When `catkin_make_isolated` is being invoked with a custom `CATKIN_TEST_RESULTS_DIR` each package will clean the test results folder before running its own tests. Therefore only leaving the test results of the last package around.

PR #738, opened 2015-05-22
only clean the test results of the current project

dirk-thomas:
Addresses #737.

### DIFF
--- a/cmake/test/remove_test_results.py
+++ b/cmake/test/remove_test_results.py
@@ -1,0 +1,32 @@
+#!/usr/bin/env python
+
+from __future__ import print_function
+
+import argparse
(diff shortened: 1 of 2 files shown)


esteve:
+1

tfoote:
+1

dirk-thomas:
I always thought that wiping the whole directory might be dangerous. Therefore I updated the clean commands to only remove files ending with `.xml`. So if someone passes in a "bad" test results dir it only wipes xml files and not everything.

tfoote:
+1 that's safer

wjwwood:
+1

esteve:
+1 again

PR #741, opened 2015-06-22
Add CONFIGURATIONS option to catkin_install_python

jakalope:
#### Extends functionality in the direction of the built-in CMake install() function.

**Use case 1:** install my script only for Debug and RelWithDebInfo CMAKE_BUILD_TYPEs (not, for example, Release)

``` cmake
catkin_install_python(PROGRAMS scripts/my_executable_python_script 
    CONFIGURATIONS Debug RelWithDebInfo
    DESTINATION ${CATKIN_GLOBAL_BIN_DIRECTORY}
)
```

**Use case 2:** install my script only for Release CMAKE_BUILD_TYPEs (not, for example, Debug)

``` cmake
catkin_install_python(PROGRAMS scripts/my_executable_python_script 
    DESTINATION ${CATKIN_GLOBAL_BIN_DIRECTORY}
    CONFIGURATIONS Release
)
```


dirk-thomas:
Instead of making this function more complex it should be possible to check the current configuration outside and only conditionally call this function. Would that be an option for your use case?

jakalope:
It would work. I would still wrap it in another function to avoid duplicate
code, but if you feel it isn't wanted upstream I will keep it to myself ;-)

dirk-thomas:
I don't see a strong use case to install Python scripts only for a specific subset of build configurations. Without knowing your use case I would assume that it is not a common one?

Conditionally calling the function sounds like a good approach to me.

If you use this more often it makes total sense to wrap the functionality in a function.

dirk-thomas:
I will close this since it is possible to work around this in user land code.

PR #747, opened 2015-07-10
fixing package_dir location logic.

asmodehn:
package_dir: {'foo':'lib'} means 'lib/**init**.py' and not 'lib/foo/**init**.py'

### DIFF
--- a/cmake/interrogate_setup_dot_py.py
+++ b/cmake/interrogate_setup_dot_py.py
@@ -76,9 +76,9 @@ def _get_locations(pkgs, package_dir):
                 if key in package_dir:
                     locations[key] = package_dir[key]
                 elif parent_location is not None:
-                    locations[key] = parent_location
+                    locations[key] = os.path.join(parent_location,splits[key_len])
                 else:
-                    locations[key] = allprefix
+                    locations[key] = os.path.join(allprefix,key)
             parent_location = locations[key]
     return locations
 
@@ -117,7 +117,7 @@ def generate_cmake_file(package_name, version, scripts, package_dir, pkgs, modul
             continue
         # check every child has the same root folder as its parent
         parent_name = '.'.join(splits[:1])
-        if location != locations[parent_name]:
+        if not locations[parent_name] in location:
             raise RuntimeError(
                 "catkin_export_python does not support setup.py files that combine across multiple directories: %s in %s, %s in %s" % (pkgname, location, parent_name, locations[parent_name]))
 
@@ -126,7 +126,7 @@ def generate_cmake_file(package_name, version, scripts, package_dir, pkgs, modul
 
     resolved_pkgs = []
     for pkg in pkgs:
-        resolved_pkgs += [os.path.join(locations[pkg], pkg)]
+        resolved_pkgs += [locations[pkg]]
 
     result.append(r'set(%s_PACKAGES "%s")' % (prefix, ';'.join(pkgs)))
     result.append(r'set(%s_PACKAGE_DIRS "%s")' % (prefix, ';'.join(resolved_pkgs).replace("\\", "/")))


asmodehn:
It seems some tests should also be changed... I d rather have some feedback about this pull request before diving into all the tests that might need to be changed to follow the same behavior as the distutils behavior ( that you can test with `python setup.py install` ) so that our devel/ and install/ folder endup being similar.

asmodehn:
references : 
https://docs.python.org/2/distutils/setupscript.html#listing-whole-packages
https://docs.python.org/2/distutils/examples.html#pure-python-distribution-by-package

dirk-thomas:
Can you please provide some more information what the current problem is as well as step-by-step instructions how to reproduce the problem.

asmodehn:
Current problem :
If your files structure to access your python modules is : 

```
python-mod-name/setup.py
python-mod-name/python_module_1/__init__.py
python-mod-name-2/setup.py
python-mod-name-2/python_module_2/__init__.py
```

which enables you to have submodules pointing to the source repository of a package by the way, instead of the usual : 

```
src/python_module_1/__init__.py
src/python_module_2/__init__.py
```

and if you have in your setup.py :

```
d = generate_distutils_setup(
    packages=[
        'python_module_1',
        'python_module_2',
    ],
    package_dir={
        'python_module_1': 'python-mod-name',
        'python_module_2': 'python-mod-name-2',
    })
```

the `catkin_make` build works fine. The `__init__.py` is found in `devel/lib/python2.7/dist-packages/python_module`.
However `catkin_make install` doesn't install the package properly. instead the `setup.py` file ends up in `install/lib/python2.7/dist-packages/python_module.`
Also `python setup.py install` is completely broken.
You can test that on this revision : https://github.com/asmodehn/flask-ext-catkin.git @ de04b20773460d86f7d19b3970eb5728c09e901b

Using the proper distutils parameter for the catkin version also doesn't work. :
Using `'python_module_1': 'python-mod-name/python_module_1',` makes catkin expect the `__init__.py` file in `python-mod-name/python_module_1/python_module_1` which is incorrect.

The reason is that the logic to find the location implemented in catkin is different than the logic implemented in distutils ( which makes things very confusing )

Solution : fixing the logic used to determine locations, and

```
d = generate_distutils_setup(
    packages=[
        'python_module_1',
        'python_module_2',
    ],
    package_dir={
        'python_module_1': 'python-mod-name/python_module_1',
        'python_module_2': 'python-mod-name-2/python_module_2',
    })
```

which matches the distutils logic and solves both problem.
`catkin_make` build and `catkin_make install` now work properly.
also `python setup.py install` works as expected.
Implemented here : https://github.com/asmodehn/flask-ext-catkin.git @ 41fd3f7077ec7d8cb4a54a89ed830a212c387a5d

cmake/interrogate_setup_dot_py.py
```diff
@@ -117,7 +117,7 @@ def generate_cmake_file(package_name, version, scripts, package_dir, pkgs, modul
             continue
         # check every child has the same root folder as its parent
         parent_name = '.'.join(splits[:1])
-        if location != locations[parent_name]:
+        if not locations[parent_name] in location:
```
dirk-thomas, line 120: I guess this is supposed to check if the parent location is actually a prefix of the location? Currently this could match anywhere within the string - not only at the start.

dirk-thomas:
Thank you for the patch!

I updated the condition I have commented on in #751. In that branch I have also updated the tests to match the new behavior.

Can you please check and confirm that the new PR also works for you?

asmodehn:
Thanks ! I will check it over the next few days and let you know.

dirk-thomas:
I am closing this in favor of the new PR #751.

PR #752, opened 2015-08-07
Replace write with append for better concurrency 

cmansley:
This mitigates the race condition somewhat, but does not solve it! This should be really fixed by a complete tooling change.

### DIFF
--- a/cmake/all.cmake
+++ b/cmake/all.cmake
@@ -24,7 +24,7 @@ if(EXISTS ${_catkin_marker_file})
   file(READ ${_catkin_marker_file} _existing_sourcespaces)
   if(_existing_sourcespaces STREQUAL "")
     # write this sourcespace to the marker file
-    file(WRITE ${_catkin_marker_file} "${CMAKE_SOURCE_DIR}")
+    file(APPEND ${_catkin_marker_file} "${CMAKE_SOURCE_DIR}")
   else()
     # append to existing list of sourcespaces if it's not in the list
     list(FIND _existing_sourcespaces "${CMAKE_SOURCE_DIR}" _existing_sourcespace_index)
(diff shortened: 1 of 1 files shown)


cmansley:
I should have mentioned that this fix lowers my failures by an order of magnitude using this script https://gist.github.com/cmansley/c6fe6424f139025d9dd4
which I modified from @jbohren. I wanted to test how much the 4k limit on atomic writes effects the issue, by creating a ton of packages.

wjwwood:
If this works for @cmansley then I don't see any harm in merging it. I'll let @dirk-thomas review it before merging though.

dirk-thomas:
If the file gets replaced between reading it and then writing / append to it wouldn't it be safer to:

```
file(APPEND ${_catkin_marker_file} ";${CMAKE_SOURCE_DIR}")
```

and make sure that all other parts using the marker file can deal with the case that the content might have a leading semicolon?

wjwwood:
Maybe? I don't know how much work that would be.

cmansley:
@dirk-thomas you and @jbohren already had the semi-colon discussion over here: https://github.com/ros/catkin/pull/671

dirk-thomas:
It looks like it is coming down to the same question as in #671 then: in order to use `APPEND` we need to find all locations where this is being used in order to ensure that this doesn't break anything.

cmansley:
This PR is just a hack to lower the exposure to the problem. This is simply fixing an oversight from earlier hacking. In other words, just use APPEND instead of WRITE. This absolutely does not solve the problem. This PR https://github.com/catkin/catkin_tools/pull/196 is the "real" solution. So, instead of hunting down everyone using this file and modify them, I would focus on the real solution. I wish we could extract just the .catkin file locking out of that PR, but it seems to be tied up into the fundamental changes.

dirk-thomas:
As mentioned in my previous comment the patch is not ready to merged as-is. If the race condition happens that another process changes the file between `READ` and `WRITE` it even makes it worse by appending the path without any separator.

And in order to always append the path with a separator in the front it needs to be ensured that every code using this file is continuing to work when it starts with a separate (which means an empty string in the list of paths).

Since the fix doesn't fix the actual problem I won't put any work on this. Therefore I will mark it with the milestone `untargeted`. Please feel free to update the PR to make it more robust or implement a completely different solution (likely separate files for each path?).

dirk-thomas:
Closing due to inactivity.

PR #753, opened 2015-08-19
Updated regex that matches function names 

thaddeustaylor:
It now supports whitespace between function or macro and the parenthesis and supports numbers in the function names and parameter names.

This is a little closer to the cmake grammar (http://www.cmake.org/cmake/help/v3.0/manual/cmake-language.7.html#grammar-token-argument).  I know this is for version 3 but I couldn't find the grammar for version 2+.

I stopped short of actually implementing the grammar because that isn't what this script is for. I just wanted to expand it simply to find a few more valid cases. It will accept non-valid function names like `function (9functionname)` but I am not relying on this for syntax checking.

This is my first open source pull request so if there is something I need to do differently let me know.

Thanks.

Thaddeus

### DIFF
--- a/doc/generate_cmake_rst.py
+++ b/doc/generate_cmake_rst.py
@@ -82,7 +82,7 @@ def generate_rst(files, skip_private=False, skip_undocumented=False):
                 else:
                     last_block.append(line.rstrip('\n'))
             else:
-                declaration = re.match('[a-zA-Z]+\([a-zA-Z_ ]+\)', line)
+                declaration = re.match('[a-zA-Z]+ *\([a-zA-Z0-9_ ]+\)', line)
                 if declaration is None:
                     last_block = []
                     last_block_public = False


doc/generate_cmake_rst.py
```diff
@@ -82,7 +82,7 @@ def generate_rst(files, skip_private=False, skip_undocumented=False):
                 else:
                     last_block.append(line.rstrip('\n'))
             else:
-                declaration = re.match('[a-zA-Z]+\([a-zA-Z_ ]+\)', line)
+                declaration = re.match('[a-zA-Z]+ *\([a-zA-Z0-9_ ]+\)', line)
```
dirk-thomas: Please remove the trailing newline to follow PEP8.

wjwwood: Trailing whitespace?

dirk-thomas: If you insist then that one instead :wink:

thaddeustaylor:
Out of curiosity where is it mentioned in PEP 8 to not have trailing whitespace(or newlines)?  I was looking today but couldn't find it.  

While I should have done this before I ran pylint(1.1.0) and pychecker(0.8.19) on my modified file and it didn't pick up that whitespace.  I know it is possible that there is a bug in those programs but I would be surprised.  

pylint did, however, find a good number of other 'conventions' and 'warnings' that could be fixed.  Is there a reason to only fix the whitespace issue on that line?

No worries about answering, I assume you are all busy, I was just curious about the procedure so next time I can get it right the first time.

It is funny how a simple task, like removing a space, can turn into an hour(s) long task if simply one asks the question, why?

dirk-thomas:
You are right that PEP 8 does not explicitly mention trailing whitespaces. I was under that assumption because linters usually flag it (e.g. pep8 w291 http://pep8.readthedocs.org/en/latest/intro.html or pylint: http://pylint-messages.wikidot.com/messages:c0303).

The existing code might not always follow current guidelines but we aim to update code which is being touched to follow it. Therefore you should only update lines which are part of the patch and not change the formatting of other unrelated lines.

For packages which have multiple branches we commonly avoid style only changes since it makes backporting packages significantly more effort.

No worries to ask questions like this. We are happy to answer them.

thaddeustaylor:
Good to know.

Thanks.

dirk-thomas:
Can you please also squash your two commits - it helps to keep the history better readable. Thanks!

thaddeustaylor:
Does that work?

dirk-thomas:
Perfect - thanks.

PR #755, opened 2015-09-21
Remove -x in python distutils shell script

wjwwood:
This prevents extraneous output in the installation step. According to the man page:

http://heirloom.sourceforge.net/sh/sh.1.html

> -x    Print commands and their arguments as they are executed.

This is the same as activating `BASH_XTRACEFD` (http://linux.die.net/man/1/sh). This is the type of output it prevents:

```
+ '[' -n '' ']'
+ cd /Users/william/jade/src/catkin
+ /usr/bin/env PYTHONPATH=/Users/william/jade/install/lib/python2.7/site-packages:/Users/william/jade/build/catkin/lib/python2.7/site-packages: CATKIN_BINARY_DIR=/Users/william/jade/build/catkin /usr/bin/python /Users/william/jade/src/catkin/setup.py build --build-base /Users/william/jade/build/catkin install --prefix=/Users/william/jade/install --install-scripts=/Users/william/jade/install/bin
```

The reason I noticed and tracked it down was because this is also output on stderr, which the `catkin_tools` command picked up on and reported as a warning (perhaps incorrectly).

### DIFF
--- a/cmake/templates/python_distutils_install.sh.in
+++ b/cmake/templates/python_distutils_install.sh.in
@@ -12,12 +12,14 @@ if [ -n "$DESTDIR" ] ; then
     DESTDIR_ARG="--root=$DESTDIR"
 fi
 
-cd "@INSTALL_CMD_WORKING_DIRECTORY@"
+echo_and_run() { echo "+ $@" ; "$@" ; }
+
+echo_and_run cd "@INSTALL_CMD_WORKING_DIRECTORY@"
 
 # Note that PYTHONPATH is pulled from the environment to support installing
 # into one location when some dependencies were installed in another
 # location, #123.
-/usr/bin/env \
+echo_and_run /usr/bin/env \
     PYTHONPATH="@CMAKE_INSTALL_PREFIX@/@PYTHON_INSTALL_DIR@:@CMAKE_BINARY_DIR@/@PYTHON_INSTALL_DIR@:$PYTHONPATH" \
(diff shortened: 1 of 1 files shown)


tfoote:
I'm not sure I'd call the outputs extraneous. That information is there so that you can see what exactly the build is running. There's no other way to get visibility into the internal state of this script unless it prints it out. This allows someone debugging the build to be able to reproduce it quite easily. Otherwise you'd only see the output of setup.py and never know what arguments or paths were passed to it when it was run.

wjwwood:
@tfoote then we should use `echo`'s as not to output on `stderr` for normal behavior, or run `set -x` only when a trace or debug environment variable is set, something like that.

Also, at least the first line or output is not very useful and honestly more confusing than helpful if you don't know where it is coming from: `+ '[' -n '' ']'`

dirk-thomas:
Using `echo` to print the major command instead of `-x` seems to be a good idea.

wjwwood:
Ok, I added https://github.com/ros/catkin/commit/8e2a9484ba5aa20618e9784ef4f868a078f9037a, but I'm still testing it in a bigger workspace now, and we should test it on Linux too.

wjwwood:
Seems to work for me on Linux and OS X.

tfoote:
+1

dirk-thomas:
+1 I can confirm that the new messages go to `stdout` and `stderr` stay clean.

Since `catkin` has just been released into _Indigo_ and _Jade_ it will take a while until this patch appears in a patch release though.

PR #757, opened 2015-09-29
[python] Collect path data which can be resused for find_in_workspaces	

wkentaro:
For https://github.com/ros/ros_comm/pull/676

### DIFF
--- a/python/catkin/find_in_workspaces.py
+++ b/python/catkin/find_in_workspaces.py
@@ -75,6 +75,27 @@ def _get_valid_search_dirs(search_dirs, project):
     return search_dirs
 
 
+def collect_path_data_in_workspaces(workspaces=None,
+                                 workspace_to_source_spaces=None,
+                                 source_path_to_packages=None):
+    """Collect path data (workspaces, source_paths and package_paths)
+    in specified workspaces."""
+    if workspaces is None:
+        workspaces = get_workspaces()
+    if workspace_to_source_spaces is None:
+        workspace_to_source_spaces = {}
+    if source_path_to_packages is None:
+        source_path_to_packages = {}
+
+    for ws in workspaces:
+        if ws not in workspace_to_source_spaces:
+            workspace_to_source_spaces[ws] = get_source_paths(ws)
+        for src_path in workspace_to_source_spaces[ws]:
+            if src_path not in source_path_to_packages:
+                source_path_to_packages[src_path] = find_packages(src_path)
+    return workspaces, workspace_to_source_spaces, source_path_to_packages
+
+
 # OUT is always a list of folders
 #
 # IN: project=None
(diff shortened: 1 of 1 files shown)


python/catkin/find_in_workspaces.py
```diff
@@ -75,6 +75,27 @@ def _get_valid_search_dirs(search_dirs, project):
     return search_dirs
 
 
+def collect_path_data_in_workspaces(workspaces=None,
```
wkentaro, line 78: Actually this function cannot be resused in `find_in_workspaces()` because `find_in_workspaces()` has `first_matching_workspace_only` arg.
And reusing can cause low performance for it, because this function see all workspaces and cannot stop after first match.

python/catkin/find_in_workspaces.py
```diff
@@ -87,7 +108,7 @@ def _get_valid_search_dirs(search_dirs, project):
 #      except for s == 'share', cand is a list of two paths: ws[0] + s + project (+ path) and ws[1] + project (+ path)
 #      add cand to result list if it exists
 #      is not defined for s in ['bin', 'lib'], bailing out
-def find_in_workspaces(search_dirs=None, project=None, path=None, _workspaces=get_workspaces(), considered_paths=None, first_matching_workspace_only=False, first_match_only=False, workspace_to_source_spaces=None, source_path_to_packages=None):
+def find_in_workspaces(search_dirs=None, project=None, path=None, _workspaces=None, considered_paths=None, first_matching_workspace_only=False, first_match_only=False, workspace_to_source_spaces=None, source_path_to_packages=None):
```
wkentaro, line 111: This is not refactoring.

https://github.com/ros/ros_comm/pull/676 depends on this change.
at https://github.com/ros/ros_comm/pull/676/files#diff-66abf4062401099443c44fc29a06e40aR36 .

dirk-thomas, line 111: Can you please describe why this change is necessary. The referenced line in `ros_comm` should work the same - with and without this patch.

python/catkin/find_in_workspaces.py
dirk-thomas, line 78: This function does not provide any new information over the existing catkin API. The algorithm calling the other existing API is trivial. Please move this function to ros/ros_comm#676 where it is being used.

python/catkin/find_in_workspaces.py
wkentaro, line 111: Sorry, I commented this before an amending in the commit at ros_comm, so now this change is not needed. I will remove this change.

python/catkin/find_in_workspaces.py
wkentaro, line 78: Ok, I amended the commit at ros/ros_comm#676.

wkentaro:
the all commits in this PR is not needed now.
So I close this.
Also see comments above at https://github.com/ros/catkin/pull/757#discussion_r41495673 https://github.com/ros/catkin/pull/757#discussion_r41493467 .

Issue #428, opened 2013-05-27
chained workspaces and include search path


stonier:
Ran across some search path problems for includes if you have chained source workspaces sitting on top of released debs.
- Old `ecl` in debs residing in `/opt/ros/groovy`.
- New source workspace for ecl with some changes in `/opt/ecl`.
- Chained workspace for kobuki on top of compiled ecl on top of installed groovy.

I source `/opt/ecl/devel/setup.bash` and proceed to `catkin_make` the `/opt/kobuki` workspace. I got some wierd errors, but just to test clearly, I removed some files that were included in the `/opt/ros/groovy/ecl/sigslots.hpp` (the one that kobuki calls) so that it would throw an error if it was using `/opt/ros/groovy` headers before `/opt/ecl/src/ecl_core/ecl_sigslots/include` headers. 

```
[...]

PR #758, opened 2015-10-05
_setup_util: do not set CPATH

dmerejkowsky:
It messes up with other build systems


dirk-thomas:
The problem with removing this environment variable now is that existing code might rely on it being present. Therefore it is hard to tell if this patch will break existing stuff.

Can you provide some more information which build system you are using and how it messes it up? As a workaround you can `unset CPATH`.

dmerejkowsky:
Say you are working on the foo lib that's already installed by ros. Your build system sets
-I src/foo/include, but CPATH is set to /opt/ros-indig/include, so you can end up using the wrong version of the foo headers.

Moreover, QtCreator for instance does not know about CFLAGS and you will get confused because everything will look fine from the IDE, but will cease to work when building from the command line.

> As a workaround you can unset CPATH.

Or we could tell people using catkin to always use include_directory(${catkin_INCLUDE_DIRS}) in their CMakeLists :)

dirk-thomas:
`CPATH` is being considered _after_ all other include directories (https://gcc.gnu.org/onlinedocs/cpp/Environment-Variables.html). Therefore in your case `/opt/ros-indigo/include` would not overlay `src/foo/include`. The package local include directories are always considered first (which is the expected behavior for overlaying packages).

Any CMake package which uses `find_package(foo)` should usually later also call `include_directories(${foo_INCLUDE_DIRS})`. That is commonly the case for ROS packages. Which package are you referring to which doesn't do that?

dmerejkowsky:
> CPATH is being considered after all other include directories 

Not when you use -isystem (for instance because you don't want to care about warning coming from the headers of the lib you are using)

See here for a repository showing the problem:
https://github.com/dmerejkowsky/cpath-considered-harmful

> Which package are you referring to which doesn't do that?

I'm not. _you_ are the one saying it would break existing code, it's not my job to find examples :)

dirk-thomas:
I agree to remove CPATH from the environment. But such a change can't be done for already released distros. I will keep this ticket open to address it with the ROS distro (k-turtle in May 2016).

In general I would not recommend using `system` for the case of suppressing warnings. It will basically break the overlaying of workspaces. There is a similar ticket about that: #428.

dmerejkowsky:
OK. Thanks for the info.

dirk-thomas:
Closing in favor of #783.

PR #762, opened 2015-11-23
fix project specific clean_test_results targets

### LINKED ISSUE #761: Project-specific clean_test_results targets broken / not created properly
The project-specific clean_test_results targets introduced in 52dbdd500f66a558e9149a33db48e7a75c750473 do not work as expected. As a result cmake throws many "dependency does not exist" errors, (e.g. https://travis-ci.org/ipa320/cob_robots/builds/92377717#L4285).

```
CMake Warning (dev) at /opt/ros/indigo/share/catkin/cmake/test/tests.cmake:148 (add_dependencies):
  Policy CMP0046 is not set: Error on non-existent dependency in
  add_dependencies.  Run "cmake --help-policy CMP0046" for policy details.
  Use the cmake_policy command to set the policy and suppress this warning.
  The dependency target "clean_test_results_cob_bringup" of target
  "_run_tests_cob_bringup_roslaunch-check_robots_cob3-2.xml" does not exist.
  /opt/ros/indigo/share/roslaunch/cmake/roslaunch-extras.cmake:35 (catkin_run_tests_target)
  cob_robots/cob_bringup/CMakeLists.txt:36 (roslaunch_add_file_check)
```

In fact instead of per-project targets a single `clean_test_results_Project` is created.
I guess the target definition (https://github.com/ros/catkin/blob/indigo-devel/cmake/test/tests.cmake#L82) must be moved into the `catkin_run_tests_target` function to get `${PROJECT_NAME}` resolved properly.

dirk-thomas:
Related to #738.

This only happend when a workspace is built with a single CMake invocation. In that case the catkin cmake files are being included before the CMake project name  is set.

Fixes #761.

### DIFF
--- a/cmake/test/tests.cmake
+++ b/cmake/test/tests.cmake
@@ -79,11 +79,6 @@ if(NOT TARGET clean_test_results)
   add_custom_target(clean_test_results
     COMMAND ${PYTHON_EXECUTABLE} "${catkin_EXTRAS_DIR}/test/remove_test_results.py" "${CATKIN_TEST_RESULTS_DIR}")
 endif()
-# create target to clean project specific test results
-if(NOT TARGET clean_test_results_${PROJECT_NAME})
-  add_custom_target(clean_test_results_${PROJECT_NAME}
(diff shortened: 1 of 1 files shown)


mathias-luedtke:
I was not able to reproduce the original warnings on my system (trusty comes with cmake 2.8 only), but at least the project-specific targets are created properly with this fix.

PR #763, opened 2015-11-24
fixing exec call to be more robust

asmodehn:
regarding http://bugs.python.org/issue16781

Regarding problems such as : http://answers.ros.org/question/221308/third-party-release-of-python-package/

### DIFF
--- a/cmake/interrogate_setup_dot_py.py
+++ b/cmake/interrogate_setup_dot_py.py
@@ -240,8 +240,7 @@ def main():
         except NameError:
             pass
 
-        with open(args.setupfile_path, 'r') as fh:
-            exec(fh.read())
+        execfile(args.setupfile_path,{})
     finally:
         distutils.core.setup = distutils_backup
         try:
--- a/cmake/interrogate_setup_dot_py.py
+++ b/cmake/interrogate_setup_dot_py.py
@@ -241,8 +241,7 @@ def main():
             pass
 
         with open(args.setupfile_path, 'r') as fh:
-            exec(fh.read(),{})
-
+            exec(fh.read(), {})
     finally:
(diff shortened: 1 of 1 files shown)


cmake/interrogate_setup_dot_py.py
```diff
@@ -241,7 +241,8 @@ def main():
             pass
 
         with open(args.setupfile_path, 'r') as fh:
-            exec(fh.read())
+            exec(fh.read(),{})
```
dirk-thomas: Please add a space after the comma (PEP8).

cmake/interrogate_setup_dot_py.py
```diff
@@ -241,7 +241,8 @@ def main():
...
+
```
dirk-thomas: Please remove the added empty line.

dirk-thomas:
Please comment on the PR after making changes. Otherwise there is no notification email to the maintainers.

dirk-thomas:
Please comment on the PR after making changes. Otherwise there is no notification email to the maintainers.

Can you please squash the commits.

asmodehn:
Done. see https://github.com/ros/catkin/pull/766. Closing this one.

dirk-thomas:
In the future you can simply squash the commits on the existing branch and force push them and the PR will be updated automatically.

asmodehn:
That works only if noone else has cloned my branch ( which I cannot actually know ).
Otherwise rewriting history will break on their side next time they ll try to pull.

wjwwood:
@asmodehn That's true, but no one should be using this pull request branch directly so it doesn't matter. You only need to worry about that on target branches (e.g. `master`, `indigo-devel`, etc.).

Issue #764, opened 2015-11-30
catkin_install_python does not update installed files


mgrrx:
installing a python script using catkin_install_python(...) installs the script correctly when executed for the first time, but, once you change the script, the installed version remains unchanged. The file will be installed again only if you touch the CMakeLists.txt.

I looked at the function in catkin_install_python.cmake and for me it looks like that it only generates the file once and does not introduce a custom target.

dirk-thomas:
Thanks for reporting this. #765 implements a straight forward fix for it. It would also be possible to move the generation logic into the installation step but that would require more changes and also result in a higher install time cost for every file installed by this function. Since commonly these scripts are only calling a main function in the Python package I went for this approach.

dirk-thomas:
Thanks for reporting this. #765 implements a straight forward fix for it. It would also be possible to move the generation logic into the installation step but that would require more changes and also result in a higher install time cost for every file installed by this function. Since commonly these scripts are only calling a main function in the Python package I went for this approach.

It would be great if you could try the PR and confirm that it fixes the problem for you.

mgrrx:
Thanks for fixing this issue. #765 fixes the problem for me.

PR #765, opened 2015-11-30
fix reinstalling Python files installed by catkin_install_python after modifying them

dirk-thomas:
Addresses #764 by ensuring that CMake is being rerun when the file has changed.


wjwwood:
lgtm

PR #766, opened 2015-12-02
fixing exec call to be more robust regarding http://bugs.python.org/issue16781

asmodehn:
squashed version of https://github.com/ros/catkin/pull/763


dirk-thomas:
Thanks.

asmodehn:
Welcome ! Third party releases just became easier :-)

PR #768, opened 2015-12-03
fix default value for _workspaces in find_in_workspaces

### LINKED ISSUE #767: find_in_workspace always look in the same place
While diving into how packages are found for this http://answers.ros.org/question/215600/how-can-i-run-roscore-from-python, I stumbled on this : 

``` python
def find_in_workspaces(search_dirs=None, project=None, path=None, _workspaces=get_workspaces(), considered_paths=None, first_matching_workspace_only=False, first_match_only=False, workspace_to_source_spaces=None, source_path_to_packages=None):
    '''
    Find all paths which match the search criteria.
    All workspaces are searched in order.
    Each workspace, each search_in subfolder, the project name and the path are concatenated to define a candidate path.
    If the candidate path exists it is appended to the result list.
    Note: the search might return multiple paths for 'share' from devel- and source-space.

    :param search_dir: The list of subfolders to search in (default contains all valid values: 'bin', 'etc', 'lib', 'libexec', 'share'), ``list``
    :param project: The project name to search for (optional, not possible with the global search_in folders 'bin' and 'lib'), ``str``
    :param path: The path, ``str``
    :param _workspaces: (optional, used for unit tests), the list of workspaces to use.
    :param considered_paths: If not None, function will append all path that were searched
    :param first_matching_workspace_only: if True returns all results found for first workspace with results
    :param first_match_only: if True returns first path found (supercedes first_matching_workspace_only)
    :param workspace_to_source_spaces: the dictionary is populated with mappings from workspaces to source paths, pass in the same dictionary to avoid repeated reading of the catkin marker file
    :param source_path_to_packages: the dictionary is populated with mappings from source paths to packages, pass in the same dictionary to avoid repeated crawling
    :raises ValueError: if search_dirs contains an invalid folder name
    :returns: List of paths
    '''
```

I think this has some unexpected behavior : 

``` python
Python 2.7.6 (default, Jun 22 2015, 17:58:13) 
[GCC 4.8.2] on linux2
>>> acc = 0
>>> def return_acc_list():
...     global acc
...     acc = acc + 1
...     return [acc]
...     
>>> def testfunc( accdef = return_acc_list()):
...     print accdef
...     
>>> testfunc()
[1]
>>> testfunc()
[1]
```

As you can see here return_acc_list is run only one time.

The same happens for `get_workspaces()`.
If one part of the code tries to `find_in_workspace()` with a list of workspace, then there is no way to change that list list later on, since the list of workspace we look for packages in will always be the same.

=> It makes it impossible to dynamically set workspaces ( by setting CMAKE_PREFIX_PATH for example) in python.
=> it s a hidden and possibly non-obvious behavior.

I m not sure if this is a problem or not in daily use, and if this should be fixed or not, but I wanted to let you know as it might bite some people in some corner cases...

dirk-thomas:
Fixes #767.

While this changes the behavior when being invoked with an explicit `None` value this is hopefully not done anywhere :pray:

A change to not change the default behavior in any case would be to choose a special default value (e.g. `_workspaces="some_default_value_nobody_is_guessing_and_using"`) but it doesn't seem to be worth in this case.

### DIFF
--- a/python/catkin/find_in_workspaces.py
+++ b/python/catkin/find_in_workspaces.py
@@ -87,7 +87,7 @@ def _get_valid_search_dirs(search_dirs, project):
 #      except for s == 'share', cand is a list of two paths: ws[0] + s + project (+ path) and ws[1] + project (+ path)
 #      add cand to result list if it exists
 #      is not defined for s in ['bin', 'lib'], bailing out
-def find_in_workspaces(search_dirs=None, project=None, path=None, _workspaces=get_workspaces(), considered_paths=None, first_matching_workspace_only=False, first_match_only=False, workspace_to_source_spaces=None, source_path_to_packages=None):
+def find_in_workspaces(search_dirs=None, project=None, path=None, _workspaces=None, considered_paths=None, first_matching_workspace_only=False, first_match_only=False, workspace_to_source_spaces=None, source_path_to_packages=None):
     '''
     Find all paths which match the search criteria.
     All workspaces are searched in order.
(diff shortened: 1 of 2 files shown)


wjwwood:
lgtm

tfoote:
+1

Issue #760, opened 2015-10-20
Catkin finds system packages instead of Catkin packages of exported dependencies


calderpg:
If a CMake-findable package exists on the system (for example, there is a `FindPACKAGENAME.cmake` in `/usr/share/cmake-2.8/Modules`), and a Catkin package with the same name exists, `find_package(catkin REQUIRED COMPONENTS PACKAGENAME)` will find the system package, not the Catkin package.

An example of when the problem occurs:
1. You have a package installed to the system that provides a CMake find package file (for example, `Findosg.cmake` for `OpenSceneGraph`)
2. You have a Catkin package named `osg`
3. You have a Catkin package that depends on `osg` via `find_package(catkin REQUIRED COMPONENTS osg)`

In this case, the system package `OpenSceneGraph` will be found, rather than your Catkin package `osg`.

In contrast, the problem does **not**  appear to occur in the comparable case of local Catkin package that shadows a system Catkin package. In this case, Catkin reorders the packages to ensure that the shadow package is processed before the dependant package. This would be much more reasonable behavior in the case of this bug.

In most cases, having a Catkin package with the same name as a system package should be avoided, but given the long list of system CMake files, some of which (like Findosg.cmake, which is deprecated in favor of FindOpenSceneGraph.cmake, but is still shipped for compatibility) have different names than their respective packages, it is fairly easy to accidentally shadow a system package.

wjwwood:
I do not believe that catkin does anything to the ordering when a catkin package overlays a another system version. I think the issue here is that `Findosg.cmake` is a CMake Module and your catkin package called `osg` produces a CMake Config file called `osgConfig.cmake` and I think, based on your error report, that the CMake's `find_package()` macro prefers CMake Modules to CMake Config files. I'm not 100% sure that's the case, but that's what I would guess.

One thing we could do is check that anything passed to `find_package(catkin REQUIRED COMPONENTS ...)` is both found and defines the `<package_name>_FOUND_CATKIN_PROJECT` CMake variable (which is set by all CMake config files generated by catkin).

dirk-thomas:
What @wjwwood described is the intended CMake behavior (https://cmake.org/cmake/help/v2.8.12/cmake.html#command:find_package):

```
If no module is found ... the command proceeds to Config mode.
```

I don't think that `catkin` behave any different then the CMake default behavior. Therefore I would suggest if you only want to find CMake config's but no modules find the specific package separately from catkin using the `NO_MODULE` option.

[...]

PR #772, opened 2016-01-07
use NO_MODULE to find exported catkin dependencies

dirk-thomas:
Addresses #760.


wjwwood:
Seems reasonable to me, but I'd say it has a medium risk of breaking a downstream package since it changes behavior.

dirk-thomas:
Closing in favor of #782.

PR #774, opened 2016-01-21
Add a script to make your catkin workspace from other directories using the CATKIN_WS environment variable

a20r:
It was really annoying to keep a terminal open in my catkin directory so I made a script that just runs catkin_make from the root catkin directory using the CATKIN_WS environment variable


mikepurvis:
You might be better served switching to catkin_tools, which has this feature, among many others.

See: http://catkin-tools.readthedocs.org/en/latest/

wjwwood:
Sorry, I haven't had time to review this.

As @mikepurvis pointed out `catkin_tools` has this feature, but `catkin build` (command in `catkin_tools`) works fundamentally different from `catkin_make`, see: http://catkin-tools.readthedocs.org/en/latest/mechanics.html#legacy-catkin-workflow

I'll also put the disclaimer on `catkin_tools` that it is still in beta, and even though we're making progress towards being able to recommend it over `catkin_make` or `catkin_make_isolated`, it isn't quite there yet.

dirk-thomas:
I am not sure if this feature justifies add a new script into `catkin`. As the previous comments mentioned `catkin_tools` provides it out-of-the-box. And it is also very easy to create an alias for this (since the `-C` option is there) which is not much different from needing to set the environment variable.

dirk-thomas:
Since there was no further comment on this for over a month I will close it for now.

PR #775, opened 2016-01-21
ensure that Python install destination exists

dirk-thomas:
Otherwise `python setup.py install` might fail: e.g. http://build.ros.org/view/Jdoc/job/Jdoc__rospeex__ubuntu_trusty_amd64/6/console


wjwwood:
lgtm

PR #777, opened 2016-02-04
avoid adding nonexistent paths to environment variables

dirk-thomas:
Without sourcing the setup file for a package contributes its nonexistent paths to all environment variables: `CPATH`, `LD_LIBRARY_PATH`, `PATH`, `PKG_CONFIG_PATH` and `PYTHONPATH`.

### DIFF
--- a/cmake/templates/_setup_util.py.in
+++ b/cmake/templates/_setup_util.py.in
@@ -161,6 +161,9 @@ def _prefix_env_variable(environ, name, paths, subfolders):
             path_tmp = path
             if subfolder:
                 path_tmp = os.path.join(path_tmp, subfolder)
+            # skip nonexistent paths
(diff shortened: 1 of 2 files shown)


dirk-thomas:
This will address the problems manifesting in failing `roswtf` tests: http://build.ros.org/job/Idev__ros_comm__ubuntu_trusty_amd64/7/testReport/__main__/TestRostopicOnline/test_offline/

tfoote:
looks good. you should probably add one more test case which checks the non-existent path gets stripped

dirk-thomas:
Good point, now covered in https://github.com/ros/catkin/pull/777/files#diff-ad1f6ee5eed3025b89ceed022166eab8R70

tfoote:
+1

PR #780, opened 2016-02-22
add custom message explaining CMake find_package error messages

dirk-thomas:
Many user questions are based on the not easy to read and interpret CMake error message in the case that `find_package(... REQUIRED)` can't find a certain CMake config file, e.g.:
- http://answers.ros.org/question/227296/catkin_make-error-when-using-mavros/
- http://answers.ros.org/question/189686/catkin_make-fails-weirdly-after-trying-to-create-a-new-package/
- http://answers.ros.org/question/227151/adding-rplidar_ros-as-package-dependency/
- http://answers.ros.org/question/221359/ros_controlconfigcmake-not-found/
- http://answers.ros.org/question/222041/rosaria-make-error-jade/
- http://answers.ros.org/question/67079/how-to-depend-on-pr2_controllers_msgs/

This patch prints a custom message before that which should provide a more helpful description.

### DIFF
--- a/cmake/catkinConfig.cmake.in
+++ b/cmake/catkinConfig.cmake.in
@@ -72,8 +72,17 @@ if(catkin_FIND_COMPONENTS)
 
       # find package component
       if(catkin_FIND_REQUIRED)
-        find_package(${component} REQUIRED NO_MODULE PATHS ${paths}
+        # try without REQUIRED first
+        find_package(${component} NO_MODULE PATHS ${paths}
           NO_DEFAULT_PATH NO_CMAKE_FIND_ROOT_PATH)
+        if(NOT ${component}_FOUND)
+          # show better message to help users with the CMake error message coming up
+          message(STATUS "Could not find the required component '${component}'. "
+            "The following CMake error indicates that you either need to install the package "
+            "with the same name or change your environment so that it can be found.")
+          find_package(${component} REQUIRED NO_MODULE PATHS ${paths}
+            NO_DEFAULT_PATH NO_CMAKE_FIND_ROOT_PATH)
+        endif()
       elseif(catkin_FIND_QUIETLY)
         find_package(${component} QUIET NO_MODULE PATHS ${paths}
           NO_DEFAULT_PATH NO_CMAKE_FIND_ROOT_PATH)


dirk-thomas:
The same logic could be applied to the generated package specific CMake config files (https://github.com/ros/catkin/blob/5975eca2c172f1c889833352306413d468a4d0cc/cmake/templates/pkgConfig.cmake.in#L165). But I have never seen it being a problem / being asked on answers.ros.org.

jbohren:
What about suppressing the CMake message entirely and printing this:

``` cmake
message(SEND_ERROR "Catkin could not find a package named '${component}'.\n"
  "If the package name is correct, this problem can be resolved in one of the following ways:\n"
  "1. Install the '${component}' package to your system if it is distributed as a binary package,\n"
  "2. Add the '${component}' package to your workspace or one of your underlay workspaces,\n"
  "3. Add the installation prefix for '${component}' to CMAKE_PREFIX_PATH,\n"
  "4. Explicitly set the `${component}_DIR` variable to the directory containing the '${component}' package's CMake configuration files.")
```

dirk-thomas:
Why should it convert a fatal error in a non-fatal error? I think there is a very good reason why this case is fatal.

jbohren:
> Why should it convert a fatal error in a non-fatal error? I think there is a very good reason why this case is fatal.

You're right, it should be `FATAL_ERROR`.

dirk-thomas:
I would rather not replace the output of CMake completely. I think keeping the actual error message but providing additional information before is better.

Also I don't think the message should contain to much information. Simply because it is pretty difficult to ship changes to the message. If more information is valuable it could reference a URL.

jbohren:
> I would rather not replace the output of CMake completely. I think keeping the actual error message but providing additional information before is better.

It's up to you guys, but you're talking about novice users who appear to be confused by pretty basic CMake errors. Adding more stuff for them to read probably won't save them. In the context of a Catkin package, the CMake error is practically misinformation.

> Also I don't think the message should contain to much information. Simply because it is pretty difficult to ship changes to the message. If more information is valuable it could reference a URL.

Yeah, referencing a URL makes sense, I only suggested that since it would replace the specific (bad) advice that CMake gives when a package isn't found.

dirk-thomas:
Can anyone recommend a URL which would contain helpful information in this case? Or does it require a new wiki page or ticket to document.

wjwwood:
I don't know of a good wiki page that currently fits the bill.

dirk-thomas:
If we don't have a good page to reference I won't have the bandwidth to set one up. Therefore I would propose tho merge the PR as is since the added message provides much better context then without.

PR #782, opened 2016-03-03
use NO_MODULE to find exported catkin dependencies

dirk-thomas:
Addresses #760.

Replaces #772 since the patch will only be applied to Kinetic.


wjwwood:
+1

PR #786, opened 2016-03-18
Add format=2 style dependencies to generated package.cmake file

NikolausDemmel:
needed for https://github.com/catkin/catkin_simple/issues/7

I opened this for indigo-devel, but it should probably be ported to all distros.

It only adds new cmake variables defined in the generated `package.cmake` file, so I would assume that this change is low risk.

### DIFF
--- a/cmake/parse_package_xml.py
+++ b/cmake/parse_package_xml.py
@@ -55,8 +55,13 @@ def _get_output(package):
     values['MAINTAINER'] = '"%s"' % (', '.join([str(m) for m in package.maintainers]))
 
     values.update(_get_dependency_values('BUILD_DEPENDS', package.build_depends))
+    values.update(_get_dependency_values('BUILD_EXPORT_DEPENDS', package.build_export_depends))
     values.update(_get_dependency_values('BUILDTOOL_DEPENDS', package.buildtool_depends))
-    values.update(_get_dependency_values('RUN_DEPENDS', package.run_depends))
+    values.update(_get_dependency_values('BUILDTOOL_EXPORT_DEPENDS', package.buildtool_export_depends))
+    values.update(_get_dependency_values('EXEC_DEPENDS', package.exec_depends))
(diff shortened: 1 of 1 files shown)


dirk-thomas:
See https://github.com/catkin/catkin_simple/issues/7#issuecomment-198448067 for a detailed description about the differences and the relationship between the tags in the manifest and the dependency variables available in CMake.

For these variable to be correctly interpretable I think there must be a CMake variable identifying the format version. Otherwise e.g. the test dependencies can't be used correctly. See [REP 140](http://docs.ros.org/independent/api/rep/html/rep-0140.html#modifications-to-rep-0127):

> `test_depend` now fulfills configuration and build-time test requirements in addition to the objects needed for running them.

Also currently some of the tests fail.

NikolausDemmel:
> For these variable to be correctly interpretable It think there must be a CMake variable identifying the format version. 

I'll add that (noone actually uses TEST_DEPENDS though; I just added it for completeness).

I'll fix the unit tests as well.

dirk-thomas:
@NikolausDemmel What is your timeline on this? I could look into it this afternoon to make sure it gets released before the major rebuild of Indigo / Jade. Otherwise it is commonly several months until the next catkin release (due to the huge overhead to rebuild all packages).

NikolausDemmel:
I'm out for dinner. I'll try to do it when I get home as beers permit. It should relly be straight forward changes, so if you are about to do a release, feel free to do it yourself if you have a spare cycle.

dirk-thomas:
I will take a look right now and post the progress here. Enjoy the :beers:

dirk-thomas:
I have created #787 which has the following differences compared to this:
- add `PACKAGE_FORMAT`
- updated the unit tests
- make the generated code use deterministic order

Thank you for this patch. I reused it to start the other PR. Closing this though since it is superseded by #787.

NikolausDemmel:
Awesome. The :beers: had the better of me anyway. Thanks for addressing this!

PR #790, opened 2016-03-27
Check for exported packages in build(tool)-export depends, not run-depends.

NikolausDemmel:
- For format 1 packages this makes no difference, since build-export-depends
  are equal to run-depends and buildtool-export-depends are empty.
- For format 2 packages this has two aspects: Firstly, it now correctly fails if
  a package that is exported in `catkin_package` is listed only as an
  exec-depend, and not a build(tool)-export depend. Secondly, it now does not
  fail, if a package exported by `catkin_package` is listed only as a
  buildtool-export-depend.

### DIFF
--- a/cmake/catkin_package.cmake
+++ b/cmake/catkin_package.cmake
@@ -166,6 +166,8 @@ function(_catkin_package)
   endforeach()
 
   # for catkin packages it can be guaranteed that they are find_package()-able and have pkg-config files
+  set(_${PROJECT_NAME}_BUILD_OR_BUILDTOOL_EXPORT_DEPENDS ${${PROJECT_NAME}_BUILD_EXPORT_DEPENDS}
+                                                         ${${PROJECT_NAME}_BUILDTOOL_EXPORT_DEPENDS})
   set(PROJECT_DEPENDENCIES "")
   foreach(depend_name ${_PROJECT_CATKIN_DEPENDS})
     # verify that all catkin packages which have been find_package()-ed are listed as build dependencies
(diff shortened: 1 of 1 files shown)


NikolausDemmel:
It should be noted, that this might break the compilation of format2 packages that export a package in `catkin_package` but list it only as an `exec_depend` not `build(tool)_export_depend`. Those packages are broken and should be fixed, but they still compile with the current version of catkin.

NikolausDemmel:
_If_ we should only export build(tool)-export depends as suggested in this PR, then the documentation has a bug: It suggests to add `message_runtime` [only as exec-depend](https://github.com/ros/catkin/blame/kinetic-devel/doc/howto/format2/building_msgs.rst#L15), but then also to [export it in `catkin_package` as a CATKIN_DEPENDS](https://github.com/ros/catkin/blame/kinetic-devel/doc/howto/format2/building_msgs.rst#L78). Either we need to add it as build-export-depend, or remove it from `CATKIN_DEPENDS`. I think the latter. Opinions?

dirk-thomas:
`message_runtime` depends on e.g. `rostime` which needs to export the library with the same name. Therefore it needs to exported for downstream packages to link against. Since the library is for the target system that is a _build export_ dependency. So removing it from `CATKIN_DEPENDS` is not an option.

That library is also being used at runtime therefore it is also an _exec_ dependency.

NikolausDemmel:
So the documentation should be changed to either 

```
<depend>message_runtime</depend>
```

or 

```
<build_export_depend>message_runtime</build_export_depend>
<exec_depend>message_runtime</exec_depend>
```

Any preferences?

I have a very slight preference for the former because of simplicity. The extra build-depend is superfluous (?), but shouldn't hurt much I guess. But either way is fine IMO.

dirk-thomas:
The documentation shouldn't recommend to add `message_runtime` as a build dependency. Therefore option 2.

NikolausDemmel:
> It should be noted, that this might break the compilation of format2 packages that export a package in catkin_package but list it only as an exec_depend not build(tool)_export_depend. Those packages are broken and should be fixed, but they still compile with the current version of catkin.

I think we will have to go with a mere warning in such cases to not break builds, at least for a transition period.

dirk-thomas:
Closing due to inactivity.

PR #798, opened 2016-04-23
Support setup.py entry_points

wkentaro:
### Related Issues
- roslint new verbs in catkin_tools (**original issue**) https://github.com/catkin/catkin_tools/pull/241
- add roslint verb as entry_points https://github.com/ros/roslint/pull/51

### DIFF
--- a/cmake/interrogate_setup_dot_py.py
+++ b/cmake/interrogate_setup_dot_py.py
@@ -160,6 +160,8 @@ def _create_mock_setup_function(package_name, outfile, install_dir, script_dir):
 
     :param package_name: name of the package
     :param outfile: filename that cmake will use afterwards
+    :param install_dir: install directory for python package
+    :param script_dir: install directory for python scripts
     :returns: a function to replace disutils.core.setup and setuptools.setup
     """
 
--- a/cmake/interrogate_setup_dot_py.py
+++ b/cmake/interrogate_setup_dot_py.py
@@ -152,7 +152,7 @@ def generate_cmake_file(package_name, version, scripts, package_dir, pkgs, modul
     return result
 
 
-def _create_mock_setup_function(package_name, outfile, install_dir):
+def _create_mock_setup_function(package_name, outfile, install_dir, script_dir):
     """
     Creates a function to call instead of distutils.core.setup or
     setuptools.setup, which just captures some args and writes them
@@ -227,9 +227,23 @@ def _create_mock_setup_function(package_name, outfile, install_dir):
                 self.egg_base = install_dir
                 self.egg_info = os.path.join(
                     self.egg_base, '%s-%s.egg-info' % (package_name, version))
-        cmd = mock_egg_info(dist)
-        cmd.finalize_options()
-        cmd.run()
+        cmd_egg_info = mock_egg_info(dist)
+        cmd_egg_info.finalize_options()
+        cmd_egg_info.run()
+
+        # easy_install command
+        import pkg_resources
+        from setuptools.command.easy_install import easy_install
+        class mock_easy_install(easy_install):
+            def finalize_options(self):
+                self.script_dir = script_dir
+                self.outputs = []
+        metadata = pkg_resources.PathMetadata(cmd_egg_info.egg_base, cmd_egg_info.egg_info)
+        dist_pkg_resources = pkg_resources.Distribution(install_dir, project_name=package_name, metadata=metadata)
+        cmd_easy_install = mock_easy_install(dist)
+        cmd_easy_install.finalize_options()
+        # TODO: support cmd.run()
+        cmd_easy_install.install_wrapper_scripts(dist_pkg_resources)
 
     return setup
 
@@ -243,6 +257,7 @@ def main():
     parser.add_argument('setupfile_path', help='Full path to setup.py')
     parser.add_argument('outfile', help='Where to write result to')
     parser.add_argument('install_dir', help='Python install directory')
+    parser.add_argument('script_dir', help='Python scripts directory')
 
     args = parser.parse_args()
 
@@ -264,7 +279,8 @@ def main():
     try:
         fake_setup = _create_mock_setup_function(package_name=args.package_name,
                                                  outfile=args.outfile,
-                                                 install_dir=args.install_dir)
+                                                 install_dir=args.install_dir,
+                                                 script_dir=args.script_dir)
 
         distutils_backup = distutils.core.setup
         distutils.core.setup = fake_setup
--- a/cmake/interrogate_setup_dot_py.py
+++ b/cmake/interrogate_setup_dot_py.py
@@ -37,11 +37,7 @@ import os
 import sys
 
 import distutils.core
-try:
-    import setuptools
-    SETUPTOOLS = True
-except ImportError:
-    SETUPTOOLS = False
+import setuptools
 
 from argparse import ArgumentParser
 
@@ -201,10 +197,6 @@ def _create_mock_setup_function(package_name, outfile, install_dir, script_dir):
                                      modules=modules)
         with open(outfile, 'w') as out:
             out.write('\n'.join(result))
-
-        # setuptools is required below
-        if not SETUPTOOLS:
-            return
 
         attrs = dict(
             name=package_name,
--- a/cmake/interrogate_setup_dot_py.py
+++ b/cmake/interrogate_setup_dot_py.py
@@ -208,7 +208,7 @@ def _create_mock_setup_function(package_name, outfile, install_dir, script_dir):
             entry_points=entry_points,
         )
 
-        from distutils.dist import Distribution
+        from setuptools.dist import Distribution
         dist = Distribution(attrs)
         dist.script_name = 'setup.py'
 
--- a/cmake/interrogate_setup_dot_py.py
+++ b/cmake/interrogate_setup_dot_py.py
@@ -92,7 +92,7 @@ def generate_cmake_file(package_name, version, scripts, package_dir, pkgs, modul
     :param package_dir: {modulename: path}
     :pkgs: [list of str] python_packages declared in catkin package
     :modules: [list of str] python modules
-    :entry_points: {dict of str and list of str} entry points
+    :entry_points: {dict of key=str, value=[list of str]} entry points
     """
     prefix = '%s_SETUP_PY' % package_name
     result = []
--- a/cmake/interrogate_setup_dot_py.py
+++ b/cmake/interrogate_setup_dot_py.py
@@ -39,8 +39,9 @@ import sys
 import distutils.core
 try:
     import setuptools
+    SETUPTOOLS = True
 except ImportError:
-    pass
+    SETUPTOOLS = False
 
 from argparse import ArgumentParser
 
@@ -83,7 +84,7 @@ def _get_locations(pkgs, package_dir):
     return locations
 
 
-def generate_cmake_file(package_name, version, scripts, package_dir, pkgs, modules, entry_points):
+def generate_cmake_file(package_name, version, scripts, package_dir, pkgs, modules):
     """
     Generates lines to add to a cmake file which will set variables
 
@@ -92,7 +93,6 @@ def generate_cmake_file(package_name, version, scripts, package_dir, pkgs, modul
     :param package_dir: {modulename: path}
     :pkgs: [list of str] python_packages declared in catkin package
     :modules: [list of str] python modules
-    :entry_points: {dict of key=str, value=[list of str]} entry points
     """
     prefix = '%s_SETUP_PY' % package_name
     result = []
@@ -149,16 +149,10 @@ def generate_cmake_file(package_name, version, scripts, package_dir, pkgs, modul
     result.append(r'set(%s_MODULES "%s")' % (prefix, ';'.join(['%s.py' % m.replace('.', '/') for m in filtered_modules])))
     result.append(r'set(%s_MODULE_DIRS "%s")' % (prefix, ';'.join([module_locations[m] for m in filtered_modules]).replace("\\", "/")))
 
-    entry_points_contents = []
-    for name, entries in entry_points.items():
-        entry_points_contents.append('[%s]\n%s\n\n' % (name, '\n'.join(entries)))
-    entry_points_contents = '\n'.join(entry_points_contents)
-    result.append(r'set(%s_ENTRY_POINTS_CONTENTS "%s")' % (prefix, entry_points_contents))
-
     return result
 
 
-def _create_mock_setup_function(package_name, outfile):
+def _create_mock_setup_function(package_name, outfile, install_dir):
     """
     Creates a function to call instead of distutils.core.setup or
     setuptools.setup, which just captures some args and writes them
@@ -202,10 +196,40 @@ def _create_mock_setup_function(package_name, outfile):
                                      scripts=scripts,
                                      package_dir=package_dir,
                                      pkgs=pkgs,
-                                     modules=modules,
-                                     entry_points=entry_points)
+                                     modules=modules)
         with open(outfile, 'w') as out:
             out.write('\n'.join(result))
+
+        # setuptools is required below
+        if not SETUPTOOLS:
+            return
+
+        attrs = dict(
+            name=package_name,
+            version=version,
+            package_dir=package_dir,
+            packages=pkgs,
+            scripts=scripts,
+            py_modules=modules,
+            entry_points=entry_points,
+        )
+
+        from distutils.dist import Distribution
+        dist = Distribution(attrs)
+        dist.script_name = 'setup.py'
+
+        # egg_info command
+        from setuptools.command.egg_info import egg_info
+        class mock_egg_info(egg_info):
+            def finalize_options(self):
+                self.egg_name = package_name
+                self.egg_version = version
+                self.egg_base = install_dir
+                self.egg_info = os.path.join(
+                    self.egg_base, '%s-%s.egg-info' % (package_name, version))
+        cmd = mock_egg_info(dist)
+        cmd.finalize_options()
+        cmd.run()
 
     return setup
 
@@ -218,6 +242,7 @@ def main():
     parser.add_argument('package_name', help='Name of catkin package')
     parser.add_argument('setupfile_path', help='Full path to setup.py')
     parser.add_argument('outfile', help='Where to write result to')
+    parser.add_argument('install_dir', help='Python install directory')
 
     args = parser.parse_args()
 
@@ -238,7 +263,8 @@ def main():
     # context of evaluating setup.py
     try:
         fake_setup = _create_mock_setup_function(package_name=args.package_name,
-                                                outfile=args.outfile)
+                                                 outfile=args.outfile,
+                                                 install_dir=args.install_dir)
 
         distutils_backup = distutils.core.setup
         distutils.core.setup = fake_setup
--- a/cmake/interrogate_setup_dot_py.py
+++ b/cmake/interrogate_setup_dot_py.py
@@ -83,7 +83,7 @@ def _get_locations(pkgs, package_dir):
     return locations
 
 
-def generate_cmake_file(package_name, version, scripts, package_dir, pkgs, modules):
+def generate_cmake_file(package_name, version, scripts, package_dir, pkgs, modules, entry_points):
     """
     Generates lines to add to a cmake file which will set variables
 
@@ -92,6 +92,7 @@ def generate_cmake_file(package_name, version, scripts, package_dir, pkgs, modul
     :param package_dir: {modulename: path}
     :pkgs: [list of str] python_packages declared in catkin package
     :modules: [list of str] python modules
+    :entry_points: {dict of str and list of str} entry points
     """
     prefix = '%s_SETUP_PY' % package_name
     result = []
@@ -148,6 +149,12 @@ def generate_cmake_file(package_name, version, scripts, package_dir, pkgs, modul
     result.append(r'set(%s_MODULES "%s")' % (prefix, ';'.join(['%s.py' % m.replace('.', '/') for m in filtered_modules])))
     result.append(r'set(%s_MODULE_DIRS "%s")' % (prefix, ';'.join([module_locations[m] for m in filtered_modules]).replace("\\", "/")))
 
+    entry_points_contents = []
+    for name, entries in entry_points.items():
+        entry_points_contents.append('[%s]\n%s\n\n' % (name, '\n'.join(entries)))
+    entry_points_contents = '\n'.join(entry_points_contents)
+    result.append(r'set(%s_ENTRY_POINTS_CONTENTS "%s")' % (prefix, entry_points_contents))
+
     return result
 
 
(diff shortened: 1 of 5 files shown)


wkentaro:
### Sample: Add new verb roslint as entry_points for catkin_tools

#### Needed PRs
- This PR
- https://github.com/ros/roslint/pull/51

#### Demo

```
$ cat /home/wkentaro/ros/indigo/build/roslint/catkin_generated/setup_py_interrogation.cmake
set(roslint_SETUP_PY_VERSION "0.11.0")
set(roslint_SETUP_PY_SCRIPTS "")
set(roslint_SETUP_PY_PACKAGES "roslint")
set(roslint_SETUP_PY_PACKAGE_DIRS "src/roslint")
set(roslint_SETUP_PY_MODULES "")
set(roslint_SETUP_PY_MODULE_DIRS "")
set(roslint_SETUP_PY_ENTRY_POINTS_CONTENT "[catkin_tools.commands.catkin.verbs]
roslint = roslint.main:description")
```

```
$ cat ~/ros/indigo/devel/lib/python2.7/dist-packages/roslint-0.11.0.egg-info/entry_points.txt
[catkin_tools.commands.catkin.verbs]
roslint = roslint.main:description
```

```
$ catkin --help
usage: catkin [-h] [-a] [--test-colors] [--force-color | --no-color]
              [build | clean | config | create | init | lint | list | locate | profile | roslint]
              ...

catkin command

optional arguments:
  -h, --help            show this help message and exit
  -a, --list-aliases    Lists the current verb aliases and then quits, all
                        other arguments are ignored
  --test-colors         Prints a color test pattern to the screen and then
                        quits, all other arguments are ignored
  --force-color         Forces catkin to output in color, even when the
                        terminal does not appear to support it.
  --no-color            Forces catkin to not use color in the output,
                        regardless of the detect terminal type.

catkin command:
  Call `catkin VERB -h` for help on each verb listed below:

    build       Builds a catkin workspace.
    clean       Deletes various products of the build verb.
    config      Configures a catkin workspace's context.
    create      Creates catkin workspace resources like packages.
    init        Initializes a given folder as a catkin workspace.
    lint        Checks catkin packages for common errors
    list        Lists catkin packages in the workspace or other arbitray folders.
    locate      Get the paths to various locations in a workspace.
    profile     Manage config profiles for a catkin workspace.
    roslint     Run roslint make target

  [build | clean | config | create | init | lint | list | locate | profile | roslint]
```

cmake/interrogate_setup_dot_py.py
```diff
@@ -92,6 +92,7 @@ def generate_cmake_file(package_name, version, scripts, package_dir, pkgs, modul
     :param package_dir: {modulename: path}
     :pkgs: [list of str] python_packages declared in catkin package
     :modules: [list of str] python modules
+    :entry_points: {dict of str and list of str} entry points
```
dirk-thomas: From the code it looks like this argument must be a `dict`?

dirk-thomas:
Can you please describe and provide an example which I can run to reproduce the new feature.

wkentaro:
You can build a workspace with below rosinstall file to reproduce the feature.

```
- git:
    local-name: catkin
    uri: https://github.com/wkentaro/catkin.git
    version: setup_py-entry-points
- git:
    local-name: roslint
    uri: https://github.com/wkentaro/roslint.git
    version: catkin_tools-entry-points
```

cmake/interrogate_setup_dot_py.py
wkentaro: Ok, fixed.

dirk-thomas:
When I build the workspace I can't find the executable for the entry point anywhere. Is it being generated for you?

wkentaro:
In this patch, I just creates the entry_points.txt file. 
So `console_scripts` is not supported at present, I will do search about the feature in original python packaging.

wkentaro:
The last commit is to reuse egg-info command in setuptools.
This makes the process more complex, but it enables `console_script` as https://github.com/ros/catkin/pull/799.

wkentaro:
Which one do you prefer?
- generate cmake file to install entry_points (generating only `entry_points.txt`)
- reuse egg_info command to extend to setup `console_scripts` with easy_install command (as shown in #799)

dirk-thomas:
To take step before discussing the details of the change. For any of this to be considered for merging it needs to:
- demonstrate the new feature (an example I can easily run, currently I don't see anything for console_scripts) (does it do anything for the install space?)
- pass all tests (currently it fails)
- not break existing behavior (the patch creates the egg folder in the devel space, will this have any side effects?)

wkentaro:
### Demonstration

catkin_tools supports extending new subcommands with `entry_points`: https://github.com/catkin/catkin_tools/blob/master/setup.py#L141 .
Below is to demonstrate installing `entry_points` for `roslint` package to get `catkin roslint` verb.

```
$ catkin --help | grep roslint

$ mkdir -p /tmp/test_ws/src
$ cd /tmp/test_ws/src
$ cat <<-EOF >> .rosinstall
- git:
    local-name: catkin
    uri: https://github.com/wkentaro/catkin.git
    version: setup_py-entry-points
- git:
    local-name: roslint
    uri: https://github.com/wkentaro/roslint.git
    version: catkin_tools-entry-points
EOF
$ wstool update

$ source /opt/ros/indigo/setup.bash
$ cd /tmp/test_ws
$ catkin build -iv
$ source devel/setup.bash

$ ls devel/lib/python2.7/dist-packages/roslint-0.11.0.egg-info
dependency_links.txt  entry_points.txt  PKG-INFO  SOURCES.txt  top_level.txt

$ cat devel/lib/python2.7/dist-packages/roslint-0.11.0.egg-info/entry_points.txt 
[catkin_tools.commands.catkin.verbs]
roslint = roslint.main:description

$ catkin --help | grep roslint
              [build | clean | config | create | init | lint | list | locate | profile | roslint]
    roslint Run roslint make target
  [build | clean | config | create | init | lint | list | locate | profile | roslint]
```

wkentaro:
> demonstrate the new feature (an example I can easily run, currently I don't see anything for console_scripts) (does it do anything for the install space?)

`console_scripts` entry points for installing executable is not supported in this PR, so that's why you don't see any executable in this PR.

FYI, this commit supports `console_scripts` to install executables. https://github.com/ros/catkin/pull/799/commits/b454271f2116e5e8e93633af819fc6b4978799d2

wkentaro:
> pass all tests (currently it fails)

As shown in https://github.com/ros/catkin/pull/798/commits , removing last commit makes this PR pass the CI.
So I'd like to discuss about whether we should reuse egg_info command. (Making test for the last commit takes time a little, so I'd like to discuss before fixing it)

wkentaro:
> not break existing behavior (the patch creates the egg folder in the devel space, will this have any side effects?)

I don't think this breaks existing behavior, because putting `*.egg-info` directory  in python library directory seems natural as below:

```
% find /usr/local/lib/python2.7/dist-packages/ -type d -iname "*.egg-info" 
/usr/local/lib/python2.7/dist-packages/slacker_cli-0.2.7-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/simplegeneric-0.8.1-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/scikit_learn-0.17.1-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/trollius-2.1-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/dlib-18.17.100-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/docutils-0.12-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/wstool_cd-0.16-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/catkin_tools-0.4.2-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/args-0.1.0-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/clint-0.5.1-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/osrf_pycommon-0.1.2-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/vcstools-0.1.38-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/cmigemo-0.1.6.egg-info
/usr/local/lib/python2.7/dist-packages/libpgm-1.3-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/catkin_pkg-0.2.10-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/pycd-0.3.28-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/gshell-2.2.4-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/ipdb-0.9.0-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/port_for-0.3.1-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/gdown-1.0.2-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/Cython-0.23.4-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/pygithub3-0.5.1-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/PyYAML-3.11-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/percol-0.2.1.egg-info
/usr/local/lib/python2.7/dist-packages/pexpect-4.0.1-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/wstool-0.1.12-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/CommonMark-0.5.4-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/slacker-0.9.2-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/tabulate-0.7.5-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/catkin_tools-0.3.1-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/toolz-0.7.4-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/gotenshita-2.0.2-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/h5py-2.5.0-py2.7.egg-info
/usr/local/lib/python2.7/dist-packages/fysom-2.1.2-py2.7.egg-info
```

NikolausDemmel:
I didn't look at the implementation yet, but wanted to quickly say that this feature would be useful for me! Thanks!

dirk-thomas:
The feature would be great indeed.

This PRs claims to support `entry_points` in the `setup.py` file. But if one of the keys is `console_scripts` it doesn't work as expected. Therefore this PR should explicitly check if console scripts are used and warn about that it is not supported. Since #799 already adds support for console scripts I will only refer to that.

The following things are missing (as mentioned before):
- There is no example which demonstrates that `console_scripts` work with #799. I could setup one myself but I expect it to be provided with the PR to be runnable without effort (and uncertainty if everyone tried the same).
- After sourcing the install space the entry points are not working. `roslint` does not show up as a verb in catkin tools.

And yes, the tests can be addressed once the patch for the feature is working - it doesn't have to be right now.

wkentaro:
I fixed for the install space in dcd52e0 .

wkentaro:
For console_scripts, please apply below patch in roslint package (cloned as above from my fork)
and try `foo` executable.

```
diff --git a/setup.py b/setup.py
index ff5b1f0..2b357df 100644
--- a/setup.py
+++ b/setup.py
@@ -7,6 +7,9 @@ setup(**generate_distutils_setup(
         '': 'src'
     },
     entry_points={
+        'console_scripts': [
+            'foo = roslint.foo:main',
+        ],
         'catkin_tools.commands.catkin.verbs': [
             'roslint = roslint.main:description',
         ],
diff --git a/src/roslint/foo.py b/src/roslint/foo.py
new file mode 100644
index 0000000..eeb70fc
--- /dev/null
+++ b/src/roslint/foo.py
@@ -0,0 +1,6 @@
+#!/usr/bin/env python
+
+def main():
+    print('>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>')
+    print('foo')
+    print('<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<')
```

dirk-thomas:
`foo` is still not available to run after sourcing the devel space. For the install space it works.

setup.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-from distutils.core import setup
+from setuptools import setup
```
dirk-thomas: You need to declare this dependency in the manifest. Otherwise it is not available on the build farm.

dirk-thomas:
Please update the tests to pass.

wkentaro:
Fixed.

dirk-thomas:
The patch drops a `dist` folder as well as a `<pkgname>.egg-info` folder into the source space of the built package. But it shouldn't change the source space at all.

Also the test does not call the `setup` function with `entry_points` and therefore doesn't cover the new feature.

Before this patch the `_create_mock_setup_function` when being called by the package only returns all relevant information but doesn't do anything yet. The information is then dumped into CMake which later gets executed. With this patch the function also executes `distutils` / `setuptools` internal functions. I am not sure if that is the good idea. It would be better if the function would just return the collected information and the actual action is happening later (when the CMake is being executed).

test/unit_tests/test_interrogate_setup.py
```diff
@@ -111,10 +111,34 @@ def test_generate_cmake_file_invalid(self):
...
+        os.makedirs(os.path.join(workdir, 'bar'))
+        open(os.path.join(workdir, 'bar/__init__.py'), 'w')
+        os.makedirs(os.path.join(workdir, 'bar/nodes'))
+        open(os.path.join(workdir, 'bar/nodes/bar'), 'w')
+        # src/__init__.py, lib/__init__.py
+        open(os.path.join(workdir, 'src/__init__.py'), 'w')
+        os.makedirs(os.path.join(workdir, 'lib'))
+        open(os.path.join(workdir, 'lib/__init__.py'), 'w')
+        # change directory for testing
+        os.chdir(workdir)
```
dirk-thomas: I don't understand what all of this is being used for.

wkentaro: They are necessary to test various combinations of `package_dir` in below lines.

like:

```
fake_setup(version='0.1.1', package_dir={}, packages=['foo', 'bar'], scripts=['bin/foo', 'nodes/bar'])
fake_setup(version='0.1.1', package_dir={'foo': 'src', 'bar': 'lib'}, packages=['foo', 'bar'],)
```

wkentaro:
So one possible solution is to generate cmake and setup.py in build space, and runs the setup.py inside the cmake?

dirk-thomas:
In [ament_tools](https://github.com/ament/ament_tools/blob/24a11a523cf900ce3880c62ed79127cc2177a4a4/ament_tools/build_types/ament_python.py#L151-L189) I pass various arguments to the setup file in order to achieve that it doesn't generate anything in the source space. Maybe that can help somehow.

wkentaro:
@dirk-thomas Sorry for late reply.
I re-built the [demo environment](https://github.com/ros/catkin/pull/798#issuecomment-215138012), and found that it does not leave anything in the source space.
Could you please review this again? I'm still interested in and demanding this feature.

```
% wstool info
workspace: /tmp/catkin_ws/src

 Localname S SCM Version (Spec)            UID  (Spec)  URI  (Spec) [http(s)://...]
 --------- - --- --------------            -----------  ---------------------------
 roslint     git catkin_tools-entry-points 072bcbc1e7aa github.com/wkentaro/roslint.git
 catkin      git setup_py-entry-points     7427c6790ac5 github.com/wkentaro/catkin.git

% wstool foreach 'git clean -fxdn'
[catkin] Would remove python/catkin/__init__.pyc
[catkin] Would remove python/catkin/environment_cache.pyc
[roslint] Would remove src/roslint/main.pyc
```

cmake/interrogate_setup_dot_py.py
```diff
@@ -199,6 +198,47 @@ def setup(*args, **kwargs):
...
+            name=package_name,
+            version=version,
+            package_dir=package_dir,
+            packages=pkgs,
+            scripts=scripts,
+            py_modules=modules,
+            entry_points=entry_points,
+        )
+
+        from distutils.dist import Distribution
```
dirk-thomas: Why are you using `distutils.dist.Distribution` here and not `setuptools.dist.Distribution`?

cmake/interrogate_setup_dot_py.py
```diff
@@ -199,6 +198,47 @@ def setup(*args, **kwargs):
...
+
+        # egg_info command
+        from setuptools.command.egg_info import egg_info
+        class mock_egg_info(egg_info):
+            def finalize_options(self):
+                self.egg_name = package_name
+                self.egg_version = version
+                self.egg_base = install_dir
+                self.egg_info = os.path.join(
+                    self.egg_base, '%s-%s.egg-info' % (package_name, version))
```
dirk-thomas: This seems to be pretty fragile to me. The custom function doesn't call the super function. Is this called from anywhere other than two lines below? If not, why is it overriding this method? Couldn't you perform the logic two lines below external of the `egg_info`?

cmake/interrogate_setup_dot_py.py
```diff
@@ -199,6 +198,47 @@ def setup(*args, **kwargs):
...
+        cmd_egg_info.finalize_options()
+        cmd_egg_info.run()
+
+        # easy_install command
+        import pkg_resources
+        from setuptools.command.easy_install import easy_install
+        class mock_easy_install(easy_install):
+            def finalize_options(self):
+                self.script_dir = script_dir
+                self.outputs = []
```
dirk-thomas: Same comment / question / concern / suggestion as for the `egg_info` above.

cmake/interrogate_setup_dot_py.py
```diff
@@ -199,6 +198,47 @@ def setup(*args, **kwargs):
...
+                self.egg_version = version
+                self.egg_base = install_dir
+                self.egg_info = os.path.join(
+                    self.egg_base, '%s-%s.egg-info' % (package_name, version))
+        cmd_egg_info = mock_egg_info(dist)
+        cmd_egg_info.finalize_options()
+        cmd_egg_info.run()
+
+        # easy_install command
+        import pkg_resources
```
dirk-thomas, line 229: Doesn't this need to be declared as a dependency?

setup.py
dirk-thomas: Do other ROS packages which use `from distutils.core import setup` have to be updated in order to continue to work?

test/unit_tests/test_interrogate_setup.py
dirk-thomas: None of these files are being cleaned up afterwards. Please see how `rootdir` below is created and always cleaned up.

This code block also keeps a lot of file handles open much longer than necessary. Maybe the easiest way would be to create a function which makes sure the folder exists and than creates an empty file. That will also make this block much more readable.

dirk-thomas:
I am sorry that I didn't have time earlier. Please see another few comments / questions above.

dirk-thomas:
We are aiming for a new round of releases early next week (Mon or Tue). It would be great f this would be ready before in order to get released.

cmake/interrogate_setup_dot_py.py
wkentaro: Both works.

dirk-thomas: Since the rest of the code is using `setuptools` I would suggest to use the same package here.

cmake/interrogate_setup_dot_py.py
wkentaro: 114b3d4 + https://github.com/ros/rosdistro/pull/15724 fix the issue.

wkentaro:
> Do other ROS packages which use from distutils.core import setup have to be updated in order to continue to work?

I'm not sure why I changed like that a year ago, but currently it is not needed. e131753 will fix it.

cmake/interrogate_setup_dot_py.py
wkentaro: 31adac9 will fix this.

wkentaro:
> None of these files are being cleaned up afterwards. Please see how rootdir below is created and always cleaned up.
>
> This code block also keeps a lot of file handles open much longer than necessary. Maybe the easiest way would be to create a function which makes sure the folder exists and than creates an empty file. That will also make this block much more readable.

db4352d will fix the issue.

wkentaro:
@dirk-thomas I think I've addressed to all of your comments, so could you please review again?

> We are aiming for a new round of releases early next week (Mon or Tue). It would be great f this would be ready before in order to get released.

Sorry, I was preparing for a robotics challenge during the final week in July, and couldn't tackle this issue.

test/unit_tests/test_interrogate_setup.py
```diff
@@ -153,4 +169,14 @@ def test_interrogate_setup_py(self):
 set(foo_SETUP_PY_MODULE_DIRS "")""", contents)
             os.remove(outfile)
         finally:
-            shutil.rmtree(rootdir)
+            shutil.rmtree(develdir)
+            shutil.rmtree(srcdir)
+            os.chdir(curdir)
```
dirk-thomas, line 174: The finally block should first change the cwd and then try to remove the other folders (otherwise on e.g. Windows the cwd will prevent deleting the cwd `srcdir`).

test/unit_tests/test_interrogate_setup.py
```diff
@@ -153,4 +169,14 @@ def test_interrogate_setup_py(self):
...
+
+
+def _create_empty_files(filenames):
+    for fname in filenames:
+        directory = os.path.dirname(fname)
+        if not os.path.exists(directory):
+            os.makedirs(directory)
+        open(fname, 'w')
```
dirk-thomas, line 182: Please close the file handle, e.g.:

```
with open(fname, 'w'):
    pass
````

File: test/unit_tests/test_interrogate_setup.py [hunk 10]
```diff
@@ -153,4 +169,14 @@ def test_interrogate_setup_py(self):
...
+            shutil.rmtree(develdir)
+            shutil.rmtree(srcdir)
+            os.chdir(curdir)
+
+
+def _create_empty_files(filenames):
+    for fname in filenames:
+        directory = os.path.dirname(fname)
+        if not os.path.exists(directory):
+            os.makedirs(directory)
```
dirk-thomas, line 181: Rather catch the exception than test for existence before since this is still subject to a race.

cmake/interrogate_setup_dot_py.py
```diff
@@ -199,6 +198,44 @@ def setup(*args, **kwargs):
...
+        from setuptools.command.easy_install import easy_install
+        cmd_easy_install = easy_install(dist)
+        # overwrite script_dir and not call finalize_options method
+        # to avoid many assertions of easy_install command like check_site_dir
+        cmd_easy_install.script_dir = script_dir
+        metadata = pkg_resources.PathMetadata(cmd_egg_info.egg_base, cmd_egg_info.egg_info)
+        dist_pkg_resources = pkg_resources.Distribution(install_dir, project_name=package_name, metadata=metadata)
+        cmd_easy_install.install_wrapper_scripts(dist_pkg_resources)
```
dirk-thomas, line 237: If a package doesn't use `entry_points` what side effects does this block (running `egg_info` and `easy_install`) have? Could it be made conditional to reduce the chance of regressions - only packages which start using entry points would exercise this logic?

dirk-thomas:
Closing due to inactivity.

PR #799, opened 2016-04-27
Support console_scripts in entry_points

wkentaro:


wkentaro:
This will be supported in #798

Issue #807, opened 2016-06-13
setup.bat does not handle spaces in paths in Windows


jmdaly:
When building catkin for Windows (running in Windows 7 64-bit), the default `setup.bat` file that is generated does not handle spaces in the path correctly. Since the default installation location for catkin is in `C:\Program File (x86)\catkin`, the script does not work correctly by default. When running `setup.bat` on my system, I get the following output,

```
C:\Program Files (x86)\catkin>setup.bat
'Files' is not recognized as an internal or external command,
operable program or batch file.
"Missing Python script: "C:/Program Files (x86)/catkin/_setup_util.py""
```

This is due to the following line in the generated `setup.bat`:

```
set _SETUP_UTIL="C:/Program Files (x86)/catkin/_setup_util.py"
```

The script believes that the "Files" part of "Program Files" is a separate argument.

To fix this, a couple of small changes need to be make to `cmake/templates/setup.bat.in`. I have forked and made the fix, and will submit a pull request shortly.

dirk-thomas:
Fixed by #808. Thanks.

PR #808, opened 2016-06-13
fix quoting of paths to handle spaces

jmdaly:
This pull request fixes issue #807. It ensures that variables containing paths with spaces are properly handled.


dirk-thomas:
Thank you for the patch.

PR #810, opened 2016-06-20
[indigo] improve cross-compile behaviour

dominiquehunziker:
Fix from openembedded's [meta-ros](https://github.com/bmwcarit/meta-ros/pull/386) layer to generate pkgConfig.cmake files which can be used also on the target platform to compile packages.

When cross-compiling the pkgConfig.cmake contains absolute paths from the host platform which makes it impossible to use the pkgConfig.cmake files to compile packages in the target platform. This PR addresses this by stripping the `CMAKE_FIND_ROOT_PATH` prefixes from the relevant variables before they are used to generate the pkgConfig.cmake file. In a second step CMake's `find_*` functions are used in the pkgConfig.cmake file to handle `CMAKE_FIND_ROOT_PATH` correctly when locating the previously stripped variables.


dirk-thomas:
Please create the pull request against the branch `kinetic-devel`. A significant change like this won't be merged into Indigo immediately (since it has been released over two years ago and aims for stability). It should first go into Kinetic and if it proves to not have an regressions it can be considered for backporting.

dominiquehunziker:
I added an new PR for kinetic [here](https://github.com/ros/catkin/pull/811). I guess this one is going to be closed?

dirk-thomas:
Yes, closing in favor of #811. If the patch gets backported that likely by cherry-picking the relevant changes.

PR #816, opened 2016-07-25
Use nmake

nizar-sallem:
@dirk-thomas hope you are doing well.
This is just an enhancement that would allow handling nmake same way as ninja.
Did only test with indigo but can compare with latest development and update my PR

### DIFF
--- a/python/catkin/builder.py
+++ b/python/catkin/builder.py
@@ -341,7 +341,7 @@ def build_catkin_package(
     path, package,
     workspace, buildspace, develspace, installspace,
     install, force_cmake, quiet, last_env, cmake_args, make_args,
-    destdir=None, use_ninja=False
+    destdir=None, use_ninja=False, use_nmake=False
 ):
     cprint(
         "Processing @{cf}catkin@| package: '@!@{bf}" +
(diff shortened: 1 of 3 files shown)


python/catkin/builder.py
```diff
@@ -403,10 +403,13 @@ def build_catkin_package(
     else:
         print('%s exists, skipping explicit cmake invocation...' % makefile_name)
         # Check to see if cmake needs to be run via make
-        if not use_ninja:
-            make_check_cmake_cmd = ['make', 'cmake_check_build_system']
-        else:
+        if use_ninja:
             make_check_cmake_cmd = ['ninja', 'build.ninja']
+        elif use_namke:
```
dirk-thomas: Spelling error fails CI.

dirk-thomas:
Thanks for the pull request - good to see you continue to work with ROS.

Once this passes CI I would merge it into kinetic-devel first. It can also be backported into Indigo / Jade if you need that since the risk of regressions seems rather low to me.

python/catkin/builder.py
nizar-sallem: Fixing

dirk-thomas:
@ros-pull-request-builder retest this please

dirk-thomas:
I squashed and cherry-picked the patch to the _kinetic-devel_ branch: d0c4ecb3bac9b38a0883253e1e350d3fb72c4a6f

nizar-sallem:
Awesome, thanks

python/catkin/builder.py
```diff
@@ -520,9 +530,11 @@ def build_cmake_package(
 
     # Run make
     if not use_ninja:
-        make_executable = 'make'
-    else:
         make_executable = 'ninja'
```
gavanderhoorn, line 533: Is this condition inverted? It looks like it now says:

``` python
if not use_ninja:
    make_executable = 'ninja'
```

That would seem to be the inverse of what is actually the intent.

dirk-thomas: For reference: #826

PR #820, opened 2016-08-06
maintain file context when invoking Python scripts through relay

dirk-thomas:
Without the patch the stacktrace only show `<string>` for the file the relay scripts are redirecting to.

With this patch the stacktrace contains the correct file path of the Python script.


wjwwood:
+1

PR #832, opened 2016-10-21
updates for #831

wjwwood:
See: https://github.com/ros/catkin/pull/831/files#r84355617

### DIFF
--- a/cmake/python.cmake
+++ b/cmake/python.cmake
@@ -15,6 +15,28 @@ if(WIN32)
   file(TO_NATIVE_PATH "${CMAKE_INSTALL_PREFIX}" SETUPTOOLS_INSTALL_PREFIX)
 endif()
 
+# a precomputed PYTHON_INSTALL_DIR can be passed along with the matching PYTHON_EXECUTABLE here
+# in order to avoid needing to call PYTHON_EXECUTABLE in the next block in each package
+# the precomputed value is only taken if the PYTHON_EXECUTABLE matches what was found above
+if(DEFINED PYTHON_INSTALL_DIR_TUPLE)
+  string(REPLACE ":" ";" _tuple ${PYTHON_INSTALL_DIR_TUPLE})
+  list(LENGTH _tuple _tuple_len)
+  if(_tuple_len EQUAL 2)
+    list(GET _tuple 0 _tuple_PYTHON_EXECUTABLE)
+    list(GET _tuple 1 _tuple_PYTHON_INSTALL_DIR)
+    if("${_tuple_PYTHON_EXECUTABLE}" STREQUAL "${PYTHON_EXECUTABLE}")
+      set(PYTHON_INSTALL_DIR "${_tuple_PYTHON_INSTALL_DIR}")
+    else()
+      message(WARNING
+        "PYTHON_INSTALL_DIR_TUPLE is set, "
+        "but the given PYTHON_EXECUTABLE '${_tuple_PYTHON_EXECUTABLE}' does not "
+        "match the PYTHON_EXECUTABLE found by CMake '${PYTHON_EXECUTABLE}'.\n"
+        "The precomputed value in the tuple will be ignored and Python will be "
+        "invoked to get the correct PYTHON_INSTALL_DIR.")
+    endif()
+  endif()
+endif()
+
 # this block determines the same value as the Python function get_python_install_dir() from python/catkin/builder.py
 if(NOT DEFINED PYTHON_INSTALL_DIR)
   execute_process(COMMAND "${PYTHON_EXECUTABLE}"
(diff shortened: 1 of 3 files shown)


dirk-thomas:
The tests are failing with Python 3 at the moment.

wjwwood:
Yeah, I was looking into them.

wjwwood:
Looks like that fixed it.

cmake/python.cmake
```diff
@@ -15,6 +15,28 @@ if(WIN32)
   file(TO_NATIVE_PATH "${CMAKE_INSTALL_PREFIX}" SETUPTOOLS_INSTALL_PREFIX)
 endif()
 
+# a precomputed PYTHON_INSTALL_DIR can be passed along with the matching PYTHON_EXECUTABLE here
+# in order to avoid needing to call PYTHON_EXECUTABLE in the next block in each package
+# the precomputed value is only taken if the PYTHON_EXECUTABLE matches what was found above
+if(DEFINED PYTHON_INSTALL_DIR_TUPLE)
+  string(REPLACE ":" ";" _tuple ${PYTHON_INSTALL_DIR_TUPLE})
+  list(LENGTH _tuple _tuple_len)
+  if(_tuple_len EQUAL 2)
```
dirk-thomas, line 24: Should it print a warning if this is not the case?

cmake/python.cmake
```diff
@@ -15,6 +15,28 @@ if(WIN32)
...
+    list(GET _tuple 0 _tuple_PYTHON_EXECUTABLE)
+    list(GET _tuple 1 _tuple_PYTHON_INSTALL_DIR)
+    if("${_tuple_PYTHON_EXECUTABLE}" STREQUAL "${PYTHON_EXECUTABLE}")
```
dirk-thomas: Both strings should end with a space to avoid double expansion.

cmake/python.cmake
```diff
@@ -15,6 +15,28 @@ if(WIN32)
...
+      set(PYTHON_INSTALL_DIR "${_tuple_PYTHON_INSTALL_DIR}")
```
dirk-thomas, line 28: With the current patch this variable is not written to the cache but relies on `PYTHON_INSTALL_DIR_TUPLE` to be in the cache in order to be recomputed every time. Should this also be written to the cache for consistency?

python/catkin/builder.py
```diff
@@ -286,10 +287,42 @@ def get_multiarch():
     return out.strip()
 
 
-def get_python_install_dir():
-    # this function returns the same value as the CMake variable PYTHON_INSTALL_DIR from catkin/cmake/python.cmake
-    relevant_subfolders = 2 if os.name == 'nt' else 3
-    return os.sep.join(get_python_lib().split(os.sep)[-relevant_subfolders:])
+def get_python_install_dir_tuple(cmake_args=None):
```
dirk-thomas, line 290: Do you have a use case for the argument `cmake_args` in mind? Currently it is not being used anywhere therefore I would rather not implement it.

python/catkin/builder.py
```diff
@@ -286,10 +287,42 @@ def get_multiarch():
...
+        __cached_python_install_dir_tuple[cmake_args] = get_python_install_dir_tuple(cmake_args)
+    return __cached_python_install_dir_tuple[cmake_args]
+
+
+def get_python_install_dir(cmake_args=None):
+    return get_python_install_dir_tuple(cmake_args)[1]
+
+
+def get_python_install_dir_cached(cmake_args=None):
+    return get_python_install_dir_tuple_cached(cmake_args)[1]
```
dirk-thomas, line 325: Do we need four separate functions for this? Could this be refactored into fewer functions with keyword arguments e.g. for the caching behavior?

python/catkin/builder.py
wjwwood: If the user passes `-DPYTHON_VERSION` then this should be passed to this function as well. I stopped short of using it.

python/catkin/builder.py
wjwwood: I think it is clearer to have a "cached" version of the function rather having a parameter to control the behavior. I left the `get_python_install_dir` function (rather than replace it completely with `get_python_install_dir_tuple`) since someone might be using this function already.

cmake/python.cmake
wjwwood: That's a good idea, added in f5d4b53.

cmake/python.cmake
wjwwood: 5e54e48

cmake/python.cmake
wjwwood: I thought I was avoiding a conflict where the user could not set `PYTHON_INSTALL_DIR` with `-D` after using `PYTHON_INSTALL_DIR_TUPLE` because the cached version of `PYTHON_INSTALL_DIR_TUPLE` would always cause `PYTHON_INSTALL_DIR` to be set according to its contents. I also expected this to be passed by the build tool, which isn't annoyed by having to send it each time.

I don't think it is necessary to cache this variable, as the `PYTHON_INSTALL_DIR` is cached, but I can do so if you wish.

dirk-thomas:
Closing due to inactivity.

PR #834, opened 2016-11-13
Add targets for gtest libraries locally

jspricke:
The CMakeLists.txt of gtest adds install targets for the gtest libraries
and includes. Instead of doing add_subdirectory(), this adds the needed
libraries directly.

### DIFF
--- a/cmake/test/gtest.cmake
+++ b/cmake/test/gtest.cmake
@@ -117,7 +117,9 @@ if(NOT GTEST_FOUND)
       get_filename_component(_CATKIN_GTEST_BASE_DIR ${_CATKIN_GTEST_SOURCE_DIR} PATH)
       # add CMakeLists.txt from gtest dir
       set(_CATKIN_GTEST_BINARY_DIR ${CMAKE_BINARY_DIR}/gtest)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Can you please provide some more information. I have never seen a gtest library being installed on local builds nor do the Debian packages seem to contain them.

jspricke:
the `CMakeLists.txt` included by the `add_subdirectory()` is [this](https://github.com/google/googletest/blob/master/googletest/CMakeLists.txt), which we only need to build the libraries, as Debian doesn't (and shouldn't) provide them. The install rules where added in version 1.8.0, which you probably don't use. See [here](https://github.com/google/googletest/commit/98d988deac06637364f6cd41c45c3db4a8a0b6bc) for a discussion as well.

dirk-thomas:
I think the different options described in the referenced discussion seem to be superior compared to just adding library targets. The CMake file from `gtest` contains more logic to handle builds e.g. for Windows (`_VARIADIC_MAX`). Therefore I think `catkin` should still reuse the full CMake file since otherwise it would need to replicate such customizations itself.

dirk-thomas:
I will close this ticket for now. Please feel free to reply and it can be reopened.

PR #835, opened 2016-11-17
Use functions instead of macros where possible to avoid leaking variables

ksperling:
My primary reason for this PR is the first commit that makes `list_insert_in_workspace_order` a function, as it leaks a variable called `cmd` which triggers a bug in `ExternalProject_Add` (the value of the `cmd` variable becomes the default BUILD and INSTALL command for the external project, so `order_paths.py` just gets run a couple more times instead of actually building anything).

I filed a PR for CMake to fix this (https://gitlab.kitware.com/cmake/cmake/merge_requests/264), but I think it would be good to also fix this on the catkin side regardless. In general it seems to me using functions instead of macros wherever possible is desirable to avoid polluting the caller's variable namespace, but I'm by no means a CMake expert.

All tests still seem to pass (and my projects still build, and `ExternalProject_Add` works as expected)


ksperling:
Bump?

dirk-thomas:
Sorry for not responding on this before. I currently have to focus on ROS 2 related tasks. Hopefully I will have time to look at this in the next weeks.

ksperling:
Thanks for the update :)

cmake/catkin_libraries.cmake
```diff
@@ -49,12 +49,13 @@ macro(catkin_filter_libraries_for_build_configuration VAR)
         message(FATAL_ERROR "catkin_package() the list of libraries '${ARG_UNPARSED_ARGUMENTS}' ends with '${lib}' which is a build configuration keyword and must be followed by a library")
       endif()
     else()
-      list(APPEND ${VAR} "${lib}")
+      list(APPEND result "${lib}")
     endif()
     math(EXPR _index "${_index} + 1")
   endwhile()
+  set(${VAR} "${result}" PARENT_SCOPE)
   debug_message(10 "catkin_filter_libraries_for_build_configuration(${VAR} ${ARG_UNPARSED_ARGUMENTS} BUILD_TYPE ${ARG_BUILD_TYPE}) ${${VAR}}")
```
dirk-thomas: Does `${${VAR}}` work in this context? Does it get the updated value set in the parent scope in the line above?

cmake/catkin_libraries.cmake
```diff
@@ -101,15 +103,16 @@ endmacro()
 #
 # @public
 #
-macro(catkin_unpack_libraries_with_build_configuration VAR)
-  set(${VAR} "")
+function(catkin_unpack_libraries_with_build_configuration VAR)
+  set(result "")
   foreach(lib ${ARGN})
     string(REGEX REPLACE "^(debug|optimized|general)${CATKIN_BUILD_CONFIGURATION_KEYWORD_SEPARATOR}(.+)$" "\\1;\\2" lib "${lib}")
     list(APPEND ${VAR} "${lib}")
```
dirk-thomas: This should append to `result`.

cmake/catkin_libraries.cmake
ksperling: Indeed it doesn't see the updated value, good catch. Seems like the child scope starts out with a copy of the parent scope at the start of the function, rather than dynamically inheriting any variables not present locally.

ksperling:
Thanks for the review, I've addressed those two issues in the last commit. Let me know if you want me to squash everything into 1 commit.

cmake/catkin_libraries.cmake
```diff
@@ -81,14 +82,15 @@ macro(catkin_pack_libraries_with_build_configuration VAR)
...
-      list(APPEND ${VAR} "${lib}${CATKIN_BUILD_CONFIGURATION_KEYWORD_SEPARATOR}${library}")
+      list(APPEND result "${lib}${CATKIN_BUILD_CONFIGURATION_KEYWORD_SEPARATOR}${library}")
     else()
-      list(APPEND ${VAR} "${lib}")
+      list(APPEND result "${lib}")
     endif()
     math(EXPR _index "${_index} + 1")
   endwhile()
+  set(${VAR} "${result}" PARENT_SCOPE)
   #debug_message(10 "catkin_pack_libraries_with_build_configuration(${VAR} ${_argn}) ${${VAR}}")
```
dirk-thomas: Just to be safe if someone enables this line in the future can you also update it to use `${result}` instead of `${${VAR}}`.

cmake/catkin_libraries.cmake
```diff
@@ -144,16 +147,17 @@ macro(catkin_replace_imported_library_targets VAR)
...
+        list(APPEND result "${lib}")
       endif()
     else()
-      list(APPEND ${VAR} "${lib}")
+      list(APPEND result "${lib}")
     endif()
   endforeach()
+  set(${VAR} "${result}" PARENT_SCOPE)
   #set(_argn ${ARGN})
   #debug_message(10 "catkin_replace_imported_library_targets(${VAR} ${_argn}) ${${VAR}}")
```
dirk-thomas: Just to be safe if someone enables this line in the future can you also update it to use `${result}` instead of `${${VAR}}`.

dirk-thomas:
Thanks for iterating on this. I added two more comments inline.

If you can squash the commits afterwards that would be great. Otherwise I can take care of it during the merge.

ksperling:
Should all be tidied up now.

dirk-thomas:
Thanks for the update. It looks good to me. I ran a prerelease test locally with 100+ packages to convince myself that the chance for regressions is minimal. Lets hope the next release builds "agree" with that :wink:

@ros/ros_team Please review.

wjwwood (approved):
lgtm

PR #840, opened 2016-12-23
Fixes #839: Adding disabled tests to catkin_test_results summary.

### LINKED ISSUE #839: catkin_test_results does not show disabled tests
In the Google Test framework, a test can be marked as disabled by prepending DISABLED_ to the test suite or test case name. In such a case the test is marked disabled in the output XML. catkin_test_results does not appear to handle disabled tests. It looks like it would only require a few lines in two files to add this feature. I'd be happy to add it.

rsinnet:
I set it to read from both `disabled` (gtest) and `skip`attributes. I myself haven't seen `skip` but it appeared in the ament commit: https://github.com/ament/ament_tools/commit/0c6003b58ff0e5442034c7e643dab9035cdaeff8.

### DIFF
--- a/bin/catkin_test_results
+++ b/bin/catkin_test_results
@@ -26,8 +26,9 @@ def main():
     try:
         results = test_results(
             test_results_dir, show_verbose=args.verbose, show_all=args.all)
-        _, sum_errors, sum_failures = aggregate_results(results)
+        _, sum_errors, sum_failures, _ = aggregate_results(results)
         print_summary(results, show_stable=args.all)
+        # Skipped tests alone should not count as a failure
         if sum_errors or sum_failures:
             sys.exit(1)
     except Exception as e:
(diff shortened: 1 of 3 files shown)


rsinnet:
Closing and reopening, deleted two debug print statements that I added and want to rerun tests.

dirk-thomas:
@rsinnet Thank you for the contribution. I have created a new PR #848 against the `kinetic-devel` branch (while maintaining your authorship of the commit). Since all these Python function are potentially being used by other code changing their return values to include the skipped tests would break existing code. Therefore I have created new functions (with the suffix `2`) which provide the skipped tests while maintaining the existing API. It would be great if you could test the referenced PR and confirm if it works for you.

rsinnet:
Closing this since it has been merged via PR #848.

PR #842, opened 2017-01-03
support catkin_DEFINITIONS

mathias-luedtke:
Adds support for catkin_DEFINITIONS as discussed in https://github.com/ros/catkin/issues/610
The actual definitions has to be set via `cfg-extra` for now.

Should I add `add_defintions` to all `find_package(catkin..` occurrences in documentation?


doc/user_guide/find_package.rst
```diff
@@ -14,13 +14,14 @@ simultaneously, use
                cpp_common rostime roscpp_traits
                roscpp_serialization sensor_msgs)
   include_directories(${catkin_INCLUDE_DIRS})
+  add_definitions(${catkin_DEFINITIONS})
 
   add_executable(myexec ...)
-  target_link_libraries(${catkin_LIBRARIES})
+  target_link_libraries(myexec ${catkin_LIBRARIES})
```
dirk-thomas, line 20: Please create a separate PR for this fix (and the one below) since they can be directly merged.

dirk-thomas:
Please see the [latest comment](https://github.com/ros/catkin/issues/610#issuecomment-270153665) on the referenced issue why simply concatenating exported definitions might not work well.

doc/user_guide/find_package.rst
mathias-luedtke: together with first commit

mathias-luedtke:
closing for now.

PR #844, opened 2017-01-08
Added support for compile options

mathias-luedtke:
After the discussions in #610 I have now refactored and improved #842 

* introduces `<comp>_COMPILE_OPTIONS` and  `catkin_COMPILE_OPTIONS`
* all entries get set-filtered (stable) with warnings for colliding entries ( `key` vs. `key=` vs. `key=val1` vs. `key=val2`) 
* `<comp>_COMPILE_OPTIONS` will be passed to CMake config in `catkin_package`
* added `COMPILE_OPTIONS` option to `catkin_package`

I made the decision to switch to compile options completely for the following reasons:
* `add_compile_options` can handle all `*_DEFINITIONS` settings
* users would need to call `add_compile_options` and `add_definitions`
* duplicates and collisions would have to be checked in compile options and definitions

Open issues:
* How to properly add `test/test_list_append_keys_from.cmake` to catkin tests? It prints CMake warnings..
* ~~Not sure if the `DEFINES` option is really needed. In most cases `<comp>_COMPILE_OPTIONS` might be more reasonable, since it can be passed to the exporting package easily (see `set_compile_options` example package)~~
* documentation is not complete


cmake/catkin_package.cmake
```diff
@@ -75,6 +77,7 @@
 #     LIBRARIES projlib1 projlib2
 #     CATKIN_DEPENDS roscpp
 #     DEPENDS Eigen
+#     DEFINES MYDEFINE=1
```
dirk-thomas: Should this be named `COMPILE_OPTIONS` to match the underlying variable as well as used CMake function? Then defines would need to be passed as `-Dkey=value` but it allows to pass anything else too.

cmake/catkin_package.cmake
```diff
@@ -374,6 +377,15 @@ function(_catkin_package)
       endif()
     endforeach()
 
+    set(PKG_COMPILE_OPTIONS ${${PROJECT_NAME}_COMPILE_OPTIONS})
+    separate_arguments(PKG_COMPILE_OPTIONS)
+    separate_arguments(PROJECT_DEFINES)
```
dirk-thomas: Why is this necessary? The argument is already received as a list from `cmake_parse_arguments`.

cmake/catkin_package.cmake
```diff
@@ -374,6 +377,15 @@ function(_catkin_package)
       endif()
     endforeach()
 
+    set(PKG_COMPILE_OPTIONS ${${PROJECT_NAME}_COMPILE_OPTIONS})
+    separate_arguments(PKG_COMPILE_OPTIONS)
```
dirk-thomas, line 381: Same as for the next line. Why should we allow the user to specify a space separated string if CMake provides support for lists natively?

cmake/list_append_keys_from.cmake
```diff
@@ -0,0 +1,22 @@
...
+  foreach(_item ${${source}})
+      set(_found "")
+      string(REGEX REPLACE "^([^=]+)(=.*)$?" "^\\1(=.*)?$" _expr ${_item})
+      foreach(_elem ${${listname}})
+        if(_elem STREQUAL _item)
+          set(_found TRUE)
+          break()
+        elseif(_elem MATCHES ${_expr})
+          set(_found TRUE)
+          message(WARNING "could not add '${_item}' (from '${source}'), already contained: '${_elem}'")
```
dirk-thomas: This warning would also appear if the two values `_item` and `_elem` are identical which doesn't make much sense.

cmake/list_append_keys_from.cmake
```diff
@@ -0,0 +1,22 @@
+#
+# Append key[=value] elements from one list to another list if they are not already in the target list.
+#
+macro(list_append_keys_from listname source)
```
dirk-thomas: Please convert this into a function (see #835).

Also passing the source list name as `source` unnecessarily restricts how this function can be called. I would suggest considering all `ARGN` (beside `listname`) to be appended.

cmake/list_append_keys_from.cmake
```diff
@@ -0,0 +1,22 @@
+#
+# Append key[=value] elements from one list to another list if they are not already in the target list.
```
dirk-thomas: The docblock should document all arguments.

cmake/catkin_package.cmake
mathias-luedtke: These are not compile options, but definitions. `-D` is added automatically: https://github.com/ipa-mdl/catkin/blob/compile-options/cmake/catkin_package.cmake#L385
(I have adopted the syntax from https://github.com/ros/catkin/issues/610)

cmake/catkin_package.cmake
mathias-luedtke: Nothing prevents the user from specifying `set(${PROJECT_NAME}_COMPILE_OPTIONS "-Wall -DBla -fomit-frame-pointer")`.
As testing for unique options is crucial, we need to split the string into single options.

cmake/catkin_package.cmake
dirk-thomas: I saw that. But internally they are stored as compile options. Why should the user not be able to pass compile options which are not definitions through `catkin_package`?

cmake/catkin_package.cmake
dirk-thomas: I would expect the arguments to be passed separately and not in a single string, e.g.:

`set(${PROJECT_NAME}_COMPILE_OPTIONS "-Wall" "-DBla" "-fomit-frame-pointer")`

And while setting a global variable like this is possible it is hard to discover. Having a documented argument in `catkin_package` is imo preferred.

cmake/catkin_package.cmake
mathias-luedtke: I wasn't aware of that, this was only a safe guard. If `cmake_parse_arguments` already takes care of it this line can be omitted.

dirk-thomas: In the context of the related discussion I should clarify that I assume the arguments are being passed as a list, like this (same as e.g. `DEPENDS`):

`catkin_package(DEFINED "something" "something-else")`

cmake/list_append_keys_from.cmake
mathias-luedtke: I just copied it from list_append_unique and missed adding proper argument documentation, I can add this after we have decided on the parameter list below.

cmake/list_append_keys_from.cmake
mathias-luedtke: I didn't want to make this macro overly complex and spared the outer loop for `ARGN` since it was not needed for the use case. Its calling convention differs from `list_append_unique`, because it simplified the calling code.

I am not sure if this can be implemented as a function. I have not found any way to read a named variable that was defined in the parent scope.

cmake/list_append_keys_from.cmake
mathias-luedtke: This case should be covered by `if(_elem STREQUAL _item)`.

cmake/list_append_keys_from.cmake
dirk-thomas: The calling code will be almost identical. Either you call `list_append_keys_from(targte_list_name source_list_name)` or `list_append_keys_from(targte_list_name ${source_list_name})`. But the second signature also support calling it without having a source list in the first place but just plain items, e.g. `list_append_keys_from(targte_list_name "one" "two" "three" ${or_even_another_list})`.

The implemenation is very similar to [list_append_unique](https://github.com/ros/catkin/blob/184317881a0b6eb28d4386b791a2375c02da7d2a/cmake/list_append_unique.cmake#L9-L10), simply loop over `${ARGN}`.

cmake/list_append_keys_from.cmake
dirk-thomas: Good point. I missed that condition above. Sorry for the noise.

cmake/catkin_package.cmake
mathias-luedtke: > I would expect the arguments to be passed separately and not in a single string

I have seen add_definitions("...") in a couple of projects, so people might pass it as full string because they have found it somewhere in a stackoverflow answer.
IMHO if we need to check for conflicting settings then we should split them.

mathias-luedtke:
What do you think of using `COMPILE_OPTIONS` in `catkin_package`, but supporting `{$PROJECT_NAME}_COMPILE_OPTIONS` as well?
(Same behaviour as for `CFG_EXTRAS` and `EXPORTED_TARGETS`)

cmake/catkin_package.cmake
dirk-thomas: Ok, if we want to support both uses cases that sounds good.

cmake/list_append_keys_from.cmake
mathias-luedtke: And I now remember the reason for not passing the elements directly: I wanted to print the source of the conflicting entry in the warning.

dirk-thomas:
Yes, if the argument is called `COMPILE_OPTIONS` you should also be able to append to a global var `${PROJECT_NAME}_COMPILE_OPTIONS`. This basically allows you to contribute to `catkin_package()` without controlling the call itself.

In `ament_cmake` the `ament_package()` function has no arguments at all. For every kind of argument there is a separate function, e.g. `ament_export_libraries(...)`, which behind the scene just extends the global variable.

cmake/list_append_keys_from.cmake
dirk-thomas: I would suggest passing an optional keyword argument like `CONTEXT` for this purpose.

mathias-luedtke: Does an optional `CONTEXT` variable work with a variable argument list?

dirk-thomas: Yes, except that the variable argument list can't contain that specific keyword, so I think `CONTEXT` would be "safe". It needs to be extracted and used with something like this:

```
cmake_parse_arguments(ARG "" "CONTEXT" "" ${ARGN})
# ARG_CONTEXT
# ARG_UNPARSED_ARGUMENTS
```

mathias-luedtke: > I have not found any way to read a named variable that was defined in the parent scope.

Not sure what I have tested back then. `${${listname}}` works like a charm...

mathias-luedtke:
@dirk-thomas: Thanks for your feedback! I have updated the code accordingly.
I haven't rebase/squashed yet to simplify the review.

In the end it will be reduced to 2-3 commis.

cmake/catkin_package.cmake
```diff
@@ -374,6 +377,11 @@ function(_catkin_package)
...
+    separate_arguments(PROJECT_COMPILE_OPTIONS)
```
dirk-thomas, line 382: Since this is a user provided variable I don't think it is a good idea to mutate it. This should store the modified value in a local variable instead.

cmake/list_insert_options.cmake
```diff
@@ -0,0 +1,55 @@
...
+# :param listname: name of the destination list
+# :type listname: string
+#
+# :param ARGN: options to be inserted
+# :type ARGN: list key[=value] strings
+#
+# :param CONTEXT: context to be printen in case of errors
+# :type CONTEXT: string
+#
+# :param LEVEL: level of the error message
```
dirk-thomas, line 13: This should mention the default level.

cmake/list_insert_options.cmake
```diff
@@ -0,0 +1,55 @@
+#
+# Append key[=value] elements from one list to another list if they are not already in the target list.
```
dirk-thomas, line 2: If the same key is being inserted multiple time with different values shouldn't the last one be used (instead of the first one)?

cmake/list_insert_options.cmake
```diff
@@ -0,0 +1,55 @@
...
+      # make a regex that matches key=[value] form
+      string(REGEX REPLACE "^([^=]+)(=.*)$?" "^\\1(=.*)?$" expr ${item}) 
+
+      foreach(elem ${${listname}})
+        if(elem STREQUAL item)
+          set(found TRUE)
+          break()
+        elseif(elem MATCHES ${expr})
+          set(found TRUE)
+          message(${ARG_LEVEL} "could not add '${item}' (from '${ARG_CONTEXT}'), already contained: '${elem}'")
```
dirk-thomas, line 46: What if this is intentionally passed? Related to the question above.

.travis.yml
```diff
@@ -1,4 +1,6 @@
 language: python
+dist: trusty
+sudo: required
```
dirk-thomas: Is this necessary for this PR? If not please remove it from this patch.

cmake/all.cmake
```diff
@@ -128,6 +128,7 @@ foreach(filename
     list_append_deduplicate
     list_append_unique
     list_insert_in_workspace_order
+    list_insert_options
```
dirk-thomas, line 131: Would `list_append_unique_options` make more sense for the name? Since the values are always appended. The only difference to the `list_append_unique` function above is that is only considers the key part and not the value when checking for existence.

dirk-thomas (commented):
I am sorry for the late response. Sadly there is only very little ROS 1 time. Since a new release of this package is planned for the near future I would like to get this in though. Please see the few inline comment / questions.

nuclearsandwich:
@ros-pull-request-builder retest this please.

cmake/catkin_package.cmake
mathias-luedtke: Good point, I will change it.

cmake/list_insert_options.cmake
mathias-luedtke: If I recall it correctly, the dependencies will be resolved first.
So changing some options might break ABI or even API.

cmake/list_insert_options.cmake
mathias-luedtke: Agreed.

cmake/list_insert_options.cmake
dirk-thomas: Well, I guess that is the problem of trying to set different values (coming from different packages). You already have the same problem with two different values both coming from dependencies because you have little control in which order they are being found.

I think it is just counter intuitive if this function uses the first value.

.travis.yml
mathias-luedtke: Not sure about `sudo: required` (I guess this is the default anyways.)
`dist: trusty` is required because of [`cmake_minimum_required(VERSION 2.8.12)`](https://github.com/ipa-mdl/catkin/blob/compile-options/test/mock_resources/src/compile_options/set_compile_options/CMakeLists.txt#L1), which is needed for  `add_compile_options` in the tests. As an alternative `add_definitions` could be used though.

cmake/all.cmake
mathias-luedtke: I am not in favor of a particular name.
I have chosen "insert" because of the ~~set~~map logic, but if you like `list_append_unique_options` better that's fine for me as well.

cmake/list_insert_options.cmake
```diff
@@ -7,7 +7,7 @@
 # :param ARGN: options to be inserted
 # :type ARGN: list key[=value] strings
 #
-# :param CONTEXT: context to be printen in case of erros
+# :param CONTEXT: context to be printen in case of errors
```
mathias-luedtke: *printed

cmake/list_insert_options.cmake
mathias-luedtke: >  You already have the same problem with two different values both coming from dependencies because you have little control in which order they are being found.

[You have proposed](https://github.com/ros/catkin/issues/610#issuecomment-270165325) that in this case the dependencies must be resolved separately and cannot use the combined compile options.

> I think it is just counter intuitive if this function uses the first value.

For me it's not counter intuitive, it resembles the C++ `map::insert` logic

dirk-thomas: We should clarify the goal here first. Then come up with the behavior and name of this function.

The function is being used when calling `find_package(catkin REQUIRED COMPONENTS foo bar)` where as both components might have (recursive) dependencies themselves. What do you expect to happen if they export different values for the same key and why?

Re-reading some of this thread in the original comment you mentioned that this would be a "collision" and considered to print a warning.

mathias-luedtke:
I can address the simple issues tomorrow.

However, I am not sure about the first/last-issue.
Looks like the values in `catkin_package` actually [get preferred](https://github.com/ipa-mdl/catkin/blob/compile-options/cmake/templates/pkgConfig.cmake.in#L219).

In which order do the dependencies get resolved?
The topological order at build time?

dirk-thomas:
> In which order do the dependencies get resolved?
> The topological order at build time?

I don't think the order will equal the topological order at build time. Especially since there might be packages which don't have to processed sequentially but can be build in parallel.

From a CMake point of view it calls `find_package` of all `COMPONENTS` sequentially and each of these packages will again `find_package` its exported dependencies. So it is more like "depth first".

cmake/list_insert_options.cmake
mathias-luedtke: > What do you expect to happen if they export different values for the same key and why?

As we discussed it should print a warning (in any case).
It might even better to issue a  fatal error, as the situation is unresolvable (?).
This way it doesn't matter which comes first or last ;)

mathias-luedtke:
> From a CMake point of view it calls find_package of all COMPONENTS sequentially and each of these packages will again find_package its exported dependencies. So it is more like "depth first".

"depth first" with the order of depends passed by the author of each package.
It is alphabetical by convention, but this is not enforced. I would consider it more or less arbitrary then.

cmake/list_insert_options.cmake
dirk-thomas: Since you are proposing the feature I will leave it up to you to decide if it is a warning by default or an error.

From my point of view this new feature will likely not be widely usable for e.g. setting the compiler standard version (which was one reason it was proposed in the first place I think). Simply because any situation where packages don't agree on the value are going to be a problem.

mathias-luedtke: Right now I don't know if we should merge it.
It adds complexity, but might lead to problems if it gets used widely.

For the C++ standards issue I have implemented [other work-arounds](https://github.com/ipa-mdl/enable_cxx11_support).

dirk-thomas: > Right now I don't know if we should merge it.
> It adds complexity, but might lead to problems if it gets used widely.

I share your concern.

> For the C++ standards issue I have implemented other work-arounds.

Depending on which CMake versions you need to support `CMAKE_CXX_STANDARD` might be an interesting option too (to be used in your `enable_cxx11_support` package).

mathias-luedtke: > Depending on which CMake versions you need to support CMAKE_CXX_STANDARD might be an interesting option too (to be used in your enable_cxx11_support package).

The package was meant for `indigo` as well, so I don't want to rely on new CMake features.
It seems like most users have settled to set the C++11 options in any depending package :-/

mathias-luedtke:
As [discussed](https://github.com/ros/catkin/pull/844#discussion_r127781182) this feature might introduce more problems (conflicting settings..) than it solves.
That's why I am closing this PR.

@dirk-thomas: Thanks for reviewing and discussing it!

dirk-thomas:
Thank you for your effort!

PR #849, opened 2017-01-26
Added notes about not filling the requires field.

peci1:
The idea for this edit originated here: https://github.com/ros-infrastructure/rosdep/issues/497 .


dirk-thomas:
:+1: Thank you!

PR #861, opened 2017-03-09
prefer reporting problems to bugtracker / website before maintainer

dirk-thomas:
If a package has a bugtracker / website url refer the user to those in case of problems. If neither of these URLs is specified in the manifest continue to reference the maintainers directly.

This might be useful across all active ROS distros. So considering it for backporting to `indigo-devel`.

@ros/ros_team Please review.


trainman419:
This looks great! thanks for the quick fix.

wjwwood (approved):
lgtm

tfoote (approved):

PR #862, opened 2017-03-28
use environment variable to extend environment in plain shell

dirk-thomas:
In a plain shell the following doesn't work: `. setup.sh --extend`

To allow extending the environment in plain shells this patch adds an environment variable which can be use to pass arguments to the `_setup_util.py` invocation: `CATKIN_SETUP_UTIL_ARGS=--extend , setup.sh`

If the environment variable is not set (default) nothing changes. So this should have no side effects if not being used.

@ros/ros_team Please review.

### DIFF
--- a/cmake/templates/setup.sh.in
+++ b/cmake/templates/setup.sh.in
@@ -5,6 +5,8 @@
 # It tries it's best to undo changes from a previously sourced setup file before.
 # Supported command line options:
 # --extend: skips the undoing of changes from a previously sourced setup file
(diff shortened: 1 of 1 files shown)


wjwwood (approved):
lgtm

What's the worst that could happen 😄

dirk-thomas:
> What's the worst that could happen :smile:

:fearful: :scream: :punch: :zap: :boom: :fire: just some ideas based on past experiences... :wink:

tfoote (approved):

Issue #872, opened 2017-05-11
catkin forbid importing in setup.py


asmodehn:
Catkin prevents using any function using importe names in `setup.py` python code.

For example, adding : 
```python
import xml.etree.ElementTree
def parsing():
    xml.etree.ElementTree.parse("package.xml")

parsing()
```

to any working `setup.py` will fail on `catkin_make`, with the error : 
```
-- catkin 0.7.6
Traceback (most recent call last):
  File "/home/alexv/Projects/catkin/cmake/interrogate_setup_dot_py.py", line 259, in <module>
    main()
  File "/home/alexv/Projects/catkin/cmake/interrogate_setup_dot_py.py", line 250, in main
    exec(fh.read(), {}, local_vars)
  File "<string>", line 10, in <module>
  File "<string>", line 8, in parsing
NameError: global name 'xml' is not defined
CMake Error at cmake/safe_execute_process.cmake:11 (message):
  
  execute_process(/home/alexv/Projects/catkin/build/catkin_generated/env_cached.sh
  "/usr/bin/python"
  "/home/alexv/Projects/catkin/cmake/interrogate_setup_dot_py.py" "catkin"
  "/home/alexv/Projects/catkin/setup.py"
  "/home/alexv/Projects/catkin/build/catkin_generated/setup_py_interrogation.cmake")
  returned error code 1
Call Stack (most recent call first):
  cmake/catkin_python_setup.cmake:69 (safe_execute_process)
  CMakeLists.txt:23 (catkin_python_setup)


-- Configuring incomplete, errors occurred!
```

This seems to be related to https://github.com/ros/catkin/blob/kinetic-devel/cmake/interrogate_setup_dot_py.py#L250

I ll work on a PR for this, although I am unsure where this part of the code is tested...

asmodehn:
PR here : https://github.com/ros/catkin/pull/873

dirk-thomas:
Addressed by #873. Thanks.

PR #873, opened 2017-05-11
replacing exec call with runpy.run_path

asmodehn:


asmodehn:
Fixing https://github.com/ros/catkin/issues/872

dirk-thomas:
I am sorry for the late reply. The patch looks great. I didn't know about `runpy`. This is much cleaner. Thanks.

asmodehn:
Looks like we need to do the same on interrogate_setup_dot_py : 
http://build.ros.org/job/Kpr__genmsg__ubuntu_xenial_amd64/8/consoleFull

asmodehn:
Ah no maybe it s just that this version didn't make it on the build farm yet... I ll wait a bit more.

dirk-thomas:
`genmsg` will only use released versions of `catkin`. So the `genmsg` PR job will only pass once a new version of `catkin` has been released with this patch.

PR #879, opened 2017-07-19
remove workaround using -l:<libpath> necessary for older pkg-config versions

### LINKED ISSUE #694: ld-2.24.90 does not support -l:/full/path/to/library.so

dirk-thomas:
Fixes #694.


mintar:
I've just tested this and it works for my use case!

dirk-thomas:
Thank you for trying the patch and confirming it works.

Issue #883, opened 2017-08-15
gtest.cmake assumes system library is a shared lib


rctoris:
We had a developer computer that had a pre-built version of gtest installed which catkin correctly found. However, it looks like it's assuming the library is a shared library: https://github.com/ros/catkin/blob/kinetic-devel/cmake/test/gtest.cmake#L166-L168

In our case it wasn't and leads to all kinds of wonky linking error messages. We fixed it by remaking gtest as a shared library but it would be nice if it could support either.

dirk-thomas:
Can you please clarify on which platform you had that problem and how that system was being setup.

Upstream highly recommends not to use a pre-build library but to build `gtest` together with your code. Therefore I don't think the maintainers will spend time supporting a static library. If you could provide a pull request for supporting this use case I would be happy to review it.

rctoris:
This was on Ubuntu 14.04 running indigo. Since catkin currently supports some conditions for this as apposed to no support, I'll look into getting it to check more dynamically.

PR #884, opened 2017-08-15
[indigo/jade] Dynamically check gtest library type

rctoris:
From the CMake docs: https://cmake.org/Wiki/CMake/Tutorials/Exporting_and_Importing_Targets

"
In some cases, such as when using the result of a find_library command to create an imported target, the type of the library (static or shared) is not known. As of CMake 2.6.2 one may specify this explicitly:
```
 find_library(FOO_LIBRARY NAMES foo)
 add_library(foo UNKNOWN IMPORTED)
 set_property(TARGET foo PROPERTY IMPORTED_LOCATION "${FOO_LIBRARY}")
 add_executable(myexe src1.c src2.c)
 target_link_libraries(myexe foo)
```
"

It looks like catkin requires `cmake_minimum_required(VERSION 2.8.3)` so the `UNKNOWN` keyword is supported. Tested without a pre-built (the most common catkin case), and with a shared/static prebuilt.

Fixes https://github.com/ros/catkin/issues/883


dirk-thomas:
I will close this in favor of #885. After changs to the latest development branch are being merged they will be considered for backporting. No need to create separate PRs for older branch (except if they are different).

rctoris:
:+1: good to know, thanks!

PR #885, opened 2017-08-15
[kinetic/lunar] Dynamically check gtest library type

rctoris:
Same as but for https://github.com/ros/catkin/pull/884 kinetic/lunar

Fixes https://github.com/ros/catkin/issues/883


dirk-thomas:
Thank you for the improvement. Looks and works great.

PR #889, opened 2017-08-31
now usable as normal python package

asmodehn:
Following the same pattern as ros/genpy#86 and ros/genmsg#68

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,9 +1,61 @@
 #!/usr/bin/env python
 
-from distutils.core import setup
-from catkin_pkg.python_setup import generate_distutils_setup
+from setuptools import setup
+import os
 
-d = generate_distutils_setup(
+import xml.etree.ElementTree
+
+
+class XmlNodes(object):
+    """
+    Simple API for xml.etree.ElementTree that collapses similar tags.
+    Specially designed for our usage here.
+    """
+
+    def __init__(self, xml_elems):
+        """Handle one list of xml elements"""
+        self.xml_elems = xml_elems
+
+    def __getattr__(self, item):
+        children = XmlNodes([e for e in self.xml_elems if e.tag == item])
+        if len(children.xml_elems) == 0:
+            raise AttributeError
+        else:
+            return children
+
+    def __call__(self, attrib_name=None):
+        if attrib_name is None:  # return the text
+            return ", ".join([e.text for e in self.xml_elems])
+        else:  # return the values of that attribute
+            return ", ".join([e.attrib.get(attrib_name) for e in self.xml_elems])
+
+    def filter(self, attrib_name, attrib_value):
+        """special function to remove nodes we are not interested in"""
+        return XmlNodes([e for e in self.xml_elems if e.attrib.get(attrib_name) == attrib_value])
+
+
+class PackageXml(XmlNodes):
+    """Extracts data from a package.xml file"""
+    def __init__(self, package_xml_path='package.xml'):
+        tree = xml.etree.ElementTree.parse(package_xml_path)
+        root = tree.getroot()
+        assert root.tag == 'package'
+        # TODO : check package format version
+        super(PackageXml, self).__init__([c for c in root])  # here we get all children already
+
+
+ros_package = PackageXml(os.path.join(os.path.dirname(__file__), 'package.xml'))
+
+setup(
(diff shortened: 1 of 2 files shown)


sloretz:
Thanks for the pull request! If I understand correctly this replaces an import of catkin_pkg in the setup.py with parsing the package's package.xml using only standard libraries. Just curious, what does this change enable?

asmodehn:
The python code part is now usable as a normal python package (pip / setuptools [is the standard python way](https://packaging.python.org)) : 

```
alexv@alexv-pc:~$ mkvirtualenv catkin_test
Running virtualenv with interpreter /usr/bin/python2
New python executable in /home/alexv/.virtualenvs/catkin_test/bin/python2
Also creating executable in /home/alexv/.virtualenvs/catkin_test/bin/python
Installing setuptools, pkg_resources, pip, wheel...done.

(catkin_test) alexv@alexv-pc:~$ pip install -e "git+https://github.com/asmodehn/catkin.git@setuptools#egg=ros_catkin"
Obtaining ros_catkin from git+https://github.com/asmodehn/catkin.git@setuptools#egg=ros_catkin
  Cloning https://github.com/asmodehn/catkin.git (to setuptools) to ./.virtualenvs/catkin_test/src/ros-catkin
Requirement already satisfied: argparse in /usr/lib/python2.7 (from ros_catkin)
Collecting catkin-pkg (from ros_catkin)
Collecting python-dateutil (from catkin-pkg->ros_catkin)
  Using cached python_dateutil-2.6.1-py2.py3-none-any.whl
Collecting docutils (from catkin-pkg->ros_catkin)
  Using cached docutils-0.14-py2-none-any.whl
Collecting six>=1.5 (from python-dateutil->catkin-pkg->ros_catkin)
  Using cached six-1.10.0-py2.py3-none-any.whl
Installing collected packages: six, python-dateutil, docutils, catkin-pkg, ros-catkin
  Running setup.py develop for ros-catkin
Successfully installed catkin-pkg-0.3.6 docutils-0.14 python-dateutil-2.6.1 ros-catkin six-1.10.0

(catkin_test) alexv@alexv-pc:~$ python -c "import catkin.find_in_workspaces; print(catkin.find_in_workspaces.find_in_workspaces(project='rospy'))"
[]
(catkin_test) alexv@alexv-pc:~$ source /opt/ros/kinetic/setup.bash 
(catkin_test) alexv@alexv-pc:~$ python -c "import catkin.find_in_workspaces; print(catkin.find_in_workspaces.find_in_workspaces(project='rospy'))"
['/opt/ros/kinetic/share/rospy']

```

I don't know of any direct practical use for catkin itself yet...
But since [roslib depends on it](https://github.com/ros/ros/blob/lunar-devel/core/roslib/src/roslib/packages.py#L49), it is needed for https://github.com/ros/ros/pull/151, even though the dependency is [not explicit](https://github.com/ros/ros/blob/lunar-devel/core/roslib/setup.py) in `setup.py`.

For more background on this, you can refer to : https://github.com/ros/genpy/issues/81

setup.py
```diff
@@ -1,9 +1,61 @@
 #!/usr/bin/env python
 
-from distutils.core import setup
-from catkin_pkg.python_setup import generate_distutils_setup
```
mikepurvis, line 4: What is gained here by removing the dependency on `catkin_pkg`, which is itself a PyPI package?

tfoote: The main benefit of this is that you do not need to have `catkin_pkg` installed to be able to run the setup.py script to invoke the installation. A standard python package you should be able to naively clone it and then run `python setup.py install` and it will find the declared dependencies install them and then run the build/install steps for the current package. Without this change if you try to follow that standard workflow w/o `catkin_pkg` preinstalled the setup.py will fail before being able to parse the dependencies that says it needs catkin_pkg and you're stuck until you manually determine that catkin_pkg has to be installed and install it yourself, then try to run catkin's setup.py again.

mikepurvis: Oh I see. So the only things you can really safely import in the `setup.py` context are basically pip, setuptools, and wheel— the stuff that's available in a bare virtualenv.

dirk-thomas:
Closing due to inactivity.

All the additional parsing logic for this corner case which isn't really supported doesn't look reasonable to me.

PR #896, opened 2017-10-16
Add support for GMock

v-lopez:
I took the changes from @kyrofa, rebased them on kinetic and updated them for GMock 1.7 which no longer has an embedded gtest version.

### DIFF
--- a/cmake/test/gtest.cmake
+++ b/cmake/test/gtest.cmake
@@ -167,7 +167,7 @@ function(catkin_find_gmock_source include_paths src_paths found base_dir
             NO_DEFAULT_PATH NO_CMAKE_FIND_ROOT_PATH
   )
 
-  # If we found gmock, ensure gtest is contained within it
+  # If we found gmock, set the variables accordingly
   if(_GMOCK_INCLUDES AND _GMOCK_SOURCES)
     get_filename_component(SOURCE_DIR ${_GMOCK_SOURCES} PATH)
     get_filename_component(BASE_DIR ${SOURCE_DIR} PATH)
@@ -208,7 +208,7 @@ if(NOT GMOCK_FOUND)
                                gmock_libs gmock_main_libs)
 
       # If we found gmock, set it up to be built (which will also build gtest,
-      # since it's bundled)
+      # since it's included by gmock's CMakeLists.txt)
       if(gmock_found)
         set(GMOCK_FROM_SOURCE_FOUND ${gmock_found} CACHE INTERNAL "")
         set(GMOCK_FROM_SOURCE_INCLUDE_DIRS ${gmock_include_dir} CACHE INTERNAL "")
--- a/cmake/test/gtest.cmake
+++ b/cmake/test/gtest.cmake
@@ -253,7 +253,7 @@ if(NOT GMOCK_FOUND)
         set_target_properties(${gmock_libs} ${gmock_main_libs}
                               PROPERTIES EXCLUDE_FROM_ALL 1)
 
-        message(STATUS "Found gmock sources under '${gmock_base_dir}': gtests will be built")
+        message(STATUS "Found gmock sources under '${gmock_base_dir}': gmock and gtests will be built")
       else() # gmock not found-- look for system-installed gtest by itself
         set(_include_paths "/usr/include/gtest")
         if(CATKIN_TOPLEVEL)
@@ -345,7 +345,7 @@ if(NOT GMOCK_FOUND)
     set(GTEST_BOTH_LIBRARIES ${GTEST_BOTH_LIBRARIES} CACHE INTERNAL "")
   endif()
 else()
-  message(STATUS "Found gmock: gtests will be built")
+  message(STATUS "Found gmock: gmock and gtests will be built")
   add_library(gmock UNKNOWN IMPORTED)
   set_target_properties(gmock PROPERTIES IMPORTED_LOCATION "${GMOCK_LIBRARIES}")
   add_library(gmock_main UNKNOWN IMPORTED)
--- a/cmake/test/gtest.cmake
+++ b/cmake/test/gtest.cmake
@@ -115,8 +115,8 @@ function(catkin_find_gtest_source include_paths src_paths found base_dir
 
   # Find the gtest sources
   find_file(_GTEST_SOURCES "gtest.cc"
-           PATHS ${src_paths}
-           NO_DEFAULT_PATH NO_CMAKE_FIND_ROOT_PATH
+            PATHS ${src_paths}
+            NO_DEFAULT_PATH NO_CMAKE_FIND_ROOT_PATH
   )
 
   # If we found gtest, set the variables accordingly
@@ -163,8 +163,8 @@ function(catkin_find_gmock_source include_paths src_paths found base_dir
 
   # Find the gmock sources
   find_file(_GMOCK_SOURCES "gmock.cc"
-           PATHS ${src_paths}
-           NO_DEFAULT_PATH NO_CMAKE_FIND_ROOT_PATH
+            PATHS ${src_paths}
+            NO_DEFAULT_PATH NO_CMAKE_FIND_ROOT_PATH
   )
 
   # If we found gmock, ensure gtest is contained within it
--- a/cmake/test/gtest.cmake
+++ b/cmake/test/gtest.cmake
@@ -216,6 +216,8 @@ if(NOT GMOCK_FOUND)
         set(GMOCK_FROM_SOURCE_LIBRARIES ${gmock_libs} CACHE INTERNAL "")
         set(GMOCK_FROM_SOURCE_MAIN_LIBRARIES ${gmock_main_libs} CACHE INTERNAL "")
 
+        # overwrite CMake install command to skip install rules for gtest targets
+        # which have been added in version 1.8.0
         set(_CATKIN_SKIP_INSTALL_RULES TRUE)
         function(install)
           if(_CATKIN_SKIP_INSTALL_RULES)
@@ -254,6 +256,9 @@ if(NOT GMOCK_FOUND)
           set(GTEST_FROM_SOURCE_LIBRARIES ${gtest_libs} CACHE INTERNAL "")
           set(GTEST_FROM_SOURCE_MAIN_LIBRARIES ${gtest_main_libs} CACHE INTERNAL "")
 
+
+          # overwrite CMake install command to skip install rules for gtest targets
+          # which have been added in version 1.8.0
           set(_CATKIN_SKIP_INSTALL_RULES TRUE)
           function(install)
             if(_CATKIN_SKIP_INSTALL_RULES)
--- a/cmake/test/gtest.cmake
+++ b/cmake/test/gtest.cmake
@@ -172,15 +172,37 @@ function(catkin_find_gmock_source include_paths src_paths found base_dir
     get_filename_component(SOURCE_DIR ${_GMOCK_SOURCES} PATH)
     get_filename_component(BASE_DIR ${SOURCE_DIR} PATH)
 
-    get_filename_component(INCLUDE_DIR ${_GMOCK_INCLUDES} PATH)
-    get_filename_component(INCLUDE_DIR ${INCLUDE_DIR} PATH)
+    set(_gtest_include_paths "/usr/include/gtest")
+    if(CATKIN_TOPLEVEL)
+      # Ensure current workspace is searched before system path
+      list(INSERT _gtest_include_paths 0 "${CMAKE_SOURCE_DIR}/googletest/googletest/include/gtest")
+    endif()
 
-    set(${found} TRUE PARENT_SCOPE)
-    set(${base_dir} ${BASE_DIR} PARENT_SCOPE)
-    set(${include_dir} ${gtest_include_dir} ${INCLUDE_DIR} PARENT_SCOPE)
-    set(${lib_dir} ${CMAKE_BINARY_DIR}/gmock PARENT_SCOPE)
-    set(${libs} "gmock" PARENT_SCOPE)
-    set(${main_libs} "gmock_main" PARENT_SCOPE)
+    set(_gtest_source_paths "/usr/src/gtest/src")
+    if(CATKIN_TOPLEVEL)
+      # Ensure current workspace is searched before system path
+      list(INSERT _gtest_source_paths 0 "${CMAKE_SOURCE_DIR}/googletest/googletest/src")
+    endif()
+
+    catkin_find_gtest_source("${_gtest_include_paths}"
+                             "${_gtest_source_paths}" gtest_found
+                             gtest_base_dir gtest_include_dir gtest_lib_dir
+                             gtest_libs gtest_main_libs)
+
+    # If we found gtest, finding gmock succeeded
+    if(gtest_found)
+      get_filename_component(INCLUDE_DIR ${_GMOCK_INCLUDES} PATH)
+      get_filename_component(INCLUDE_DIR ${INCLUDE_DIR} PATH)
+
+      set(${found} TRUE PARENT_SCOPE)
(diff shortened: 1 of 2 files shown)


dirk-thomas:
In Ubuntu Xenial `gtest` 1.7 is being shipped without `gmock` (https://packages.ubuntu.com/xenial/libgtest-dev). Only as of Zesty `googletest` 1.8 contain both `gtest` and `gmock` (https://packages.ubuntu.com/zesty/googletest). So this can be targeted for the next Ubuntu LTS and ROS Melodic but not for ROS Kinetic and Lunar.

v-lopez:
Actually the modifications I made are for the case where gtest and gmock are not bundled together, as is the case in Xenial, and should also work in the case only gtest is installed. 

In Trusty they were separated as well, but gmock contained it's own version of gtest.

cmake/test/gtest.cmake
```diff
@@ -85,97 +85,261 @@ function(catkin_add_executable_with_gtest target)
...
+  # Find the gtest headers
+  find_file(_GTEST_INCLUDES "gtest.h"
+            PATHS ${include_paths}
+            NO_DEFAULT_PATH NO_CMAKE_FIND_ROOT_PATH
+  )
+
+  # Find the gtest sources
+  find_file(_GTEST_SOURCES "gtest.cc"
+           PATHS ${src_paths}
+           NO_DEFAULT_PATH NO_CMAKE_FIND_ROOT_PATH
```
kyrofa: The indentation here seems off by a space.

cmake/test/gtest.cmake
```diff
@@ -85,97 +85,261 @@ function(catkin_add_executable_with_gtest target)
...
+  # Find the gmock headers
+  find_file(_GMOCK_INCLUDES "gmock.h"
+            PATHS ${include_paths}
+            NO_DEFAULT_PATH NO_CMAKE_FIND_ROOT_PATH
+  )
+
+  # Find the gmock sources
+  find_file(_GMOCK_SOURCES "gmock.cc"
+           PATHS ${src_paths}
+           NO_DEFAULT_PATH NO_CMAKE_FIND_ROOT_PATH
```
kyrofa: Off by a space here as well.

dirk-thomas:
I see. In Xenial they are not bundled in the same Debian package anymore but as separate packages where `google-mock` depends on `libgtest-dev`. I will add some comments / questions inline.

cmake/test/gtest.cmake
```diff
@@ -85,97 +85,261 @@ function(catkin_add_executable_with_gtest target)
...
+        # Ensure current workspace is searched before system path
+        list(INSERT _source_paths 0 "${CMAKE_SOURCE_DIR}/gmock/src")
+      endif()
+
+      catkin_find_gmock_source("${_include_paths}" "${_source_paths}" gmock_found
+                               gmock_base_dir gmock_include_dir gmock_lib_dir
+                               gmock_libs gmock_main_libs)
+
+      # If we found gmock, set it up to be built (which will also build gtest,
+      # since it's bundled)
```
dirk-thomas: The comment still mentions that `gtest` is bundled.

cmake/test/gtest.cmake
v-lopez: Nice catch, fixed both.

cmake/test/gtest.cmake
```diff
@@ -85,97 +85,261 @@ function(catkin_add_executable_with_gtest target)
...
+          endif()
+          _install(${ARGN})
+        endfunction()
+        add_subdirectory(${gmock_base_dir} ${gmock_lib_dir})
+        set(_CATKIN_SKIP_INSTALL_RULES FALSE)
+
+        set_target_properties(${gmock_libs} ${gmock_main_libs}
+                              PROPERTIES EXCLUDE_FROM_ALL 1)
+
+        message(STATUS "Found gmock sources under '${gmock_base_dir}': gtests will be built")
```
dirk-thomas: The message seems inconsistent. The first part mentions gmock - the second part gtest.

cmake/test/gtest.cmake
```diff
@@ -85,97 +85,261 @@ function(catkin_add_executable_with_gtest target)
...
+          endfunction()
+          add_subdirectory(${gtest_base_dir} ${gtest_lib_dir})
+          set(_CATKIN_SKIP_INSTALL_RULES FALSE)
+          set_target_properties(${gtest_libs} ${gtest_main_libs}
+                                PROPERTIES EXCLUDE_FROM_ALL 1)
+
+          message(STATUS "Found gtest sources under '${gtest_base_dir}': gtests will be built")
+        else()
+          if(CATKIN_TOPLEVEL)
+            message(STATUS "gtest not found, C++ tests can not be built. Please install the gtest headers globally in your system or checkout gtest (by running 'svn checkout http://googletest.googlecode.com/svn/tags/release-1.6.0 gtest' in the source space '${CMAKE_SOURCE_DIR}' of your workspace) to enable gtests")
```
dirk-thomas: Should this reference at least version 1.7 or even newer and the GitHub repo instead?

cmake/test/gtest.cmake
```diff
@@ -85,97 +85,261 @@ function(catkin_add_executable_with_gtest target)
...
+        set(GMOCK_FROM_SOURCE_LIBRARIES ${gmock_libs} CACHE INTERNAL "")
+        set(GMOCK_FROM_SOURCE_MAIN_LIBRARIES ${gmock_main_libs} CACHE INTERNAL "")
+
+        set(_CATKIN_SKIP_INSTALL_RULES TRUE)
+        function(install)
+          if(_CATKIN_SKIP_INSTALL_RULES)
+            return()
+          endif()
+          _install(${ARGN})
+        endfunction()
```
dirk-thomas, line 347: Please keep the comment why this is necessary.

Same below.

cmake/test/gtest.cmake
```diff
@@ -85,97 +85,261 @@ function(catkin_add_executable_with_gtest target)
...
-  add_library(gtest UNKNOWN IMPORTED)
-  set_target_properties(gtest PROPERTIES IMPORTED_LOCATION "${GTEST_LIBRARIES}")
-  add_library(gtest_main UNKNOWN IMPORTED)
-  set_target_properties(gtest_main PROPERTIES IMPORTED_LOCATION "${GTEST_MAIN_LIBRARIES}")
-  set(GTEST_FOUND ${GTEST_FOUND} CACHE INTERNAL "")
-  set(GTEST_INCLUDE_DIRS ${GTEST_INCLUDE_DIRS} CACHE INTERNAL "")
-  set(GTEST_LIBRARIES ${GTEST_LIBRARIES} CACHE INTERNAL "")
-  set(GTEST_MAIN_LIBRARIES ${GTEST_MAIN_LIBRARIES} CACHE INTERNAL "")
-  set(GTEST_BOTH_LIBRARIES ${GTEST_BOTH_LIBRARIES} CACHE INTERNAL "")
+  message(STATUS "Found gmock: gtests will be built")
```
dirk-thomas: Inconsistent message.

cmake/test/gtest.cmake
```diff
@@ -85,97 +85,261 @@ function(catkin_add_executable_with_gtest target)
...
+      set(GMOCK_LIBRARIES ${GMOCK_FROM_SOURCE_LIBRARIES})
+      set(GMOCK_MAIN_LIBRARIES ${GMOCK_FROM_SOURCE_MAIN_LIBRARIES})
+      set(GMOCK_BOTH_LIBRARIES ${GMOCK_LIBRARIES} ${GMOCK_MAIN_LIBRARIES})
+
+      set(GTEST_FOUND ${GMOCK_FOUND})
+      set(GTEST_INCLUDE_DIRS ${GMOCK_INCLUDE_DIRS})
+      set(GTEST_LIBRARY_DIRS ${GMOCK_LIBRARY_DIRS})
+      set(GTEST_LIBRARIES ${GMOCK_LIBRARIES})
+      set(GTEST_MAIN_LIBRARIES ${GMOCK_MAIN_LIBRARIES})
+      set(GTEST_BOTH_LIBRARIES ${GMOCK_BOTH_LIBRARIES})
```
dirk-thomas: Why are the gtest variables set to the same values as gmock here?

Same below.

dirk-thomas (commented):
The overall logic seems weird to me:

If gmock is found (https://github.com/ros/catkin/pull/896/files#diff-84d1585b7758a3723f688bcff3f8602cR212) it is being built. But gtest is neither being considered nor being built in this case.

dirk-thomas:
Can you please describe the different use cases you have tried this patch with as well as steps to reproduce these tests.

cmake/test/gtest.cmake
v-lopez: Done in 4d698414adb6801bcbf0668cba56f122b6109d24

cmake/test/gtest.cmake
v-lopez: Fixed in 1d5e21c, when gmock is found and built, so is gtests.

cmake/test/gtest.cmake
v-lopez: GMOCK_ variables include the GTEST headers and libraries as well.

The purpose of most of the code here, is to make sure that both gmock and gtest use the same version and are compiled together if they, that's why catkin_find_gmock_source calls for catkin_find_gtest_source.

Currently catkin_find_gmock_source does not export the GTEST_ variables to the parent scope, if this is needed I can add it.

cmake/test/gtest.cmake
v-lopez: Like above, fixed in 1d5e21c

cmake/test/gtest.cmake
v-lopez: Fixed, it references 1.8 in the github repo.

v-lopez:
For testing this I have created 2 packages that are in this repository: https://github.com/pal-robotics/catkin_gtest_tests.git
I tested them on a Xenial Kinetic docker.

Install libgtest-dev and google-mock and build everything:
```
mkdir -p my_ws/src
cd my_ws/src/
git clone https://github.com/pal-robotics/catkin -b gmock-kinetic
git clone https://github.com/pal-robotics/catkin_gtest_tests.git
cd ..
catkin_make tests
```
It should all work fine, you should be able to run the dummy tests as wells after sourcing your workspace with:
```
rosrun gtest_test gtest_test-test 
rosrun gmock_test gmock_test-test
```


Now remove google-mock, clean workspace and build again
```
sudo apt-get -y remove google-mock
rm -rf build devel
catkin_make tests
```
It should build the gtest package but fail when compiling  the package that requires gmock.


Now remove gtest, I force a removal without removing it's dependencies for convenience
```
sudo dpkg -r --force-depends libgtest-dev
rm -rf build devel
catkin_make tests
```
It should now fail on a call to catkin_gtest_tests

Now add googletest from sources and build again:
```
git clone  https://github.com/google/googletest.git -b release-1.8.0 src/googletest
rm -rf build devel
catkin_make tests
```

Now everything should work again.

cmake/test/gtest.cmake
v-lopez: @dirk-thomas Fixed in 2ccb83bb64feef257c58076be5bc1593ac0e655a as well

cmake/test/gtest.cmake
dirk-thomas: I guess the question (not yet asked / answered) is: how do we expect a user to use this?

Based on the current patch I assume that users should continue to call `catkin_add_gtest` / `catkin_add_executable_with_gtest` and are then able to use `gtest` as well as `gmock` (when available. Is that assumption correct?

I am not sure that is a good way. I would have expected that the existing functions just link against `gtest` (as they did before, without changing). And a new set of functions `catkin_add_gmock` / `catkin_add_executable_with_gmock` would be introduced which links against both `gmock` as well as `gtest`. A user then needs to opt-in explicitly by updating the called CMake function name to use `gmock`.

v-lopez: Yes your understanding of this patch's behavior is correct.

I understand your point, for any pair of libraries that were unrelated I would completely agree. 
In this case gmock is an extension of gtest, that in theory should be harmless. So I would say that if keeping them together keeps things simple and doesn't introduce any side effects (that I've seen) it might be worth it.

dirk-thomas: I am not use if that might not confuse users that `catkin_add_gtest` "sometimes" works with gmock and sometimes doesn't (e.g. when they use different versions of the ROS). ROS packages might start using `gmock` in their tests without realizing that the same API won't work in older ROS distros.

@clalancette @dhood @mikaelarguedas @tfoote @wjwwood Maybe any of you has feedback on this?

wjwwood: +0

I don't have enough context to know if there is a better way to do this. (I might say +1 if I knew there was not for sure)

Developers might run into a mismatch in features (fails to compile in older distros after some changes), but if we document the potential change in behavior, then that's the best we can do. I don't think that risk is enough to avoid the improvement proposed.

dirk-thomas: @wjwwood An option would be to introduce new CMake functions with `gmock` in the name. So users have to actively switch and those are not available in older distros.

mikaelarguedas: IMO there is value in keeping behavior consistent across distros and to let users opt-in to use new features. So I'd be in favor of having new cmake functions for users that want to use gmock. Then I don't have enough knowledge to estimate the risk and could be convinced that the risk is minimal enough for allowing a documented change of behavior

wjwwood: Having new function/macro names sounds fine to me, I didn't realize that was an option. I thought this was a case where the new version of googletesting is just gmock and gtest isn't separate anymore. My confusion.

v-lopez: Ok, I will add `catkin_add_gmock` and `catkin_add_executable_with_gmock`.

Before I continue with it, there will be a lot of similarities between the gtest and gmock functions.
My idea was to make the current 2 functions' code generic in tow new "private" functions, let's say `_catkin_add_google_test` and `_catkin_add_executable_with_google_test`, and have the gtest and gmock functions call this with minimum code duplication.

Would this be your preferred approach? It would make the code a little bit more complex, the other option is to duplicate and modify the code for the gmock versions, and leave the originals as they were.

dirk-thomas: > ... with minimum code duplication.

@v-lopez That sounds good. Whatever avoids code duplication. Either some "private" functions which are used from the "public" ones. Or the `gmock` ones calling the `gtest` ones internally (if that is feasible).

v-lopez: I've made the discussed changes in https://github.com/ros/catkin/pull/896/commits/3c4c7b2326fdec3f480fdc3a5934d59d7f05f1ea

There's one thing that I don't know how to test. All my changes are focused towards the gtest & gmock from sources approach, which as far as I know, is the recommended way of using google test.

But there's a `find_package(GMock QUIET)` and `find_package(GTest QUIET)` that I don't know how should behave, because there's not a unique FindGMock.cmake file, there are many implementations with different behaviours.

dirk-thomas:
I started adding some changes here (using the GitHub web interface) but quickly reached a limit. Therefore I created #897 from a branch in this repo.

cmake/test/gtest.cmake
```diff
@@ -1,4 +1,5 @@
 _generate_function_if_testing_is_disabled("catkin_add_gtest")
+_generate_function_if_testing_is_disabled("catkin_add_gmock")
```
dirk-thomas, line 2: This doesn't work since the first invocation already calls `return`.

I addressed this in https://github.com/ros/catkin/pull/897/commits/aa28d0c8acdb48846cd53945f9e05b7d8ec56fe5.

dirk-thomas:
Closing in favor of #897.

PR #897, opened 2017-10-27
add support for GMock

dirk-thomas:
Builds on top of #896 and addresses some issues.

@v-lopez FYI

### DIFF
--- a/cmake/test/gtest.cmake
+++ b/cmake/test/gtest.cmake
@@ -167,7 +167,7 @@ function(catkin_find_gmock_source include_paths src_paths found base_dir
             NO_DEFAULT_PATH NO_CMAKE_FIND_ROOT_PATH
   )
 
-  # If we found gmock, ensure gtest is contained within it
+  # If we found gmock, set the variables accordingly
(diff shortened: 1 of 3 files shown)


v-lopez:
@dirk-thomas Thanks for the changes, lgtm.

dirk-thomas:
@mikepurvis Would it be possible for you to use this branch for your testing for a little while? That would give us more confidence that it doesn't introduce any regressions...

mikepurvis:
I've switched our distribution to this branch. FYI @tspicer01

mikepurvis:
We're not actively using the new gmock support, but I believe none of our unittests have gone up in flames as a result of this change.

dirk-thomas:
@v-lopez Thank you for providing this. I will go ahead and merge this so it will be part of the next release of this package.

dirk-thomas:
What I didn't realize when review this is that the patch duplicates large blocks of the logic and copy-n-pastes it into the new function `catkin_find_gmock_source`. In light of this patch failing to build for melodic (which has the `googletest` 1.8.0) (https://github.com/ros/catkin/issues/910#issuecomment-359968542) this should not only be fixed but also be cleaned up.

Issue #901, opened 2017-11-27
[catkin_prepare_release] UnicodeEncodeError: 'ascii' codec can't encode character u'\xf6' in position 1592: ordinal not in range(128)


130s:
```
$ git clone git@github.com:ros-planning/geometric_shapes.git && cd geometric_shapes/
$ catkin_prepare_release 
Prepare the source repository for a release.
Repository type: git
Found packages: geometric_shapes
Prepare release of version '0.5.3' [Y/n]?
Trying to push to remote repository (dry run)...
Everything up-to-date
Checking if working copy is clean (no staged changes, no modified files, no untracked files)...
Traceback (most recent call last):
  File "/opt/ros/kinetic/bin/catkin_prepare_release", line 390, in <module>
    main()
  File "/opt/ros/kinetic/bin/catkin_prepare_release", line 319, in main
    update_changelog_sections(missing_changelogs_but_forthcoming, new_version)
  File "/opt/ros/kinetic/lib/python2.7/dist-packages/catkin/package_version.py", line 131, in update_changelog_sections
    f.write(data)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xf6' in position 1592: ordinal not in range(128)
$ git log -n 1
commit 70262c6303f8e1ae348b92db1aca588c61890b6e
$ which  catkin_prepare_release
/opt/ros/kinetic/bin/catkin_prepare_release
$ apt-cache policy ros-kinetic-catkin
ros-kinetic-catkin:
  Installed: 0.7.6-0xenial-20170217-212821-0800
  Candidate: 0.7.6-0xenial-20170217-212821-0800
```

`xf6` in unicode seems to be [o + umlaut](http://www.utf8-chartable.de/unicode-utf8-table.pl?names=-&unicodeinhtml=hex), which occurs in [CHANGELOG.rst](https://github.com/ros-planning/geometric_shapes/blob/70262c6303f8e1ae348b92db1aca588c61890b6e/CHANGELOG.rst).

bmagyar:
I second that.

Just encounterd this today with `u"\xfc"` corresponding to `ü`.

mikaelarguedas:
I faced this issue too and the patch in #902 fixed it, allowing me to release

dirk-thomas:
Addressed by #902.

fmessmer:
I'm still getting this error...for me #902 only fixed the exception during the `catkin_generate_changelog`-step

PR #902, opened 2017-11-28
Fix Unicode write error.

mikeodr:
Make writing data back out encoded to a unicode supported format.
This prevents unicode exceptions:
`UnicodeEncodeError: 'ascii' codec can't encode character u'\u2026' in position 1149: ordinal not in range(128)`

### DIFF
--- a/test/unit_tests/test_package_version.py
+++ b/test/unit_tests/test_package_version.py
@@ -2,9 +2,11 @@ import os
 import unittest
 import tempfile
 import shutil
+import mock
+import datetime
 
 try:
-    from catkin.package_version import _replace_version, update_versions
+    from catkin.package_version import _replace_version, update_versions, update_changelog_sections
 except ImportError as impe:
     raise ImportError(
         'Please adjust your pythonpath before running this test: %s' % str(impe))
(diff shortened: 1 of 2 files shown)


mikeodr:
Was releasing a package and ran into the following:
```
catkin_prepare_release --bump minor
Prepare the source repository for a release.
Repository type: git
Found packages: REDACTED
Prepare release of version 'x.x.x' [Y/n]?
Trying to push to remote repository (dry run)...
To REDACTED
  xxxxxxxx..xxxxxxxx  devel -> devel
Checking if working copy is clean (no staged changes, no modified files, no untracked files)...
  File "/path/to/catkin_prepare_release", line 393, in <module>
    main()
  File "/path/to/catkin_prepare_release", line 322, in main
    update_changelog_sections(missing_changelogs_but_forthcoming, new_version)
  File "/path/to/catkin/package_version.py", line 131, in update_changelog_sections
    f.write(data)
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2026' in position 1149: ordinal not in range(128)```

**Comment by mikepurvis** [comment]
Is the fix verified to do the right thing under both Python 2 and Python 3?

**Comment by mikeodr** [comment]
Travis CI unittest for py3 passes.

**Comment by mikeodr** [comment]
Additionally spawning a `python3` interpreter and doing:

```
with tempfile.NamedTemporaryFile() as f:
...     f.write('hello world'.encode('utf-8'))
... 
11
``` 

shows no errors.

**Comment by mikepurvis** [comment]
But the unit tests passed prior to this change as well, right? So unless we were adding a test which exposed the particular bug seen, a CI pass isn't super meaningful.

**Comment by iwanders** [comment]
Try with a minimum (non) working example, example character here is not a unicode character, but an extended ascii one, same result though. Fails in both Python 2 and Python 3:
```python
import tempfile
with tempfile.NamedTemporaryFile() as f:
    f.write(u"\xc3")
```
Adding the encoding fixes the issue for both:
```python
import tempfile
with tempfile.NamedTemporaryFile() as f:
    f.write(u"\xc3".encode('utf-8'))
```
I think it stems from the default file mode [`tempfile`](https://docs.python.org/2.7/library/tempfile.html) uses; `mode='w+b'`, which will make it work with Python 2 if there are no special characters in the string (it will get treated as a byte string). On Python 3 the original code always causes error, since you are required to write bytes to the opened file and not a unicode string?

dirk-thomas (dismissed):
The current patch fails with Python 3 since the file handle is currently opened in text mode (`'w'`) but the modified code tries to write `bytes`.

This can be easily verified by running these two lines in a Python interpreter:

```
h = open('tmp', 'w')
h.write('hello world'.encode('utf-8'))
```

The solution is probably to open the file in binary mode (`wb`). Please update the patch accordingly and make sure to actually run the code with both Python 2 as well as 3 interpreters before (expected to fail) and after (working with both) applying the patch.

dirk-thomas:
For the record: this should address #901.

mikeodr:
Added a unit test to catch for the exception.
Updated to `wb` for file write.

Passing in travis 2.7 and 3.x

test/unit_tests/test_package_version.py
```diff
@@ -2,9 +2,10 @@
 import unittest
 import tempfile
 import shutil
+import mock
```
mikepurvis, line 5: Not in Python2 standard library, but this package already depends on `python-mock`, so 👍 .

mikeodr: Yeah I checked for that and saw it in the package.xml so went with it.

test/unit_tests/test_package_version.py
```diff
@@ -40,3 +41,30 @@ def test_update_versions(self):
...
+                                  '* This is a line that has unicode' u'\xfc''\n'
+                                  '\n'
+                                  '0.0.9 (2017-01-30)\n'
+                                  '------------------\n'
+                                  '* This is old version.\n')
+
+            # Create tuple with expected entires.
+            missing_changelogs_but_forthcoming['fake_pkg'] = (temp_file.name, mock_changelog, 'Forthcoming')
+            # Should not raise an exception
+            update_changelog_sections(missing_changelogs_but_forthcoming, '1.0.0')
```
mikepurvis, line 71: Worthwhile to actually validate the output since you're here anyway?

Either way, LGTM.

mikeodr: I can, but the issue here would be adding a sleep to the test.

I've tried flushing the file in the past and calling the system flush, but still end up with a race condition where the file hasn't finished populating to disk yet.

I'm open to suggestions.

mikeodr: Doesn't seem to be effected by disk buffer issues on my PC, added a read-back assertion.

mikeodr: Readback test added, with python2 & 3 working.

130s:
I confirmed that this fixes #901 for moveit (https://github.com/ros-planning/moveit/commit/4bce541153b26c184cb5d2a335fb894d744b1732 is made with this patch).

test/unit_tests/test_package_version.py
```diff
@@ -40,3 +42,57 @@ def test_update_versions(self):
...
+                        b'* This is my changelog entry',
+                        unicode_line,
+                        b'',
+                        b'0.0.9 (2017-01-30)',
+                        b'------------------',
+                        b'* This is old version.']
+
+            # Open the file written, and compare each line written to
+            # the one read back.
+            content = open(temp_file.name, 'rb').read().split(b'\n')
```
dirk-thomas: The file handle is not being closed. Please use `with` to scope the handle. Additionally use `splitlines()` rather than a manual split. E.g.

```
with open(temp_file.name, 'rb') as h:
    content = h.read()
lines = content.splitlines()
```

dirk-thomas:
Thank you for the patch and for iterating on it.

fmessmer:
seems this fixes the `catkin_generate_changelog`-step
but I'm still getting a similar Unicode exception during the `catkin_prepare_release`-step when proceeding in the release process
see #908

PR #906, opened 2017-12-15
cleanup environment changes in tests reliably

dirk-thomas:


at-wat:
`PATH` environment variable seems not set inside `nosetests test/unit_tests test/local_tests`. 
(KeyError occurs if I add `os.environ['PATH']` in test script.)
But,
```
 script:
   - echo -e "import os\nprint(os.environ['PATH'])" | python
```
in travis.yml returns correct PATH.

PR #907, opened 2017-12-15
Provide default values to unbound variables

### LINKED ISSUE #904: Unbound variable CATKIN_SETUP_UTIL_ARGS in setup.sh
https://github.com/ros/catkin/blob/3519c73394d2121b85ddbee488e4c44d4d942118/cmake/templates/setup.sh.in#L60

Sourcing setup.sh fails with `set -eu` since `CATKIN_SETUP_UTIL_ARGS` is not set by default.
This causes a problem in CI environment in my use case.

It should be better to set a default value to the variable.

at-wat:
This commit provides default values to unbound variables in setup.sh.in.
This change makes `source setup.sh` work with `set -u`.

---
rebased version of #905 
fixes #904

### DIFF
--- a/cmake/templates/setup.sh.in
+++ b/cmake/templates/setup.sh.in
@@ -46,7 +46,7 @@ fi
 
 # invoke Python script to generate necessary exports of environment variables
 # use TMPDIR if it exists, otherwise fall back to /tmp
-if [ -d "${TMPDIR}" ]; then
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thank you for the patch.

at-wat:
Thank you!

PR #914, opened 2018-01-24
Add GTest 1.8 compatibility and reduce duplicated code

v-lopez:


cmake/test/gtest.cmake
```diff
@@ -389,13 +325,12 @@ if(NOT GMOCK_FOUND)
             endif()
             _install(${ARGN})
           endfunction()
-          add_subdirectory(${gtest_base_dir} ${gtest_lib_dir})
+          add_subdirectory(${base_dir} ${gtest_lib_dir})
           set(_CATKIN_SKIP_INSTALL_RULES FALSE)
           set_target_properties(${gtest_libs} ${gtest_main_libs}
                                 PROPERTIES EXCLUDE_FROM_ALL 1)
```
dirk-thomas, line 331: This property also needs to be set for the gmock targets.

dirk-thomas (commented):
Thank you for the very quick PR.

After reviewing the patch I only have one comment.

cmake/test/gtest.cmake
v-lopez: You are right, I have ammended the commit to include this change.

dirk-thomas:
This looks good in a small workspace on Melodic. Thank you again for the patch!

cmake/test/gtest.cmake
```diff
@@ -178,127 +178,112 @@ function(_catkin_add_executable_with_google_test type target)
...
-      set(${gtest_libs} ${_gtest_libs} PARENT_SCOPE)
-      set(${gtest_main_libs} ${_gtest_main_libs} PARENT_SCOPE)
-    else()
-      message(WARNING "Found gmock, but it did not contain gtest! Please ensure gmock is installed correctly.")
-    endif()
+  _catkin_find_google_source("gmock" "${_gmock_include_paths}"
+                             "${_gmock_source_paths}" _gmock_found
+                             _gmock_base_dir _gmock_include_dir _gmock_lib_dir
+                             _gmock_libs _gmock_main_libs)
+  if(gmock_found)
```
dirk-thomas, line 269: I think this line is incorrect. See #919.

PR #916, opened 2018-01-25
Prevent reading non-whitelisted properties of interface targets.

de-vri-es:
Currently, `catkin_replace_imported_library_targets(...)` can attempt to read non-whitelisted properties of imported interface libraries. After a recent update on Arch Linux this caused atleast `pcl-ros` to fail at the cmake step with the below error.

My guess is that this is new behaviour introduced in cmake 3.10.2 for which I couldn't find release notes yet, or otherwise caused by an interface library now showing up in `PCL_LIBRARIES`.

This PR fixes the issue by detecting interface libraries and replacing them with their `INTERFACE_LINK_LIBRARIES` property instead of `IMPORTED_LOCATION`.

On an unrelated note, imported library targets may also have `IMPORTED_LINK_INTERFACE_LIBRARIES(_$CONFIG)` which don't seem to be handled. If that is desired I'd prefer to keep it out of this PR though.

Error message without this PR running cmake on `pcl-ros`:
```
CMake Error at /opt/ros/lunar/share/catkin/cmake/catkin_libraries.cmake:137 (get_target_property):
  INTERFACE_LIBRARY targets may only have whitelisted properties.  The
  property "IMPORTED_LOCATION" is not allowed.
  /opt/ros/lunar/share/catkin/cmake/catkin_package.cmake:258 (catkin_replace_imported_library_targets)
  /opt/ros/lunar/share/catkin/cmake/catkin_package.cmake:102 (_catkin_package)
  CMakeLists.txt:63 (catkin_package)


CMake Error at /opt/ros/lunar/share/catkin/cmake/catkin_libraries.cmake:143 (get_target_property):
  INTERFACE_LIBRARY targets may only have whitelisted properties.  The
  property "IMPORTED_LOCATION_vtkm_imported_configurations-NOTFOUND" is not
  allowed.
  /opt/ros/lunar/share/catkin/cmake/catkin_package.cmake:258 (catkin_replace_imported_library_targets)
  /opt/ros/lunar/share/catkin/cmake/catkin_package.cmake:102 (_catkin_package)
  CMakeLists.txt:63 (catkin_package)


CMake Error at /opt/ros/lunar/share/catkin/cmake/catkin_libraries.cmake:137 (get_target_property):
  INTERFACE_LIBRARY targets may only have whitelisted properties.  The
  property "IMPORTED_LOCATION" is not allowed.
  /opt/ros/lunar/share/catkin/cmake/catkin_package.cmake:258 (catkin_replace_imported_library_targets)
  /opt/ros/lunar/share/catkin/cmake/catkin_package.cmake:102 (_catkin_package)
  CMakeLists.txt:63 (catkin_package)


CMake Error at /opt/ros/lunar/share/catkin/cmake/catkin_libraries.cmake:143 (get_target_property):
  INTERFACE_LIBRARY targets may only have whitelisted properties.  The
  property "IMPORTED_LOCATION_vtkm_imported_configurations-NOTFOUND" is not
  allowed.
  /opt/ros/lunar/share/catkin/cmake/catkin_package.cmake:258 (catkin_replace_imported_library_targets)
  /opt/ros/lunar/share/catkin/cmake/catkin_package.cmake:102 (_catkin_package)
  CMakeLists.txt:63 (catkin_package)


-- Configuring incomplete, errors occurred!
See also "/tmp/makepkg/ros-lunar-pcl-ros/src/build/CMakeFiles/CMakeOutput.log".
See also "/tmp/makepkg/ros-lunar-pcl-ros/src/build/CMakeFiles/CMakeError.log".
```


de-vri-es:
The issue was most likely triggered by an update to VTK 8.1, and not by a cmake update. VTK 8.1 now exports an interface target called `vtkm` that PCL re-exports.

de-vri-es:
Fixed property name: `INTERFACE_LINK_LIBRARIES` instead of `INTERFACE_LIBRARIES`.

de-vri-es:
I also had to recursively resolve any imported/interface targets in `INTERFACE_LINK_LIBRARIES`. Otherwise dependent packages will fail on the cmake step, since some exported libraries might still be imported targets.

That does mean that cycles could cause infinite recursion. If that happens, it would appear that cmake segfaults rather than hitting some recursion limit.

cmake/catkin_libraries.cmake
```diff
@@ -131,8 +131,15 @@ function(catkin_replace_imported_library_targets VAR)
       # sometimes cmake dependencies define imported targets, in which
       # case the imported library information is not the target name, but
       # the information embedded in cmake properties inside the imported library
+      # For interface libraries, this is the INTERFACE_LINK_LIBRARIES property.
+      # For regular imported libraries, this is IMPORTED_LOCATION(_$CONFIG).
+      get_target_property(${lib}_type ${lib} TYPE)
       get_target_property(${lib}_imported ${lib} IMPORTED)
-      if(${${lib}_imported})
+      if(${${lib}_type} STREQUAL "INTERFACE_LIBRARY")
```
dirk-thomas, line 138: I couldn't find any documentation about the "string" `INTERFACE_LIBRARY`.

Maybe `get_property(... SET/DEFINED)` could be used instead if this can't be relied on?

de-vri-es: Documentation is here: https://cmake.org/cmake/help/v3.1/prop_tgt/TYPE.html

It first appeared in cmake 3.1, but since older cmakes will never report `"INTERFACE_LIBRARY"`, it is automatically backwards compatible.

dirk-thomas: Ok, thanks for the reference. I was indeed looking at the 3.0 docs.

dirk-thomas:
Thanks for the patch.

PR #917, opened 2018-01-26
Fix add_dependencies() to refer to targets instead of libraries.

de-vri-es:
`add_dependencies(...)` is for cmake targets, not for libraries. I believe that the intended behaviour is what this PR does.

Without this PR, adding a gtest results in an error like this if gtest is already pre-built:
```
CMake Error at /opt/ros/lunar/share/catkin/cmake/test/gtest.cmake:178 (add_dependencies):
  The dependency target "/usr/lib/libgtest.so" of target
  "dr_camera_parameters_test_camera_parameters" does not exist.
  /opt/ros/lunar/share/catkin/cmake/test/gtest.cmake:79 (_catkin_add_executable_with_google_test)
  /opt/ros/lunar/share/catkin/cmake/test/gtest.cmake:28 (_catkin_add_google_test)
  /home/maarten/dev/ros/base/src/dr_base/dr_cmake/cmake/dr_cmake.cmake:38 (catkin_add_gtest)
  CMakeLists.txt:45 (dr_add_gtest)
```

### DIFF
--- a/cmake/test/gtest.cmake
+++ b/cmake/test/gtest.cmake
@@ -173,9 +173,6 @@ function(_catkin_add_executable_with_google_test type target)
 
   assert(${type_upper}_LIBRARIES)
   target_link_libraries(${target} ${${type_upper}_LIBRARIES} ${THREADS_LIBRARY})
-
-  # make sure gtest/gmock is built before the target
-  add_dependencies(${target} ${${type_upper}_LIBRARIES})
 endfunction()
 
 # Internal function for finding gtest or gmock sources


cmake/test/gtest.cmake
```diff
@@ -175,7 +175,7 @@ function(_catkin_add_executable_with_google_test type target)
   target_link_libraries(${target} ${${type_upper}_LIBRARIES} ${THREADS_LIBRARY})
 
   # make sure gtest/gmock is built before the target
-  add_dependencies(${target} ${${type_upper}_LIBRARIES})
+  add_dependencies(${target} ${type} ${type}_main)
```
dirk-thomas: I don't like the fact that this duplicates the logic about the `_main` suffix. That should not be repeated here - there should be a variable instead which contains the necessary information.

dirk-thomas (commented):
Can you please include a reproducible example.

de-vri-es:
Minimal example here: https://github.com/de-vri-es/catkin_gtest_kapot

Without this PR, it results in this error during the cmake stage. This is on Arch Linux with gtest installed (but not gmock) as system package, with the libraries already compiled.
```
-- Found PythonInterp: /usr/bin/python (found version "3.6.4") 
-- Using PYTHON_EXECUTABLE: /usr/bin/python
-- Using default Python package layout
-- Found PY_em: /usr/lib/python3.6/site-packages/em.py  
-- Using empy: /usr/lib/python3.6/site-packages/em.py
-- Using CATKIN_ENABLE_TESTING: ON
-- Call enable_testing()
-- Using CATKIN_TEST_RESULTS_DIR: /home/maarten/dev/catkin_gtest_kapot/build/test_results
-- Found gtest: gtests will be built
-- Using Python nosetests: /usr/bin/nosetests-3.6
-- catkin 0.7.10
-- Configuring done
CMake Error at /opt/ros/lunar/share/catkin/cmake/test/gtest.cmake:178 (add_dependencies):
  The dependency target "/usr/lib/libgtest.so" of target
  "catkin_gtest_kapot_test_boolean" does not exist.
  /opt/ros/lunar/share/catkin/cmake/test/gtest.cmake:79 (_catkin_add_executable_with_google_test)
  /opt/ros/lunar/share/catkin/cmake/test/gtest.cmake:28 (_catkin_add_google_test)
  CMakeLists.txt:7 (catkin_add_gtest)


-- Generating done
-- Build files have been written to: /home/maarten/dev/catkin_gtest_kapot/build
```

/edit: briefly posted the wrong error =]

dirk-thomas:
Thank you for the additional information. I won't be able to reproduce the problem since Ubuntu follows the recommendation og `googletest` to not ship prebuild libraries.

Do you see a way to avoid the duplication of the target names?

de-vri-es:
Well, if really desired the two can be combined in new variables `gtest_TARGETS` and `gmock_TARGETS`.

But I wonder if this is needed at all. CMake already adds the dependencies automatically if you link against a library built in the same project (including subdirectories). So a much nicer fix is to simply drop the whole call.

From the docs:
> If a library name matches that of another target in the project a dependency will automatically be added in the build system to make sure the library being linked is up-to-date before the target links.

https://cmake.org/cmake/help/v3.0/command/target_link_libraries.html

de-vri-es:
Updated the PR to simply drop the call. I strongly believe it was redundant in the cases where it didn't trigger an error.

dirk-thomas:
Good question why it has been added in the first place: https://github.com/ros/catkin/commit/38eb9aaf8ef4918c02e66401d3a56915ae8d1536#diff-84d1585b7758a3723f688bcff3f8602cR43 :thinking:

The CMake docs clearly state:

> If a library name matches that of another target in the project a dependency will automatically be added in the build system to make sure the library being linked is up-to-date before the target links.

So I agree with your proposal to just drop the call. Thank you for iterating on this. I think this can be safely merged.

de-vri-es:
Thanks =)

PR #918, opened 2018-01-26
add option to pass specific version to catkin_prepare_release

dirk-thomas:
The new option allows to specify arbitrary versions in case bumping major/minor/patch isn't sufficient.

### DIFF
--- a/bin/catkin_prepare_release
+++ b/bin/catkin_prepare_release
@@ -3,6 +3,7 @@
 from __future__ import print_function
 import argparse
 import os
+import re
 import subprocess
 import sys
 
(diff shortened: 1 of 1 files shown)


sloretz (approved):
I tested up to where it wants to push to the remote repo with and without --version specified. Does anything after catkin_prepare_release force a major.minor.patch format for the version? The script will happily accept any string at the moment.

dirk-thomas:
> Does anything after catkin_prepare_release force a major.minor.patch format for the version? The script will happily accept any string at the moment.

Correct, no enforcement. The caller is in charge of its own fate :wink:

dirk-thomas:
In expectation of the "worst" I added the same regex check from `catkin_pkg`: 65dc6fb6b2c192939133066553a878334451d391

sloretz:
65dc6fb works for me. It would be more DRY to add and import a `check_version` function from `catkin_pkg.package_version`, but as it's just one line I'm fine with it as it.

PR #919, opened 2018-01-30
fix logic when only gtest is present

dirk-thomas:
@v-lopez the current code doesn't work when only `gtest` is present but not `gmock`.

I think the fixed variable name of this patch should address the problem. I would appreciate if you could test the branch in various combinations to make sure it doesn't introduce any regressions when being released into Kinetic / Lunar.


v-lopez:
Nice catch.

I've tried to build actionlib tests without your patch, only with gtest in kinetic and it fails, with gmock it doesn't.
With your patch both versions work.

On lunar, even though gmock is embedded in the googletest package, there is still a google-mock debian for backwards compatibility. Without this debian installed and with/without your patch it still works.

dirk-thomas:
Thank you for the extra pair of eyes and testing. :bowing_man:

I will do another release with this patch then.

PR #924, opened 2018-03-12
Remove Python 3.3 specific code because it is end of life

cclauss:
Remove the special case that was leftover for Python 3.3 which is [EOL](https://devguide.python.org/#branchstatus).


dirk-thomas:
Thanks.

PR #927, opened 2018-03-20
Do not add_library() gmock and gtest if targets already exist

ojura:
Fixes build problems with Google Cartographer (https://github.com/googlecartographer/cartographer_ros/issues/776), which provides its own `FindGMock.cmake`.


dirk-thomas:
And I was just about to ask for the `gtest` case. But here it is already: 54fdf041bea2dfe31f5bdbe54a01b118660a3441

Thank you for the patch.

ojura:
@dirk-thomas is there a chance you could make a release? CI for cartographer_ros is currently kind of stuck because of this.

dirk-thomas:
A release especially in ROS distros with a lot of packages is a significant effort (requiring to rebuild all packages across all targeted platforms as well as all users downloading new Debian packages). Therefore low level packages like `catkin` are not released very often. What ROS distro are you targeting `cartographer_ros` for?

ojura:
`cartographer_ros` targets indigo, kinetic and lunar.

dirk-thomas:
The `gmock` support is not part of Indigo (we usually don't backport new features in to older ROS distros) so this change doesn't apply to the `indigo-devel` branch. You might want to test Indigo separately.

For Kinetic and higher and will try to get this released but it might be a couple of weeks from now.

tfoote:
@ojura There are other blockers for older targets due to incompatible versions of ceres. https://github.com/googlecartographer/cartographer_ros/issues/158 I believe that there is a goal to get it working on Kinetic if we exclude Jessie, but won't be able to get it onto indigo target platforms. @mikaelarguedas FYI

ojura:
@tfoote Sure. I am not a Cartographer maintainer though, just a third-party contributor. So, I am not involved much with releasing to ROS. My interest here was to get `cartographer_ros`'s CI unblocked so we can merge PRs which have piled up.

mikaelarguedas:
> There are other blockers for older targets due to incompatible versions of ceres. googlecartographer/cartographer_ros#158 I believe that there is a goal to get it working on Kinetic if we exclude Jessie, but won't be able to get it onto indigo target platforms

Regarding debs:
I haven't tried cartographer 0.3.0 on kinetic/lunar yet, but I expect the protobuf version to be a blocker. to release anything later than 0.2.0 in these distros. I don't think this PR is necessary to release 0.2.0 in kinetic/lunar.

0.3.0 has been released successfully in melodic (https://github.com/ros/rosdistro/pull/17242), but this PR will be required to enable the release of `cartographer_ros`.

ojura:
We have just merged a mitigation for this in libcartographer (https://github.com/googlecartographer/cartographer/pull/1011), so it's not a pressing issue anymore for cartographer_ros master :)

mikaelarguedas:
thanks @ojura for the pointer to the mitigation PR  :+1:, I'll use that for releasing pending the release of this fix

PR #928, opened 2018-03-23
Modernize Python 2 code to get ready for Python 3

cclauss:
Make the minimal, safe changes required to convert the repo's code to be syntax compatible with both Python 2 and Python 3.  There may be other changes required to complete a port to Python 3 but this PR is a minimal, safe first step.

Run: [futurize --stage1 -w **/*.py](http://python-future.org/futurize_cheatsheet.html)

See Stage 1: "safe" fixes http://python-future.org/automatic_conversion.html#stage-1-safe-fixes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,6 @@ python:
   - "3.4_with_system_site_packages"
 # command to install dependencies
 install:
-  # fetch "normal" debian python for 3.3
-  - if [ "$TRAVIS_PYTHON_VERSION" == "3.3" ]; then sudo add-apt-repository -y ppa:fkrull/deadsnakes && sudo apt-get -y update && sudo apt-get install python3.3 python3.3-dev && virtualenv -p /usr/bin/python3.3 ~/virtualenvs/3.3_debian && source ~/virtualenvs/3.3_debian/bin/activate; fi
   - pip install argparse catkin-pkg empy mock nose
 # command to run tests
 script:
--- a/test/checks/fuerte/distro-to-install.py
+++ b/test/checks/fuerte/distro-to-install.py
@@ -5,13 +5,12 @@ import sys
 
 def go(argv):
   if len(argv) != 2:
-    print 'wrong number args'
+    print('wrong number args')
     sys.exit(1)
 
   d=rospkg.distro.load_distro(argv[1])
   ri=rospkg.distro.distro_to_rosinstall(d, 'release')
-  print ri
+  print(ri)
 
 if __name__ == '__main__':
   go(sys.argv)
-
--- a/test/local_tests/test_with_mock_workspace.py
+++ b/test/local_tests/test_with_mock_workspace.py
@@ -74,7 +74,7 @@ class MockTest(AbstractCatkinWorkspaceTest):
 
         out = self.cmake(CMAKE_PREFIX_PATH=self.installdir,
                          expect=fail)
-        print("failed as expected, out=", out)
+        print("failed as expected, out={}".format(out))
 
         self.assertTrue(b"catkin_package() PROJECT_NAME is set to 'Project'" in out, out)
         # assert 'You must call project() with the same name before.' in out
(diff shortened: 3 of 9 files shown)


test/checks/fuerte/distro-to-install.py
```diff
@@ -1,16 +1,17 @@
 #!/usr/bin/env python
 
+from __future__ import print_function
```
dirk-thomas: This is not necessary in this case (see http://wiki.ros.org/python_2_and_3_compatible_code#Print_is_now_a_function).

Same below for many other cases.

test/local_tests/test_with_mock_workspace.py
```diff
@@ -74,7 +75,7 @@ def test_noproject(self):
 
         out = self.cmake(CMAKE_PREFIX_PATH=self.installdir,
                          expect=fail)
-        print("failed as expected, out=", out)
+        print(("failed as expected, out=", out))
```
dirk-thomas: The extra parenthesis don't seem to make any sense. Why are they being introduced?

dirk-thomas (changes requested):

cclauss:
The import keeps Python 2 developers from breaking Python 3 code because when the import is used then print without the parens is a syntax error in both Python 2 and Python 3.  The import also deals gracefully with commas inside a print statement thus the double parens for trying to preserve the output...

$ python2
```
Python 2.7.14 ....
>>> print 0, 1
0 1
>>> print(0, 1)
(0, 1)
>>> print((0, 1))
(0, 1)
>>> from __future__ import print_function
>>> print 0, 1
  File "<stdin>", line 1
    print 0, 1
          ^
SyntaxError: invalid syntax
>>> print(0, 1)
0 1
>>> print((0, 1))
(0, 1)
```

test/network_tests/test_basics.py
```diff
@@ -1,11 +1,12 @@
 #!/usr/bin/env python
 
+from __future__ import absolute_import
```
dirk-thomas: Why is this needed here?

Same question below.

dirk-thomas:
> The import keeps Python 2 developers from breaking Python 3 code because when the import is used then print without the parens is a syntax error in both Python 2 and Python 3. The import also deals gracefully with commas inside a print statement thus the double parens for trying to preserve the output...

I understand. The wiki page I referenced mentions that. I just don't think it should be used when not necessary.

Wasn't there one occasion where print is called with multiple arguments? In that the future import would make sense. But I can't see it anywhere anymore since you force pushed.

.travis.yml
```diff
@@ -4,7 +4,12 @@ python:
   - "3.4_with_system_site_packages"
 # command to install dependencies
 install:
-  - pip install argparse catkin-pkg empy mock nose
+  - pip install argparse catkin-pkg empy flake8 mock nose
+before_script:
+  # stop the build if there are Python syntax errors or undefined names
+  - flake8 . --count --select=E9,F821,F822,F823 --show-source --statistics
+  # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
+  - flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
```
dirk-thomas: Please move the check into a unit test so that it is being run locally too.

cclauss:
> Wasn't there one occasion where print is called with multiple arguments?
https://github.com/ros/catkin/blob/kinetic-devel/test/local_tests/test_with_mock_workspace.py#L77

test/network_tests/test_basics.py
cclauss: https://www.python.org/dev/peps/pep-0328/#rationale-for-absolute-imports

dirk-thomas:
> https://github.com/ros/catkin/blob/kinetic-devel/test/local_tests/test_with_mock_workspace.py#L77

For that line future import would make sense. I doubt the intention in Python 2 was to print parenthesis around the arguments.

cclauss:
I used __str.format()__ instead of the print_function import to make the intention explicit.

dirk-thomas (approved):

dirk-thomas:
Looks good to me now. Thank you for creating the patch and iterating on it.

PR #929, opened 2018-03-30
add CMake option for symlink install

dirk-thomas:
In the discussion about the unified build tool the question came up if the devel space should be continued to be supported. In ROS 2 it has intentionally been removed due to the extra complexity and custom logic which have to be put into every package to support that custom concept. Instead ROS 2 offers a "symlink install" option which replaces most CMake `install()` calls with symlinks on platforms which support that. In the case we want to do the same in the future in ROS 1 I ported the symlink install logic from `ament_cmake_core` to `catkin`.


NikolausDemmel:
Interesting.

Assuming that the usual development-workspace workflow for catkin currently is to only work with the devel space, this change would allow to switch to symlink-install for development and use the install space instead, right? The devel space would probably still be there (e.g. for generated files, right?), but not directly used.

Are there any downsides of the symlink-install vs the devel workflow for development as far as you can tell?

dirk-thomas:
> Assuming that the usual development-workspace workflow for catkin currently is to only work with the devel space, this change would allow to switch to symlink-install for development and use the install space instead, right?

Yes, the idea is that users would always use the install option and optional use symlink install to have the same benefit as the devel space.

> The devel space would probably still be there (e.g. for generated files, right?), but not directly used.

Yes, if you don't specify an explicit location the `devel` directory is actually created as a subdirectory of `build` so it wouldn't be "directly visible".

> Are there any downsides of the symlink-install vs the devel workflow for development as far as you can tell?

None, as far as I know. But the advantage is that the CMake code in ROS packages wouldn't need to handle the devel case separately. E.g. no need for a separate `CFG_EXTRAS` for `*.develspace.*`.

NikolausDemmel:
Cool, so there could be a long term goal of phasing out the devel space. But it would have to be really long term to not be disruptive. From my experience people working with non-released packages often don't bother to define install rules if they only work with the devel space anyway. If phasing out the devel space in ROS 1 is even desirable, one would need to make a convincing argument why it is *necessary*.

In any case, it would be great to get support for this in catkin_tools as well once it is merged here.

wjwwood (approved):
👍, I'm very much in favor of back porting this feature and recommending that people use install rules in all packages and then prefer install space with symlink install over the devel space when developing their code.

I share the concern of @NikolausDemmel, which is (hopefully I paraphrase correctly) that it let's us _plan_ for a tick-tock style removal of the devel space in ROS 1, but it does not make it obvious that we _should_ remove the devel space. We need to discuss that with the community in addition to having this feature before we decide. But I think that's what @dirk-thomas was saying too.

However, if we add the option and advertise it, most everyone might switch over to the install space only on the their own accord, and at that point it would be easier to justify going ahead and removing it. So the early the better on this feature, imo.

NikolausDemmel:
> I share the concern of @NikolausDemmel, which is (hopefully I paraphrase correctly) that it let's us plan for a tick-tock style removal of the devel space in ROS 1, but it does not make it obvious that we should remove the devel space. We need to discuss that with the community in addition to having this feature before we decide.

This is exactly what I mean, just in much clearer words :+1:.

PR #930, opened 2018-04-04
fix an error in the comment.

csukuangfj:


dirk-thomas:
Thank you for the fix.

PR #932, opened 2018-04-05
pkgConfig.cmake: add lib/x86_64-linux-gnu to search path

po1:
This search path is added to the run-time environment, but
is missing from the build-time environment, which results in
binaries not linking properly, or linking against the wrong
version of a library.

Using the GNUInstallDirs CMake module:
https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html

### DIFF
--- a/cmake/catkin_package.cmake
+++ b/cmake/catkin_package.cmake
@@ -80,6 +80,9 @@
 #
 # @public
 #
+
+include(GNUInstallDirs)
+
 macro(catkin_package)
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Can you please describe a reproducible use case which demonstrates how this patch is used.

dirk-thomas:
Closing due to no response. Please feel free to comment with the requested information and the ticket can be reopened.

PR #941, opened 2018-05-30
move catkin_prepare_release script as well as dependencies to catkin_pkg

dirk-thomas:
Removed the following scripts:
* `catkin_package_version`
* `catkin_prepare_release`

Replace the following modules with a "redirect" to the new location in `catkin_pkg`:
* `package_version`
* `terminal_color`
* `workspace_vcs`

The patch also bumps the minimum version of `catkin_pkg` to the upcoming version 0.4.3 which will contain ros-infrastructure/catkin_pkg#216.

PR #943, opened 2018-06-06
terminal_color is now in catkin_pkg

jspricke:


dirk-thomas:
Thank you for the fix. Fixes a regression of #941. Needs an immediate release into Melodic.

PR #954, opened 2018-08-24
Use CATKIN_GLOBAL_ETC_DESTINATION instead of etc

jspricke:


dirk-thomas:
Thanks for the patch.

PR #955, opened 2018-08-28
Fix parallel builds

rhaschke:
This PR addresses an issue originally raised in https://github.com/catkin/catkin_tools/issues/422, but which is actually related to the underlying catkin cmake files.
* 1dbc0ad addresses the original issue of cmake's `file RENAME` only working within a filesystem. It fails when trying to move a file _across_ file systems. However, probably the operation within a single filesystem is what makes the operation atomic in cmake.
* To counteract this regression, 6787656 also adds a `file LOCK` on the destination folder as suggested by @guihomework in https://github.com/catkin/catkin_tools/issues/422#issuecomment-414774603. This additionally avoids issues of several parallel catkin cmake processes writing the same environment files simultaneously (which is a real issue for our jenkins setup deploying directly into a shared install-space.)


cmake/atomic_configure_file.cmake
```diff
@@ -1,9 +1,9 @@
...
+  get_filename_component(atomic_file ${output} NAME)
+  set(atomic_file "${CMAKE_BINARY_DIR}/${atomic_file}")
   configure_file("${input}" "${atomic_file}" ${ARGN})
   get_filename_component(output_path ${output} PATH)
-  if(NOT EXISTS ${output_path})
-    file(MAKE_DIRECTORY ${output_path})
-  endif()
-  file(RENAME "${atomic_file}" "${output}")
+  # sync multiple catkin cmake processes writing to that location
+  file(LOCK "${output_path}" DIRECTORY GUARD FUNCTION)
```
dirk-thomas: This signature is only available as of CMake 3.2. Please use it only conditionally.

While this ensures mutually exclusive access in CMake are any of the files generated by this function accessed in parallel from non-CMake processes?

rhaschke:
@dirk-thomas Valid point. Fixed.

cmake/atomic_configure_file.cmake
```diff
@@ -1,9 +1,11 @@
 function(atomic_configure_file input output)
-  set(atomic_file "${CMAKE_BINARY_DIR}/atomic_configure_file")
+  get_filename_component(atomic_file ${output} NAME)
+  set(atomic_file "${CMAKE_BINARY_DIR}/${atomic_file}")
   configure_file("${input}" "${atomic_file}" ${ARGN})
   get_filename_component(output_path ${output} PATH)
-  if(NOT EXISTS ${output_path})
-    file(MAKE_DIRECTORY ${output_path})
+  # sync multiple catkin cmake processes writing to that location
+  if (CMAKE_VERSION VERSION_GREATER 3.2)
```
dirk-thomas: The condition evaluates to `FALSE` for CMake 3.2 for which it should be `TRUE`.

cmake/atomic_configure_file.cmake
```diff
@@ -1,9 +1,11 @@
 function(atomic_configure_file input output)
-  set(atomic_file "${CMAKE_BINARY_DIR}/atomic_configure_file")
+  get_filename_component(atomic_file ${output} NAME)
+  set(atomic_file "${CMAKE_BINARY_DIR}/${atomic_file}")
```
dirk-thomas: To reduce the chance of collisions if the basename of `output` is the same as an existing file in the binary dir please keep `atomic_configure_file` as a prefix to the name: `set(atomic_file "${CMAKE_BINARY_DIR}/atomic_configure_file_${atomic_file}")`

cmake/atomic_configure_file.cmake
rhaschke: Fixed.

cmake/atomic_configure_file.cmake
rhaschke: This was changed intentionally. Using a different name than the final one, the copying needs two steps: first copying and then renaming the file at the destination location. (Remember that COPY doesn't allow for a simultaneous rename).
For me, this contradicts the atomic idea.

dirk-thomas: Good point - I didn't consider that. Can you place the file in a subdirectory `atomic_configure_file` then to minimize the chance of collisions.

dirk-thomas:
Thank you for the patch and for iterating on it.

PR #958, opened 2018-09-19
Catkin logo

dgossow:
Clearly, catkin needs a logo.

I made this roughly 6 years ago and just rediscovered the file in an old backup. Somehow I never got around to putting it anywhere, but now has come the time!

bildschirmfoto 2018-09-19 um 13 33 29

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,3 +1,5 @@
+.. image:: doc/catkin.svg
+   :width: 100px
 catkin
 ======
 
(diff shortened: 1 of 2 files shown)


wjwwood:
Yours looks more anatomically correct :), but there’s actually already a logo: 
https://github.com/catkin (icon used in organization)

I don’t care which we use, yours looks nice though :). Maybe we should also use it for ament.

dirk-thomas:
Thank you for creating this logo. If there is no strong reason to change it I would rather keep the exiting one (from the `catkin` org unit). I created #959 to use the Gravatar image in the README as in this patch.

dirk-thomas:
Closing in favor of #959.

dgossow:
Nice logo, I didn't know this existed!
Thanks for the review and follow up PR.

PR #962, opened 2018-10-05
Document how to install python subpackages

jdlangs:
As I recently discovered via ROS Answers [#273090](https://answers.ros.org/question/273090/setuppy-not-installing-python-module-during-catkin-build/), there is a non-obvious method to make catkin install python subpackages. Hopefully an explicit note in the docs will help one or two other people faster down the road.

Thanks,
Josh

### DIFF
--- a/doc/howto/format1/installing_python.rst
+++ b/doc/howto/format1/installing_python.rst
@@ -73,6 +73,14 @@ layout, it looks like this::
   
   setup(**setup_args)
 
+Note: `As in setuptools
+<https://packaging.python.org/guides/distributing-packages-using-setuptools/#packages>`_,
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Thank you for the improvement. While this is standard Python setuptools behavior it might help users to mention it explicitly.

Please update the PR to include the same change in the sibling directory `format1`.

jdlangs:
I have to assume there are plenty of other ROS users like me who have never used setuptools so the docs probably shouldn't assume any standard behavior would be known. I did find where this was mentioned in the python packaging docs so I added a link there as well as the `format1` version.

dirk-thomas:
Thank you for improving the documentation!

PR #966, opened 2018-10-10
fix symbolic link

eisoku9618:
travis fails because CMakeLists.txt cannot be found.
- https://travis-ci.org/ros/catkin/jobs/439639963

It's because the way to create a symbolic link of CMakeLists.txt is not correct.

### DIFF
--- a/test/utils.py
+++ b/test/utils.py
@@ -84,9 +84,8 @@ def create_catkin_workspace(pth):
     workspace_cmake = os.path.join(pth, "CMakeLists.txt")
     if os.path.isfile(workspace_cmake):
         os.remove(workspace_cmake)
-    succeed(["/bin/ln", "-s", "catkin/cmake/toplevel.cmake",
-             "CMakeLists.txt"],
-            cwd=pth)
+    succeed(["/bin/ln", "-s", pth + "/catkin/cmake/toplevel.cmake",
+             pth + "/CMakeLists.txt"])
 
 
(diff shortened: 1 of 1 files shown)


dirk-thomas:
> It's because the way to create a symbolic link of CMakeLists.txt is not correct.

Can you elaborate on this a bit more since it is unclear to me what was incorrect before and how the patch changes it.

Also the current state still seems to fail Travis CI.

eisoku9618:
> Also the current state still seems to fail Travis CI.

sorry, it's my misunderstanding...
(I thought that ``${from_path}`` and ``${to_path}`` have to be absolute path with ``ln -s ${from_path} ${to_path}``, but it's not.)

Currently, origin/indigo-devel fails Travis CI as follows.

| | origin/kinetic-devel | origin/indigo-devel |
| --- | --- | --- |
| status | pass | fail |
| evidence | https://github.com/ros/catkin/pull/967 | https://github.com/ros/catkin/pull/968 |

eisoku9618:
By the way, origin/indigo-devel can pass the tests in docker

```bash
docker run --rm -it ubuntu:14.04 /bin/bash
apt update && apt install -y git cmake python-pip
pip install argparse catkin-pkg empy mock nose
cd ~/
git clone -b indigo-devel https://github.com/ros/catkin.git
cd catkin
nosetests test/unit_tests test/local_tests
```

PR #967, opened 2018-10-13
check whether travis test pass with the latest branch

eisoku9618:


dirk-thomas:
I will close this since it seems to have served its purpose of showing that the `kinetic-devel` branch passes CI atm.

PR #968, opened 2018-10-13
check whether travis test pass with indigo-devel branch

eisoku9618:


dirk-thomas:
I couldn't find the problem on a quick look. I created #971 to track the problem.

PR #965, opened 2018-10-10
need to sanitize catkin_ws paths before cprint

eisoku9618:
When the catkin_ws includes "@" (at sign), catkin_make_isolated does not work since ``ColorTemplate`` of ``terminal_color.py`` uses "@" as delimiter.
So ``terminal_color.py`` has a function named ``sanitize`` to replace "@" to "@@", but the authors seem to miss to apply ``sanitize`` at some points at ``builder.py``.

PR #969, opened 2018-10-13
need to sanitize catkin_ws paths before cprint

eisoku9618:
This PR is the latest distro version of https://github.com/ros/catkin/pull/965 as suggested in https://github.com/ros/catkin/pull/965#issuecomment-428637367


dirk-thomas:
Thanks for the patch.

Issue #933, opened 2018-04-06
gtest gmock mix up


simon-rotter:
https://github.com/ros/catkin/blob/a9f6b62cddff3348a47e7f083a677d547e9bcdbc/cmake/test/gtest.cmake#L404

The file " catkin/cmake/test/gtest.cmake" seems to cause trouble on some machines in the latest catkin version (0.7.11) regarding the gmock / gtest pathes. Although the according lines were not changed during the last update; due to some specific configuration (additionally setting the according pathes at different places) these specific lines in this block were attracting attention and might be worth to be checked, It seems that some mix up might happen here:

For the following lines compare the sample code with the debug output were a simple test setup was created:
${GTEST_BOTH_LIBRARIES} and ${GMOCK_BOTH_LIBRARIES} seem to contain the same path, which leads to the output 

   Expected path to GTEST_BOTH_LIBRARIES:    ../libgmock.so/usr/lib/libgmock_main.so
   Expected path to GMOCK_BOTH_LIBRARIES:    ../libgmock.so/usr/lib/libgmock_main.so

In my case the issue becomes clear if having a closer look at the execution of the tests. 

[gtest_gmock_bug.tar.gz](https://github.com/ros/catkin/files/1884443/gtest_gmock_bug.tar.gz)

sloretz:
Hi @simon-rotter, thanks for creating the issue. I'm not sure what you mean by "cause trouble". Are you saying some packages are failing to build tests on some systems? How do I need to set my system up to reproduce the issue you're seeing?

It looks like the line of code referenced happens if there is a find module for GMock. The googlemock documentation says "[Google Mock is implemented on top of the Google Test C++ testing framework, and depends on it. You must use the bundled version of Google Test when using Google Mock ...](https://github.com/google/googlemock/blob/master/googlemock/README.md#requirements-for-end-users)". It sounds reasonable to me that `GTEST_...` variables would get set to gmock locations.

Related PR #897

dirk-thomas:
Closing due to no response.

PR #973, opened 2018-11-18
Fixed gtest.cmake when both gtest and gmock are built from source

### LINKED ISSUE #970: GMock is not able to be found/build when GTest is properly installed
CMake by default comes with FindGTest.cmake but is does not come with FindGMock.cmake. 

In gtest.cmake:289 and onwards, if gmock is not found and gtest is found(which is the case because we have FindGTest.cmake and not FindGMock.cmake), it would goto line 365 and only try to setup and build gtest and completely ignores gmock.

This is not behavior I expected, I think it should also try to search for gmock with catkin_find_google_test_source before it completely gives up.

tech4me:
This pull request address #933. Under a system where both gtest and gmock are installed, the cmake variables for GTEST_INCLUDE_DIRS GTEST_LIBRARY_DIRS GTEST_LIBRARIES GTEST_MAIN_LIBRARIES GTEST_BOTH_LIBRARIES were set incorrectly.
i.e. when trying to link to both gtest and gmock, only gmock will be linked(and linked twice, with linker flag -gmock -gmock)

This doesn't resolve #970, that will require a more significant refactor to gtest.cmake


dirk-thomas:
Please describe a reproducible case (platform, installed packages, workspace, steps) which fails without this patch and passes with this patch.

dirk-thomas:
Closing due to no response.

PR #975, opened 2018-12-02
strip "-l" from "-lpthread" library, to fix a build failure on ros-ros-comm

jspricke:


dirk-thomas:
This looks like a duplicate of #856.

jspricke:
Some more reasoning for this here: https://bugs.debian.org/915148

mathias-luedtke:
@dirk-thomas: TL;DR: Upstream CMake just started to inject `-lpthread` into `Boost_LIBRARIES` .

dirk-thomas:
With the changes in CMake being rolled out I guess we don't have another option than to accept this case and pass it along.

Thanks for the patch.

wajeehulhassanvii:
> _No description provided._

**Felix Geyer
@jspricke
strip "-l" from "-lpthread" library, to fix a build failure on ros-ro…** 

Felix, can you please let me know where can I apply this patch in simple steps to get it working. Currently I am stuck at -lpthread error and since I'm new to ROS therefore there's nothing much I can do.

Thanks

xlla:
> @dirk-thomas: TL;DR: Upstream CMake just started to inject `-lpthread` into `Boost_LIBRARIES` .

when built ros melodic in raspberry pi, stretch, I am encounter same error, even no '-l' before 'pthread'.
after change to boost-libraries, it work.

PR #976, opened 2019-01-08
add double quotes around cmake_command to handle spaces

kejxu:
there might be spaces within `${CMAKE_COMMAND}` so need to wrap it with double quotation marks

### DIFF
--- a/cmake/test/nosetests.cmake
+++ b/cmake/test/nosetests.cmake
@@ -73,7 +73,15 @@ function(catkin_add_nosetests path)
   set(output_path ${CATKIN_TEST_RESULTS_DIR}/${PROJECT_NAME})
   # make --xunit-file argument an absolute path (https://github.com/nose-devs/nose/issues/779)
   get_filename_component(output_path "${output_path}" ABSOLUTE)
-  set(cmd "${CMAKE_COMMAND} -E make_directory ${output_path}")
+
+  if(WIN32)
+    # add quotes to handle spaces properly
+    set(cmd \"${CMAKE_COMMAND}\")
+  else()
+    set(cmd ${CMAKE_COMMAND})
+  endif()
+
+  set(cmd "${cmd} -E make_directory ${output_path}")
   if(IS_DIRECTORY ${_path_name})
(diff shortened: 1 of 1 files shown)


cmake/test/nosetests.cmake
```diff
@@ -73,7 +73,15 @@ function(catkin_add_nosetests path)
   set(output_path ${CATKIN_TEST_RESULTS_DIR}/${PROJECT_NAME})
   # make --xunit-file argument an absolute path (https://github.com/nose-devs/nose/issues/779)
   get_filename_component(output_path "${output_path}" ABSOLUTE)
-  set(cmd "${CMAKE_COMMAND} -E make_directory ${output_path}")
+
+  if(WIN32)
```
dirk-thomas: Why should this be Windows specific? Couldn't the same happen on other platforms?

kejxu: Thanks for the question! Yes, supposedly this should be a problem across platforms; we ran into this problem when CMake was installed into a path with spaces. Since the entire command would be passed into https://github.com/ros/catkin/blob/kinetic-devel/cmake/test/tests.cmake#L131 as a string (`${_testing_COMMAND}`), this was our workaround to make sure spaces are not a problem.

I'm not sure why this was never hit on a ubuntu distribution; maybe it's because packages are usually installed into paths without a space by default (/etc, ...). On Windows, "Program Files" is often the default location, so it's a bit easier to hit.

seanyen: What if we do a smoke test to remove the WIN32 and see if `set(cmd \"${CMAKE_COMMAND}\")`  also works as expected on Ubuntu, and make the changes platform agnostic?

And it's nice to add comments on the examples why it is a problem for some platform. For example, CMAKE_COMMAND could be pointed to somewhere under "C:\Program Files" (on Windows) and spaces not quoted correctly results in problems for Windows shell mistaking what executable path to run.

dirk-thomas: I applied the quoting generically.

dirk-thomas:
Thank you for the patch.

PR #977, opened 2019-01-09
use ${PYTHON_EXECUTABLE} in catkin_download

kejxu:
use `${PYTHON_EXECUTABLE}` instead of `${PYTHON}` in the `catkin_download` function.

In execution, `${PYTHON}` is interpreted as an empty string, and the command would not be executed on Windows (a python script is not executable without extra setup). this might not be noticeable on Ubuntu since python scripts can be executed without the `python` prefix


dirk-thomas:
Thanks for the fix.

PR #980, opened 2019-01-09
check both IMPORTED_IMPLIB_ and IMPORTED_LOCATION_ in catkin_libraries.cmake

kejxu:
Check both locations (one for dll imported and the other for static lib) for Windows.

Otherwise, for the package exporting static lib, the lib path won't be published correctly to downstream. For example, `orocos_kdl` won't be exported from `tf2_kdl`.


dirk-thomas:
Thanks for the patch.

PR #981, opened 2019-01-09
add script.bat.in template for general use

kejxu:
similar to the use of the bash script template, `script.bat.in` will be used as a batch script wrapper for `rosrun.bat` (https://github.com/ms-iot/ros/blob/init_windows/tools/rosbash/CMakeLists.txt#L27)

### DIFF
--- a/cmake/templates/script.bat.in
+++ b/cmake/templates/script.bat.in
@@ -1,4 +1,4 @@
 @echo off
 REM generated from catkin/cmake/templates/script.bat.in
 
-"@BAT_SCRIPT@" %*
+call "@BAT_SCRIPT@" %*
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Is this relay script used anywhere else beside `rosbash`? If not I would rather prefer the logic to be moved to the `rosbash` package or e.g. simply copy the file rather than using this "emulated symlink style".

kejxu:
> Is this relay script used anywhere else beside `rosbash`? If not I would rather prefer the logic to be moved to the `rosbash` package or e.g. simply copy the file rather than using this "emulated symlink style".

Totally agree! We are comfortable with this proposed change because there are existing wrappers in cmake\templates already, this added file will be a Windows counterpart of the existing [script.bash.in](https://github.com/ros/catkin/blob/kinetic-devel/cmake/templates/script.bash.in). If later one day we/ros decide to move these into `rosbash`, they can be moved all at the same time =)

dirk-thomas:
Thank you for the patch.

PR #982, opened 2019-01-10
update builder.py to add Windows support

kejxu:
update `builder.py` to support building on Windows:
- exclude drive letter so that `path[1:]` gives the correct result on Windows
- abstract the action of writing env and setup scripts into script writers
- define script templates in a separate step (for better readability)
- add templates for Windows batch scripts
- expand variables for [env.bat template](https://github.com/ros/catkin/blob/kinetic-devel/cmake/templates/env.bat.in) and [setup.bat template](https://github.com/ros/catkin/blob/kinetic-devel/cmake/templates/setup.bat.in)

this code change does not alter the existing logic on Linux

### DIFF
--- a/python/catkin/builder.py
+++ b/python/catkin/builder.py
@@ -1074,8 +1074,9 @@ def build_workspace_isolated(
     if not os.path.exists(develspace):
         os.makedirs(develspace)
     if not build_packages:
+        env_script = 'env'
         if sys.platform == 'win32':
-            env_script = 'env.bat'
+            env_script += '.bat'
             env_script_content = """\
 @echo off
 REM generated from catkin.builder module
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thanks for the patch.

python/catkin/builder.py
```diff
@@ -554,13 +655,14 @@ def build_cmake_package(
         make_install_cmd = [last_env] + make_install_cmd
     run_command(make_install_cmd, build_dir, quiet)
 
-    # If we are installing, and a env.sh exists, don't overwrite it
-    if install and os.path.exists(prefix_destdir(os.path.join(install_target, 'env.sh'), destdir)):
+    env_script = 'env' + '.bat' if sys.platform == 'win32' else '.sh'
```
dirk-thomas, line 658: This change broke non-Windows platforms: see #1000.

kejxu: my apologies for the regression!

PR #983, opened 2019-01-10
ensure desired Python path is added into PATH in setup.bat

kejxu:
thanks to the community's interest in the project and trying out ROS on Windows changes (thanks!), one (minor) issue that has been reported is that when there is existing Python on the build machine, it could be tricky to find and use the desired python version/path.

To address this, update `setup.bat` template to check for the desired python path and add that into the `PATH` variable


dirk-thomas:
Thanks for the patch.

PR #984, opened 2019-01-10
update windows.cmake to fix common build issues on Windows

kejxu:
while building on Windows, it's not uncommon to encounter various platform-specific issues and compiler-specific issues (if using visual c++ compiler, MSVC).

For the debug flags and compiler options, a lot of these issues are already known to the community, and most of them could be resolved with existing good practices:
- `-DNOMINMAX` helps avoid collision with `std::min`/`std::max` if there's `windows.h`/`windef.h`
- `-DNO_STRICT` (and `-DQ_NOWINSTRICT` for QT) disables STRICT type checking
- `-D_USE_MATH_DEFINES` allows using Math Constants since they are not defined in Standard C/C++
- `-DWIN32_LEAN_AND_MEAN` excludes various APIs from windows.h to improve build time since basic inclusion would be enough

### DIFF
--- a/cmake/platform/windows.cmake
+++ b/cmake/platform/windows.cmake
@@ -60,4 +60,9 @@ if(WIN32)
   add_definitions(-DNOMINMAX)   # not to define min/max macros
   add_definitions(-DNO_STRICT)  # not to define STRICT macros (minwindef.h or boost\winapi\basic_types.hpp)
   add_definitions(-DQ_NOWINSTRICT)  # not to define STRICT macros (qtgui\qwindowdefs_win.h)
+  add_definitions(-D_USE_MATH_DEFINES)  # enable Math Constants https://docs.microsoft.com/en-us/cpp/c-runtime-library/math-constants
+endif()
+
+if(MSVC)
+  add_compile_options(/Zc:__cplusplus) # https://blogs.msdn.microsoft.com/vcblog/2018/04/09/msvc-now-correctly-reports-__cplusplus/
 endif()
--- a/cmake/platform/windows.cmake
+++ b/cmake/platform/windows.cmake
@@ -41,9 +41,10 @@ if(BUILD_SHARED_LIBS)
   if(WIN32)
     function(add_library library)
       # Check if its an external, imported library (e.g. boost libs via cmake module definition)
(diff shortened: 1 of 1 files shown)


cmake/platform/windows.cmake
```diff
@@ -41,9 +41,10 @@ if(BUILD_SHARED_LIBS)
   if(WIN32)
     function(add_library library)
       # Check if its an external, imported library (e.g. boost libs via cmake module definition)
-      list(FIND ARGN "IMPORTED" FIND_POS)
+      list(FIND ARGN "IMPORTED" FIND_IMPORTED)
+      list(FIND ARGN "ALIAS" FIND_ALIAS)
```
kejxu, line 45: Avoid using set_target_properties on `ALIAS` targets

This change is to fix the error like below:
```
2018-11-30T10:40:39.7610068Z CMake Error at C:/opt/ros/melodic/x64/share/catkin/cmake/platform/windows.cmake:47 (set_target_properties):
2018-11-30T10:40:39.7610963Z   set_target_properties can not be used on an ALIAS target.
2018-11-30T10:40:39.7611263Z Call Stack (most recent call first):
2018-11-30T10:40:39.7611903Z   abseil_cpp/CMake/AbseilHelpers.cmake:67 (add_library)
2018-11-30T10:40:39.7612185Z   abseil_cpp/absl/base/CMakeLists.txt:74 (absl_library)
```
https://github.com/ms-iot/catkin/commit/8545d9f28f0d5bd93ab0f6fa2348e1585a74bea7#diff-2249c319ebe44cfef5e351de7305d244

dirk-thomas:
While I am not a big fan of working around the various definitions globally within `catkin` I have to acknowledge that for a large existing code base like ROS 1 it is probably the more realistic approach.

Thank you for the patch.

kejxu:
> While I am not a big fan of working around the various definitions globally within `catkin` I have to acknowledge that for a large existing code base like ROS 1 it is probably the more realistic approach.

Thanks a lot for your understanding! Hopefully this kind of change could be eventually removed from the global build tool (`catkin`) when all the packages could provide native support Windows (some day); before that, however, this might be an easier way to enable everything working on Windows =)

PR #986, opened 2019-01-28
fix PYTHONPATH missing in cached environment

### LINKED ISSUE #974: Importing python module fails, need to run catkin_make twice
Somehow related to https://github.com/ros/catkin/issues/841 yet I am not sure if entirely the same.

I created a minimal example here: https://github.com/agutenkunst/bug_python_test_deps

If I start with a clean workspace and do
`catkin_make`
`catkin_make run_tests_bug_python_test_deps_rostest_test_foo_test.test`
it will fail with **ImportError: No module named bug_python_test_deps.foo**

If I now do a `touch build/CMakeCache.txt`
and run the test again with
`catkin_make run_tests_bug_python_test_deps_rostest_test_foo_test.test`
everything works fine.

Printing the pythonpath inside my test yields that at the second run a additional
**/home/[username]/catkin_ws/devel/lib/python2.7/dist-packages** 
appears in the path, which is exactly the missing location in the first run.

Is there something that I didn't get correct?

dirk-thomas:
Fixes #974.


sloretz (approved):
I tested that this fixes the original issue on xenial (modified the example slightly to use `catkin_add_nosetests()` instead of `rostest` and `rclpy`)

PR #989, opened 2019-02-01
[doc] Recommend PACKAGE, not GLOBAL, in user's doc.

130s:
Installing executable into `CATKIN_GLOBAL_BIN_DESTINATION` makes it globally accesible. While it's certainly nice, I'm not sure if that's what Catkin's documentation that generally recommends installing into each package space should do. It makes sense for e.g. ROS core packages, but not so much for users' package IMO.

Although I found an old related thread https://github.com/ros/catkin/issues/410#issuecomment-17312357, I'm not sure if that was where the decision to use global space was made.


dirk-thomas:
Please read the lines right below the modified snippets:

> The runtime destination is used for `.dll` file on Windows which must be placed in the global bin folder.

dirk-thomas:
Just to clarify: this part of the documentation is for library targets - not executables.

PR #990, opened 2019-02-06
Switch docs to suggest 'test' path instead of 'tests'.

mikepurvis:
Just noticed that basically all of ROS upstream uses a `test` dir, but this doc suggests `tests`, which has led to an inconsistency in our internal packages, with some following the documented example and others following upstream.

Note that this change is also consistent with the generated `CMakeLists.txt` template:

https://github.com/ros-infrastructure/catkin_pkg/blob/e4b9bcd884b5f7fb9e6ddcfaa976aa3c72c22654/src/catkin_pkg/templates/CMakeLists.txt.in#L192-L193


dirk-thomas:
Thanks for the patch. Singular makes indeed more sense.

PR #992, opened 2019-02-20
[Windows][kinetic] Fix python_distutils_install.bat.in

seanyen:
* Fix: `PYTHONPATH` isn't really passed to `python setup.py` by `cmd /c /v` method.
* Fix: For clean build, `@CMAKE_INSTALL_PREFIX@\@PYTHON_INSTALL_DIR@` could not exist, so use `mkdir` to ensure the existence. (`python_distutils_install.sh.in` did the same thing.)
* Fix: move quotation from `set ABC="xyz"` to `set "ABC=xyz"` to avoid extra double quotes in the variable, otherwise, which subsequently caused problems for Python to parse the `PYTHONPATH`.

Here is the issue how does the problem manifest: https://github.com/ms-iot/ROSOnWindows/issues/67


dirk-thomas:
Thanks for the patch.

seanyen:
Thanks!

PR #998, opened 2019-03-12
Protection against double -l

lepalom:
I have been working rebuilding ROS Melodic for Stretch using the debian packages
and I have been blocked in an error in qt_gui_cpp:

/usr/bin/ld: cannot find -l-lpthread

that error have been found in several places:

https://aur.archlinux.org/packages/ros-melodic-qt-gui-cpp/

https://github.com/ros/ros-overlay/issues/711

and probably more places.

You have been working in a patch for something similar:

https://github.com/ros/catkin/pull/975

But I think that the problems comes from another place and  @ipa-mdl point me to
 the error when here said:

ipa-mdl commented on 2 Jan

@dirk-thomas: TL;DR: Upstream CMake just started to inject -lpthread into
Boost_LIBRARIES .

The problem comes from the the .pc files. It's strange to me because from all
the pc files generated by catkin that we have in debian (Buster/Sid), I have
only detected rosconsole.pc that had -l-lpthread.

But, in the backported version that I have, there are more. I'm using
cmake-3.13.2-1 and maybe, it comes because that version. The vast majority of
debian packages were generated before cmake 3.12 or 3.13 arrived to the archive.
In Ubuntu bionic cmake is 3.10.2.

The solution is simple, is this patch.

I have to admit that in gazebo, Jose Rivero incorporated a patch from Bas
Couwenberg solving a very similar issue.

What do you think?


dirk-thomas:
Thanks for the patch. I verified that the problem exists in e.g. Ubuntu Disco (which has a new enough CMake version to export `-lpthread`) and this patch fixes the generated `.pc` files.

lepalom:
Not solving yet all the issues :-(
https://github.com/ros-visualization/qt_gui_core/issues/174

dirk-thomas:
> Not solving yet all the issues :-(
> ros-visualization/qt_gui_core#174

This report wasn't used the latest state of all patches.

lepalom:
well, I was using the 0.3.4 version  that seem that is the version in melodic:
https://github.com/ros-gbp/python_qt_binding-release

what about to release a 0.3.5 with all that patches?

dirk-thomas:
> what about to release a 0.3.5 with all that patches?

Sounds good: ros/rosdistro#20577

PR #1000, opened 2019-03-18
Fix setup.sh/env.sh filename

at-wat:
Filenames `setup.sh` and `env.sh` get `.sh` after https://github.com/ros/catkin/pull/982.
This seems to cause test failures of pure cmake packages on the buildfarm.

Following builds on the buildfarm looks affected by this bug.

http://build.ros.org/view/Mdev/job/Mdev__yp-spur__ubuntu_bionic_amd64/18/console
```
02:29:23 Invoking '. /opt/ros/melodic/setup.sh && . /tmp/ws/devel_isolated/setup.sh && PYTHONIOENCODING=utf_8 PYTHONUNBUFFERED=1 catkin_make_isolated --force-cmake --cmake-args -DBUILD_TESTING=1 -DCATKIN_ENABLE_TESTING=1 -DCATKIN_SKIP_TESTING=0 -DCATKIN_TEST_RESULTS_DIR=/tmp/ws/test_results --catkin-make-args run_tests' in '/tmp/ws'
02:29:23 /bin/sh: 9: .: Can't open /tmp/ws/devel_isolated/ypspur/setup.sh
```
http://build.ros.org/view/Mdev/job/Mdev__rc_genicam_api__ubuntu_bionic_amd64/17/console
```
09:42:52 Invoking '. /opt/ros/melodic/setup.sh && . /tmp/ws/devel_isolated/setup.sh && PYTHONIOENCODING=utf_8 PYTHONUNBUFFERED=1 catkin_make_isolated --force-cmake --cmake-args -DBUILD_TESTING=1 -DCATKIN_ENABLE_TESTING=1 -DCATKIN_SKIP_TESTING=0 -DCATKIN_TEST_RESULTS_DIR=/tmp/ws/test_results --catkin-make-args run_tests' in '/tmp/ws'
09:42:52 /bin/sh: 4: .: Can't open /tmp/ws/devel_isolated/rc_genicam_api/setup.sh
```

---
Expression evaluation order of python:
```python
>>> 'a' + 'b' if True else 'c'
'ab'
>>> 'a' + 'b' if False else 'c'
'c'
```

### DIFF
--- a/python/catkin/builder.py
+++ b/python/catkin/builder.py
@@ -655,7 +655,7 @@ def build_cmake_package(
         make_install_cmd = [last_env] + make_install_cmd
     run_command(make_install_cmd, build_dir, quiet)
 
-    env_script = 'env' + '.bat' if sys.platform == 'win32' else '.sh'
+    env_script = 'env' + ('.bat' if sys.platform == 'win32' else '.sh')
     # If an env script already exists, don't overwrite it
     if install and os.path.exists(prefix_destdir(os.path.join(install_target, env_script), destdir)):
         return
@@ -677,7 +677,7 @@ def build_cmake_package(
 
     # Generate setup script for chaining to catkin packages
     # except if using --merge which implies that new_setup_path equals last_setup_env
-    setup_script = 'setup' + '.bat' if sys.platform == 'win32' else '.sh'
+    setup_script = 'setup' + ('.bat' if sys.platform == 'win32' else '.sh')
     new_setup_path = os.path.join(install_target, setup_script)
     if install:
         new_setup_path = prefix_destdir(new_setup_path, destdir)


kejxu:
Thanks for the fix, my apologies for the regression!

dirk-thomas:
Thank you for fixing the regression. I will make a new patch release right away.

AAlon:
@dirk-thomas When can we expect 0.7.17 to be available via apt?
Thanks!

dirk-thomas:
> When can we expect 0.7.17 to be available via apt?

Yes, atm it is only in the building repos. As soon as the rebuild of downstream packages has finished it will become available in the testing repo (~ within a day). After that it will arrive in main as soon as a new Melodic sync is manually triggered which is also announced on Discourse (I would guess in a couple of weeks): http://repositories.ros.org/status_page/ros_melodic_default.html?q=catkin

AAlon:
> > When can we expect 0.7.17 to be available via apt?
> 
> Yes, atm it is only in the building repos. As soon as the rebuild of downstream packages has finished it will become available in the testing repo (~ within a day). After that it will arrive in main as soon as a new Melodic sync is manually triggered which is also announced on Discourse (I would guess in a couple of weeks): http://repositories.ros.org/status_page/ros_melodic_default.html?q=catkin

Would it be possible to perform a one-off sync in order to get this fix out?  This is blocking prerelease testing (& releasing) Melodic packages. Not an issue on Kinetic since ros-kinetic-catkin is still @ 0.7.14. If it's too much trouble we'll have to look into injecting an updated catkin into the docker during the build, but that would only unblock prerelease testing - actual releasing would still be blocked.

dirk-thomas:
> This is blocking prerelease testing for Melodic packages (Not an issue on Kinetic since ros-kinetic-catkin is still @ 0.7.14...).

As mentioned 0.7.17 will be in testing within the next 24h - there is no way to speed this process up since downstream packages have to be rebuild for that. The Melodic version in main is still 0.7.14. None of the in between versions have been merged into main.

Based on your comment I don't see what the problem is. Please provide more context / details.

AAlon:
Oh, my confusion, missed the fact that the 0.7.16 version was pulled from testing. Sounds good then, thanks.

dirk-thomas:
> missed the fact that the 0.7.16 version was pulled from testing

It wasn't "pulled" from testing - it got replaced with the newly build 0.7.17 (the ROS apt repos only store the latest version of a package).

dirk-thomas:
With the new `catkin` version in the testing repo the latest devel builds passed again:

* http://build.ros.org/view/Mdev/job/Mdev__yp-spur__ubuntu_bionic_amd64/19/
* http://build.ros.org/view/Mdev/job/Mdev__rc_genicam_api__ubuntu_bionic_amd64/18/

kejxu:
> With the new catkin version in the testing repo the latest devel builds passed again:
> 
> http://build.ros.org/view/Mdev/job/Mdev__yp-spur__ubuntu_bionic_amd64/19/
> http://build.ros.org/view/Mdev/job/Mdev__rc_genicam_api__ubuntu_bionic_amd64/18/

Thanks for the fix @at-wat, and thanks so much for organizing and getting everything working again @dirk-thomas!

PR #1001, opened 2019-03-20
doc: dyn_recfg: add_deps(..) is only needed with C++

gavanderhoorn:
As per subject.

Triggered by [CMake Error: File /cmake/setup_custom_pythonpath.sh.in does not exist.](https://answers.ros.org/question/318918) on ROS Answers.


dirk-thomas:
Can you please apply the same change for format 1.

gavanderhoorn:
Sure. See 8e32c904.

dirk-thomas:
Thanks for the patch.

PR #1003, opened 2019-03-28
include threads library

mgrrx:
Due to https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=868234 ligtest.a is now shipped with libgtest-dev. Consequently gtest targets in catkin link against this library on newer Debian / Ubuntu versions.
For some reason, the tools/threads.cmake file was commented out in all.cmake but the variable THREADS_LIBRARY is actually used when linking against the libgtest.a (https://github.com/ros/catkin/blob/kinetic-devel/cmake/test/gtest.cmake#L177). Removing the comment fixes the linking issues for me.

### DIFF
--- a/cmake/all.cmake
+++ b/cmake/all.cmake
@@ -158,7 +158,7 @@ foreach(filename
     tools/libraries
     tools/rt
 
-#    tools/threads
+    tools/threads
   )
   include(${catkin_EXTRAS_DIR}/${filename}.cmake)
 endforeach()


sloretz:
@mgrrx It looks like the debian bug issue was resolved as `Fixed in version googletest/1.8.0-7` in April. Is this PR still needed on Debian Buster? If so, how do I reproduce the linking issues?

jspricke:
>@mgrrx It looks like the debian bug issue was resolved as `Fixed in
>version googletest/1.8.0-7` in April. Is this PR still needed on Debian
>Buster? If so, how do I reproduce the linking issues?

Fixed as in libgtest.a is in the package and this patch is needed since then, so you could test it with buster.

dirk-thomas:
Why should this be done unconditionally? Wouldn't it be cleaner to only do that when `gtest` is being used?

dirk-thomas:
One other Debian Buster related PR has been merged (#1022) and another one is going to be merged soon (#1021). With this we were able to build ROS packages on Buster.

Can you please double check if using those you still experience the problem? If it still doesn't work for you on Buster please comment with additional information how to reproduce.

dirk-thomas:
Closing this for now due to no response.

PR #1009, opened 2019-05-23
Clarify consequences of running setup.py manually

nickolai-voyage:
Suggesting that the two installs will "influence each other" is incorrect. In most cases the version installed by setup.py will be ignored if the setup.sh script is sourced, but I believe it is better to inform users more accurately as opposed to imposing a blanket ban. This language should also help users who have run setup.py manually understand what their options are in terms of rectifying their situation.


dirk-thomas:
Thanks for the improvement.

PR #1010, opened 2019-05-31
[Windows][kinetic] Fix devel and install space intermixes in %PATH% after devel\setup.bat is sourced

seanyen:
This change is to address an issue originally reported here: https://github.com/ms-iot/ROSOnWindows/issues/69

On Windows, we observed that `%PATH%` intermixes `devel` and `install` space for `bin` and `lib`. Ideally the `devel\bin` and `devel\lib` should always come before the `install` space after `devel\setup.bat` is sourced. Otherwise, for example, at runtime, it could favor the content from `install space` over the one in `devel space` and subsequently cause unexpected problems, especially when we build\test a package both existed in `devel space` and `install space`.

This change is to propose to make sure `bin` and `lib` on Windows for `ENV_VAR_SUBFOLDERS['PATH']` in `_setup_util.py`, and then let `_setup_util.py` to prepend the workspace prefix accordingly. In this case, `_setup_util.py` can produce `%PATH%` like:
`C:\catkin_ws\devel\bin;C:\catkin_ws\devel\lib;C:\opt\ros\melodic\x64\bin;C:\opt\ros\melodic\x64\lib;...`

### DIFF
--- a/cmake/templates/_setup_util.py.in
+++ b/cmake/templates/_setup_util.py.in
@@ -51,8 +51,8 @@ IS_WINDOWS = (system == 'Windows')
 
 PATH_TO_ADD_SUFFIX = ['@CATKIN_GLOBAL_BIN_DESTINATION@']
 if IS_WINDOWS:
-    # 3rdparty packages often put dll's into lib (convention is bin) and
-    # windows finds it's dll's via the PATH variable. Make that happen here!
+    # while catkin recommends putting dll's into bin, 3rd party packages often put dll's into lib
+    # since Windows finds dll's via the PATH variable, prepend it with path to lib
     PATH_TO_ADD_SUFFIX.extend([@CATKIN_LIB_ENVIRONMENT_PATHS@])
 
 # subfolder of workspace prepended to CMAKE_PREFIX_PATH
--- a/cmake/templates/_setup_util.py.in
+++ b/cmake/templates/_setup_util.py.in
@@ -49,11 +49,17 @@ system = platform.system()
 IS_DARWIN = (system == 'Darwin')
 IS_WINDOWS = (system == 'Windows')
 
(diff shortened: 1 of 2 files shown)


dirk-thomas:
I am a bit confused how this patch relates to the described "`devel` / `install` mixup" described. Can you please elaborate what you are referring to with that.

seanyen:
@dirk-thomas 

Sorry for any confusions if I messed up on `catkin` terminology.

What I am talking about is that, before this change, say you have a ROS installation under `c:\opt\ros\melodic\x64\`, then you create a workspace, checkout some code, and do a `catkin_make`. Then you want to do some testing, so `devel\setup.bat` is sourced. Now you check `%PATH%`, and it will be something like `C:\catkin_ws\devel\lib;C:\opt\ros\melodic\x64\lib;C:\opt\rosdeps\x64\lib;C:\catkin_ws\devel\bin;C:\opt\ros\melodic\x64\bin;C:\opt\rosdeps\x64\bin;...`. And as you can notice, the `lib` paths with overlapping prefixes go first before all `bin` paths.

In this case, say if there is a package I am working on, which already existed on `C:\opt\ros\melodic\x64` prefix and placed a DLL under `lib`, and then I try to correct it to place it to `bin` and test my change in devel workspace. At runtime, due to the DLL search order honors the path order in `%PATH%`, the DLL under `C:\opt\ros\melodic\x64\lib` will be found first, instead of the new location I placed for testing, which is `C:\catkin_ws\devel\bin`. That's one example where we hit problems due to the search order.

That's why I figured the correct order should be more like `C:\catkin_ws\devel\bin;C:\catkin_ws\devel\lib;C:\opt\ros\melodic\x64\bin;C:\opt\ros\melodic\x64\lib;...`, where `bin` and `lib` should always go together in order of overlapping prefix.

dirk-thomas:
Thank you for the contribution as well as the clarification. Now I understand how the patch works.

PR #1011, opened 2019-06-01
[Windows][kinetic] Add catkin default parameter for Windows developers

seanyen:
This change is to make `catkin` more convenient for Windows developers to kick off a build.

* Currently the `catkin` is using `make` as the Makefile tool by default but it doesn't exist in MSVC toolchain on Windows, so Windows developers most likely need to add `--use-nmake` every time to successfully kick off  a build.

* Favor `RelWithDebInfo` over  the default `Debug`, which makes the Windows developers by default to build out portable `Release` binaries + its debug symbol files. CMake `Debug` configuration will link against the debug MSVC runtime (e.g., `vcruntime140d.dll`) which is not redistributable separately from Visual Studio.

### DIFF
--- a/bin/catkin_make
+++ b/bin/catkin_make
@@ -19,6 +19,7 @@ from catkin.builder import handle_make_arguments
 from catkin.builder import print_command_banner
 from catkin.builder import run_command
 from catkin.builder import run_command_colorized
+from catkin.builder import build_platform_specific_defaults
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Thank you for the patch. I made some small modifications.

seanyen:
Thanks for the merge!

PR #1012, opened 2019-06-13
Add additional variables to CMake build to help packages detect ROS

johnwason:
Currently there is no easy way for CMake packages to detect a ROS build. This pull request adds additional CMake variable args CATKIN_VERSION, CATKIN_ROS_VERSION, and CATKIN_ROS_DISTRO to the cmake command. This will allow packages to detect a ROS build and the ROS version so it can adjust accordingly.

@Levi-Armstrong

### DIFF
--- a/python/catkin/builder.py
+++ b/python/catkin/builder.py
@@ -48,6 +48,7 @@ except ImportError:
     from io import StringIO
 import subprocess
 import sys
+from os import environ
 
 try:
     from catkin_pkg.cmake import configure_file, get_metapackage_cmake_template_path
@@ -610,6 +611,11 @@ def build_cmake_package(
             '-DCMAKE_INSTALL_PREFIX=' + install_target
         ]
         cmake_cmd.extend(cmake_args)
+        cmake_cmd.append('-DCATKIN_VERSION:STRING=0.7.18')
+        if 'ROS_VERSION' in environ:
+            cmake_cmd.append('-DCATKIN_ROS_VERSION:STRING=' + environ['ROS_VERSION'])
+        if 'ROS_DISTRO' in environ:
+            cmake_cmd.append('-DCATKIN_ROS_DISTRO:STRING=' + environ['ROS_DISTRO'])
         isolation_print_command(' '.join(cmake_cmd), build_dir)
         if last_env is not None:
             cmake_cmd = [last_env] + cmake_cmd


python/catkin/builder.py
```diff
@@ -610,6 +611,11 @@ def build_cmake_package(
             '-DCMAKE_INSTALL_PREFIX=' + install_target
         ]
         cmake_cmd.extend(cmake_args)
+        cmake_cmd.append('-DCATKIN_VERSION:STRING=0.7.18')
+        if 'ROS_VERSION' in environ:
+            cmake_cmd.append('-DCATKIN_ROS_VERSION:STRING=' + environ['ROS_VERSION'])
+        if 'ROS_DISTRO' in environ:
+            cmake_cmd.append('-DCATKIN_ROS_DISTRO:STRING=' + environ['ROS_DISTRO'])
```
sloretz, line 618: It looks like `ROS_DISTRO` and `ROS_VERSION` can be gotten by a cmake package using `$ENV{ROS_DISTRO}` or `$ENV{ROS_VERSION}` https://cmake.org/cmake/help/v3.5/variable/ENV.html

johnwason: This is possible, however these extra CMake variables will help deal with potentially ambiguous versioning. For packages attempting to target different ROS versions and pure cmake builds it is better to be explicit.

johnwason:
Related: https://github.com/colcon/colcon-ros/pull/64

python/catkin/builder.py
sloretz: What causes the version to be ambiguous? Are you saying there is a case with this PR where `CATKIN_ROS_DISTRO` would be different from `$ENV{ROS_DISTRO}`?

johnwason: The build can either be done using Colcon, Catkin, or plain CMake, and they all have $ENV{ROS_DISTRO} set. There needs to be a way to detect that this is a ROS build and what tool is being used.

sloretz: Is this need based on an assumption that `colcon` means ROS 2, `catkin` means ROS 1, and `cmake` means no ROS at all? If so, I don't think that's a safe assumption to make.

johnwason: No, colcon can be used to build ROS 1 packages as well...

sloretz: > No, colcon can be used to build ROS 1 packages as well...

I'm trying to figure out what the need is. What does the cmake package want to do differently for each build tool?

> There needs to be a way to detect that this is a ROS build and what tool is being used.

It seems like `CATKIN_VERSION` and `COLCON_VERSION` are enough to say which tool called `cmake`, and `$ENV{ROS_DISTRO}` or `$ENV{ROS_VERSION}` are enough to say which version of ROS is being used. What are `CATKIN_ROS_DISTRO` and `COLCON_ROS_DISTRO` for?

johnwason: It is because Colcon and Catkin can be used outside of ROS! And yes, Colcon can be used to build ROS 1 packages. For a library maintainer there needs to be a way to EXPLICITLY detect what command has been executed. Right now the ROS-I team is doing acrobatics to attempt to detect and configure the different possible build conditions. It should not be this complicated.

clalancette:
For what it is worth, I don't think this will be sufficient for ROS 1.  Large swaths of the community have switched over to using https://github.com/catkin/catkin_tools , so we'd also have to add similar things there for packages to be able to build using either (which is the usual expectation).

python/catkin/builder.py
dirk-thomas: > For a library maintainer there needs to be a way to EXPLICITLY detect what command has been executed.

Why?

The package shouldn't care *how* `cmake` was involved. It should consider in what environment it is being built - e.g. has ROS been sourced or not.

dirk-thomas:
Please elaborate on your use case *what* you want to achieve. Only then we can give feedback on *how* that should be achieved.

johnwason:
@Levi-Armstrong can you describe your requirements?

Levi-Armstrong:
If I remember correctly from the ROS-I developers meeting. The repository is mainly a pure CMake repository, but provides a ROS wrapper node for the repository. They currently pass there own build argument to indicate that it is being built for ROS. They are interested in a solution that remove the need for the user to provide a build flag and have it automatically detect if it is being build with catkin or ament. It was asked if they could leverage environment variables to detect this but there was an issue, but maybe @johnwason can provide more details on why the environment variables were not sufficient.

Levi-Armstrong:
@johnwason What repository are you currently working with, it may help to see how you are currently building things?

johnwason:
The main use case is for libraries that have ROS specific options that shouldn't be enabled except for in a ROS workspace. Catkin is nice because it passes a lot of information to the project being built, but this is not available in pure CMake builds. For users that have limited computer skills, having the packages be able to automatically configure is critical. If a user attempts to build portable software outside of a workspace it may still wind up with the ROS capabilities enabled even if it isn't intended.

Homebrew for Mac and vcpkg for Windows handle this scenario better than ROS does currently. These build systems have a separate control/formula file that can make decisions about how the package is built independently from the CMakeLists.txt file. A similar approach may be something worth considering.

dirk-thomas:
> Catkin is nice because it passes a lot of information to the project being built, but this is not available in pure CMake builds.

If your non-ROS build environment doesn't have `catkin` why don't you use that difference to distinguish the two environments?

```
find_package(catkin)
if(catkin_FOUND)
  ...
else()
  ...
endif()
```

Levi-Armstrong:
Based on @dirk-thomas response, I found another example of a [repository](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/master/CMakeLists.txt#L94-L123) leveraging his suggestion.

NikolausDemmel:
Just throwing my experience with a use case we had at my previous job: We had a project that was usually built standalone with plain cmake, but we wanted it to seamlessly behave like catkin package if built in a catkin workspace.

Since developers that had ROS installed (and sourced in bashrc) also needed the plain cmake build, finding catkin didn't work to distinguish a catkin and plain cmake build, since find_package(caktin) would succeed in both cases. I don't actually remember how we solved it back then, but seems to me explicitly passing some info about the build tool (catkin_make, catkin build, colcon) to cmake is a reasonable way to do that...

@dirk-thomas wrote:

> The package shouldn't care how cmake was involved. It should consider in what environment it is being built - e.g. has ROS been sourced or not.

Considering my use case, I would disagree here. Maybe it is not very common to have package work with and without ROS, but I would argue that invoking cmake manually outside a catkin / colcon workspace and expecting a ROS package to build as a ROS package is even more exotic (at least for a non-existing build dir; for existing build dirs, re-invoking cmake directly is common, e.g. from within IDEs like QtCreator). But I could be wrong of course.

dirk-thomas:
Lets give the two build modes of your package names to distinguish them:
1. a "standalone mode" (as you called it above)
2. a "ROS mode"

I assume they differ somehow in what they install and/or where they install artifacts.

In my opinion the choice of how you invoke the build of your package isn't coupled to your desired build mode.

If the user wants to trigger the behavior of build mode 1 he can still choose to:
* invoke plain `cmake / make / make install` directly or
* use `colcon` which only adds some convenience but does the same under the hood

The same goes for the case where the build mode 2 is desired. The user could still choose either of the pipelines. While the manual invocation of CMake would be cumbersome for a workspace containing multiple packages the user could still do it and get the same result.

So the ultimate choice doesn't come from the tool but from the desire of the user to trigger a specific behavior. If there is nothing in the environment (like the presence of `catkin`) which can make that decision I would highly recommend to explicitly pass that decision (using an environment variable, a CMake definition or something like that).

---

In contrast to the proposed definitions I could see an argument to expose the used tool in an environment variable. Than every developer can choose to use that in a detection logic if they like to (even if I think it would be the wrong choice in this case). This would be similar to what other services do:

* Travis CI set `TRAVIS=true` ([docs](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables))
* AppVeyor sets `APPVEYOR=true` ([docs](https://www.appveyor.com/docs/environment-variables/))

In a similar fashion `colcon` could set `COLCON=true` in all subprocesses it invokes? Similar `CATKIN_MAKE` / `CATKIN_MAKE_ISOLATED` / `CATKIN_TOOLS` could be set.

NikolausDemmel:
> If the user wants to trigger the behavior of build mode 1 he can still choose to:

I see your point. Back then there was no colcon and invoking a plain cmake build within a multi-package workspace (without an explicit manifest file) wasn't a thing.

Choosing the more common case as default and having an explicitly set flag or environment variable to switch to the other mode is maybe in general the sanest option.

However, I like your environment variable proposal. It still allows users to make some decision based on the build tool, if their specific setup allows / requires to make a reasonable decision based on that. So maybe for some company workflow you have "manual cmake --> mode 1, colcon/catkin_tools --> mode 2". At least that was our workflow (without colcon).

Just out of curiosity, for such a multi-mode package with a ROS1 `package.xml`, that would by default be treated like a normal ROS1 package (with colcon or catkin_tools), can you easily make colcon treat it like a plain cmake package without changing / removing the manifest file? I guess that would e.g. influence the build order w.r.t. other packages.

wjwwood:
I think the idea of each tools setting an environment variable for all subprocesses is the best option, e.g. `COLCON=true`, `CATKIN_MAKE=true`, `CATKIN_TOOLS=true`, and so on...

dirk-thomas:
@johnwason Could you update this PR with the discussed proposal? I guess the easiest way to achieve this is to add the key/value to `os.environ` as soon as the script is invoked...

Also PRs for the other build tools would be highly appreciated.

johnwason:
I can do the updates next week

I'd prefer to have an environmental variable like `ROS_WORKSPACE_BUILD_TOOL=catkin_make` etc instead of a variable for each build tool type.

dirk-thomas:
> I can do the updates next week

Great, thanks.

> I'd prefer to have an environmental variable like ROS_WORKSPACE_BUILD_TOOL=catkin_make etc instead of a variable for each build tool type.

Please use a tool specific variable instead. There is no reason to introduce a ROS specific environment variable for this.

johnwason:
The colcon build uses a ROS plugin, which is distinct from vanilla colcon. Isn't colcon intended to be general purpose? Something like `COLCON_ROS` would make sense there, otherwise it wouldn't help in detecting a ROS build.

Maybe a variable like `ROS_PACKAGE_XML_FILENAME` would make more sense than specifying the build tool? This would allow the CMake package to have access to the package.xml file which could then be parsed for additional ROS specific options.

dirk-thomas:
> The colcon build uses a ROS plugin, which is distinct from vanilla colcon.

I guess you mean vanilla CMake?

> Something like `COLCON_ROS` would make sense there, otherwise it wouldn't help in detecting a ROS build.

You want an environment variable to indicate the extension within `colcon` which is processing the package? I don't think that is a good idea. What if someone comes up with an improved subclass like `colcon-ros-nicer`? All of a sudden your logic in the package will fail again.

> This would allow the CMake package to have access to the package.xml file

The `package.xml` file is in the root of the package directory. You can check for its existence and parse it without the need for an environment variable.

dirk-thomas:
@johnwason Are you still interested in adding environment variables specific to the build tool? If not, this ticket should be closed.

johnwason:
I don't think the currently suggested environmental variables are specific enough to detect that the user wants a ROS build. @NikolausDemmel describes the scenario well, and this is an increasingly common problem as the ROS-I teams are leaning toward creating ROS optional packages with pure CMake build configurations. The users of ROS-I software tend to not have the experience to understand the subtleties of the build system so it needs to be as plug-and-chug as possible.

dirk-thomas:
I will go ahead and close this pull request then since in the current form the patch doesn't make sense imo. Please feel free to continue discussing the topic here (or on a new issue) to figure out how to support your use case.

Issue #1006, opened 2019-04-30
Tried to find Eigen3::Eigen_interface_link_libraries-NOTFOUND


Levi-Armstrong:
I have a pure cmake package that I am leveraging in a ros package. I switched the pure cmake package to using the target Eigen3::Eigen when linking but I am now getting the following build error from catkin `tried to find library Eigen3::Eigen_interface_link_libraries-NOTFOUND`. I believe issue is when we check if it is an interface library shown below it calls `get_target_property` but it is not checking if it returns `NOTFOUND` so it is adding an incorrect item to list.

https://github.com/ros/catkin/blob/f9a573fcdc9e85f5f5ef99e9f9b04374096d404b/cmake/catkin_libraries.cmake#L138-L142

Levi-Armstrong:
To verify that this is the case, I added the following to my ROS package CMakeLists.txt file.
```
find_package(Eigen3 REQUIRED)
get_target_property(test_type Eigen3::Eigen TYPE)
message(${test_type})
get_target_property(test_imported Eigen3::Eigen IMPORTED)
message(${test_imported})

if(test_type STREQUAL "INTERFACE_LIBRARY")
    get_target_property(test_interface_link_libraries Eigen3::Eigen INTERFACE_LINK_LIBRARIES)
    catkin_replace_imported_library_targets(test_resolved_libs ${test_interface_link_libraries})
    message(${test_resolved_libs})
endif()
```
Terminal output:

```
Warnings << tesseract_rosutils:check /home/larmstrong/catkin_ws/trajopt_only_ws/logs/tesseract_rosutils/build.check.019.log                                                                                                                    
INTERFACE_LIBRARY
TRUE
test_interface_link_libraries-NOTFOUND
```

dirk-thomas:
Please consider to provide a pull request for this issue.

Levi-Armstrong:
@dirk-thomas I just submitted a PR that addresses this issue.

dirk-thomas:
Addressed by #1013.

PR #1013, opened 2019-06-30
Fix interface targets that do not have libraries

Levi-Armstrong:
This is to address issue #1006


Levi-Armstrong:
It looks travis passed on 2.7 but for 3.4 it was unable to download something it required.

dirk-thomas:
@Levi-Armstrong Please rebase your branch against the latest state of the target branch in order to pass CI.

Levi-Armstrong:
@dirk-thomas I just rebased the branch.

dirk-thomas:
Thanks for the patch.

Please use the following phrase in the future when referencing the issue a PR addresses since it will make sure the ticket gets closed automatically when the PR is merged:

> Fixes #1006

PR #1014, opened 2019-07-05
Correctly set gtest 1.8 root directory

dgossow:
In gtest version 1.8 or higher, there is a CMakeLists.txt in the
parent directory of the gtest source directory which we need to find.
This can be either /usr/src/googletest or ${CMAKE_SOURCE_DIR}/googletest.

The current version would only check in /usr/src/googletest even if
gtest was found in the project's src/ directory.

This would break the build on Ubuntu 18.04, where the system install of gmock
is version 1.8, if your project includes a newer version of gtest in its src/ directory.
The result would be that headers from src/ would be used, but the test
executables would be linked to the binaries from /usr/src.

### DIFF
--- a/cmake/test/gtest.cmake
+++ b/cmake/test/gtest.cmake
@@ -280,9 +280,12 @@ function(catkin_find_google_test_source include_path src_path
   #In googletest 1.8, gmock and gtest where merged inside googletest
   #In this case, including gmock builds gmock twice, so instead we need to include googletest
   #which will include gtest and gmock once
-  set(_global_base_dir "${src_path}/googletest")
-  if(EXISTS "${_global_base_dir}/CMakeLists.txt")
-    set(${base_dir} ${_global_base_dir} PARENT_SCOPE)
+  if(_gtest_found)
+    get_filename_component(_gtest_base_dir_realpath ${_gtest_base_dir} REALPATH)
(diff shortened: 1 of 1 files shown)


dgossow:
@wjwwood @dirk-thomas Could you take a look at this?

dgossow:
The CI build fails like this:

```
Unable to download 3.4_with_system_site_packages archive. The archive may not exist. Please consider a different version.
```

Does not seem related though.

I still need to test this on Ubuntu 16.04 and 18.04 with and without gtest sources inside the workspace to make sure it works.

dirk-thomas:
> The CI build fails like this:

@dgossow Please rebase your change on top of the latest state of the target branch.

dirk-thomas:
@dgossow Friendly ping.

dgossow:
Sorry, I was distracted. So I essentially rebased this on top of kinetic-devel like requested.

I also tested this on 16.04 and 18.04.

16.04: 
gtest found under /usr/src/gmock
gmock found under /usr/src/gmock

18.04:
gtest found under /usr/src/googletest
gmock found under /usr/src/googletest

Finding the current version (master) of gmock/gtest under src/googletest works in both cases.

dgossow:
@dirk-thomas comments addressed

dirk-thomas:
Thanks for the patch.

PR #1015, opened 2019-07-05
[Windows][kinetic-devel] Fixed colcon build issues

seanyen:
Two issues are found when building `catkin`-based projects on Windows by `colcon`.

* For MSBUILD generator (which is selected by `colcon`), it will place binaries to a relative location like `bin\Debug\` or `bin\Release\` by `CMAKE_BUILD_TYPE` and it is not an expected FHS by ROS, so I explicitly define the output directories for each different build types.

* The path from `CATKIN_WORKSPACES` are eventually passed to CMake functions which expect CMake-style paths. However, `colcon` passed them as the native paths on Windows (backward slash), and it causes build breaks, so I normalize the path when `catkin` constructs `CATKIN_WORKSPACES`.


seanyen:
The travis CI failure seemed not to do with this PR:
```
3.4_with_system_site_packages is not installed; attempting download
Downloading archive: https://storage.googleapis.com/travis-ci-language-archives/python/binaries/ubuntu/16.04/x86_64/python-3.4_with_system_site_packages.tar.bz2
0.14s$ curl -sSf -o python-3.4_with_system_site_packages.tar.bz2 ${archive_url}
curl: (22) The requested URL returned error: 404 Not Found
```

seanyen:
https://github.com/ms-iot/ROSOnWindows/issues/84

cmake/all.cmake
```diff
@@ -57,10 +57,11 @@ set(CMAKE_PREFIX_PATH_AS_IS ${CMAKE_PREFIX_PATH})
 # list of unique catkin workspaces based on CMAKE_PREFIX_PATH
 set(CATKIN_WORKSPACES "")
 foreach(path ${CMAKE_PREFIX_PATH})
-  if(EXISTS "${path}/.catkin")
-    list(FIND CATKIN_WORKSPACES ${path} _index)
+  file(TO_CMAKE_PATH "${path}" cmake_path)
+  if(EXISTS "${cmake_path}/.catkin")
```
dirk-thomas: Please keep this condition as is and instead move the `file(TO_CMAKE_PATH ...)` line into the if block since it is only used there.

dirk-thomas (commented):
Please also rebase against the latest state of `kinetic-devel` to pass CI.

seanyen:
@dirk-thomas I re-based this pull request and address the feedback. Please take a look. Thanks!

dirk-thomas:
The patch looks good and CI is green :+1:

> Two issues are found when building catkin project on Windows by colcon.
> For MSVC, by default it will place binaries to a relative location like bin\Debug\ or bin\Release\ by CMAKE_BUILD_TYPE ...

Out of curiosity: how is this specific to `colcon`? I would expect the same behavior with `catkin_make(_isolated)`.

seanyen:
> Out of curiosity: how is this specific to `colcon`? I would expect the same behavior with `catkin_make(_isolated)`.

Here is how I kick off a colcon build: (You might notice I use ` --cmake-args` to specify some custom flags for the build.)
https://github.com/ms-iot/ros-windows-build/blob/master/ros-colcon-build/build.bat#L21

dirk-thomas:
What happens if you use the same arguments for `catkin_make_isolated`?

seanyen:
@dirk-thomas If I directly pass the arguments to `catkin_make_isolated`, it won't be a problem because it is either using `NMake Makefiles` or `Ninja` as the generator on Windows, and which honors the output directory variables (without the per-config suffix). However, when I kick off the build from `colcon`, `Visual Studio 16 2019` is chosen there, and in this case, output directories like `RUNTIME_OUTPUT_DIRECTORY_DEBUG` needs to be overridden to get rid of the subfolder `Deubg\Release` created by `MSBUILD`.

I cannot find any comprehensive doc from CMake on this, but here are some related discussion and solution for this issues:
* https://stackoverflow.com/questions/543203/cmake-runtime-output-directory-on-windows
* https://cmake.org/Bug/view.php?id=9163

dirk-thomas:
Thanks for the additional context. That was what I was looking for.

PR #1016, opened 2019-07-08
Fix python 3 version used by travis

sloretz:
Travis now uses [xenial](https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment) as the default build environment. The CI uses system site packages, but this option only works with the default python 3 version. In xenial this version is [python 3.5](https://packages.ubuntu.com/xenial/python3).

This PR bumps the python version to 3.5, and uses `system_site_packages: true` since [that's what the travis documentation does](https://docs.travis-ci.com/user/languages/python/#travis-ci-uses-isolated-virtualenvs).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
 language: python
 python:
-  - "2.7_with_system_site_packages"
-  - "3.4_with_system_site_packages"
+  - "2.7"
+  - "3.5"
+virtualenv:
+  system_site_packages: true
 # command to install dependencies
 install:
   - pip install argparse catkin-pkg empy mock nose


sloretz:
Travis build passes: https://travis-ci.org/ros/catkin/builds/555875881

sloretz:
I'll merge this without waiting for an approval since it only changes `.travis.yml`, and appears to resolve the python 3 job issue.

If I overlooked something the worst case is it makes the travis CI on the `kinetic-devel` branch invalid. This wouldn't be a big deal because the branch is already covered by ROS PR jobs for python 2, and the python 3 travis job does not work without this PR.

PR #1020, opened 2019-07-30
fix catkin_make on Windows

kejxu:
on Windows, `catkin_make` fails without executing `catkin_init_workspace` first for a workspace that contains `catkin`:

```
src
  |- catkin
  |- pkg1
  |- ...
```

When the workspace is built for the first time, there is no top-level `CMakeLists.txt` under `src`, so `catkin` would attempt to find and copy from an existing source. When `catkin` is also in the workspace, `init_workspace.py` goes to https://github.com/ros/catkin/blob/kinetic-devel/python/catkin/init_workspace.py#L88, and by calling

```python
src_file_path = os.path.relpath(src, workspace_dir)
```

the source path is assigned to this relative path: `catkin/cmake/toplevel.cmake`. This works with `os.symlink()`; however, since `os.symlink()` is not available on Windows, `_symlink_or_copy()` would instead attempt to copy from `catkin/cmake/toplevel.cmake`. Since `catkin_make` is usally called from a level higher, this would fail since the relative path points to nowhere.

Changing to use absolute path on Windows to ensure proper copying.

### DIFF
--- a/python/catkin/init_workspace.py
+++ b/python/catkin/init_workspace.py
@@ -87,7 +87,7 @@ def init_workspace(workspace_dir):
     # look in to-be-initialized workspace first
     src = os.path.join(workspace_dir, 'catkin', 'cmake', 'toplevel.cmake')
     if os.path.isfile(src):
-        src_file_path = os.path.relpath(src, workspace_dir)
+        src_file_path = src
     else:
         checked.append(src)
 
--- a/python/catkin/init_workspace.py
+++ b/python/catkin/init_workspace.py
@@ -33,6 +33,7 @@
 from __future__ import print_function
 import os
 import shutil
+import sys
 from catkin.workspace import get_source_paths, get_workspaces
 
 def _symlink_or_copy(src, dst):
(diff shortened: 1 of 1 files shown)


kejxu:
reworking

kejxu:
since there is not yet native support from Python to create symlink on Windows, return absolute path and copy on Windows

this potentially causes error when the 2 `CMakeLists.txt`s go out of sync. (just throwing it out there, maybe the build tool should be decoupled from the rest of the source code so this would actually never happen?) However, the same thing could happen even if the top-level `CmakeLists.txt` is copied from another workspace; when that workspace (e.g. global installation) gets updated, the 2 files eventually become out of sync.

Ideally, `catkin` probably needs to do a file comparison (e.g. checksum) each time before consuming the file, but that might not be really necessary since the build tool is gradually moving towards `colcon`.

ooeygui:
This fix looks good. 
The workaround for this is to manually initialize the catkin work space by calling catkin_init_workspace.

python/catkin/init_workspace.py
```diff
@@ -87,7 +88,11 @@ def init_workspace(workspace_dir):
     # look in to-be-initialized workspace first
     src = os.path.join(workspace_dir, 'catkin', 'cmake', 'toplevel.cmake')
     if os.path.isfile(src):
-        src_file_path = os.path.relpath(src, workspace_dir)
+        if sys.platform.startswith('win32'):
```
dirk-thomas: Other code in this repos checks for `sys.platform == 'win32'`. When does it ever contain a longer string starting with `win32`?

kejxu: according to https://docs.python.org/2/library/sys.html#sys.platform, `sys.platform` would not resolve to `win32...`; this is a style mismatch, changing to `sys.platform == 'win32'`

`.startswith` is only needed when checking for `linux`

dirk-thomas:
Thanks for the patch!

PR #1022, opened 2019-08-03
Gmock from source on Debian buster

sloretz:
This fixes some issues with gmock tests on Debian buster. The `gtest` rosdep key [resolves to libgtest-dev](https://github.com/ros/rosdistro/blob/cf147fd20c209c580aa2ea5221f1f8b561940a1c/rosdep/base.yaml#L1218) on Debian buster. This package [depends on `googletest`](https://packages.debian.org/buster/libgtest-dev), and also installs gtest headers to `/usr/include`. The package `googletest` includes both gtest and gmock, as source files in `/usr/src/googletest/`. Additionally there is `FindGTest.cmake` module installed by the package `cmake-data`. This situation tricks the current logic.

First the search paths here weren't considering that gtest/gmock could be installed in `/usr/src`. This PR adds those paths.

Second Because `GTest` in `/usr/include` is found by the cmake module, the logic sees `GTEST_FOUND` but not `GMOCK_FOUND`. The `find_package()`'d gtest is used gmock is considered unavailable, but gmock would be available if this code tried to find its sources. This PR refactors the logic to the following:

1. If both gtest and gmock are found using `find_package()`, use that
1. If neither gtest or gmock are found, try to find it them from source and use whatever is found
1. If only 1 of gtest or gmock is found, try to find both from source. If both are found from source, use that. If only one is found from source, use what was found by the `find_package()` calls.
1. Finally try to create `gtest`, `gtest_main`, `gmock` and `gmock_main` targets if they don't exist no matter where we're using gtest/gmock from.


lromor:
Thank you for this patch!

GladOSkar:
Thanks, works like a charm :)

dirk-thomas:
Thanks for the patch.

cmake/test/gtest.cmake
```diff
@@ -287,118 +291,131 @@ function(catkin_find_google_test_source include_path src_path
...
+find_package(GTest QUIET)
+if(NOT GMOCK_FOUND OR NOT GTEST_FOUND)
+  # If we find one but not the other, see if we can get both from source
+  # only add gmock/gtest directory once per workspace
+  if(NOT TARGET gtest AND NOT TARGET gmock)
+    # Fall back to system-installed gmock source (e.g. Ubuntu)
+    set(_include_paths "/usr/include" "/usr/src")
+    set(_source_paths "/usr/src")
+
+    catkin_find_google_test_source("${_include_paths}" "${_source_paths}" gtest_found
```
dirk-thomas, line 303: @sloretz As pointed out in https://github.com/ros/catkin/issues/1030#issuecomment-540263776 passing a list of include paths to `catkin_find_google_test_source` seems wrong. The function only expects a single path  - e.g. it appends a subdirectory to it (`set(_gtest_include_paths "${include_path}/gtest")`). So either:
* only a single path should be passed or
* the function needs to be updated to append the subdirectory to each item of the list or
* it needs to be clear why only appending the suffix to only the last item is intentional.

I am not sure which of the three options is the right way of resolving this.

sloretz: > the function needs to be updated to append the subdirectory to each item of the list or

I think this would be the correct fix

dirk-thomas: Sounds good to me. Please create a PR with the change.

dirk-thomas: See #1040.

PR #1025, opened 2019-08-08
use condition attributes to specify Python 2 and 3 dependencies

dirk-thomas:
Uses condition attributes introduced in format 3 to declare Python 2 specific as well as Python 3 specific dependencies.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -1,12 +1,15 @@
 <?xml version="1.0"?>
-<package format="2">
+<?xml-model
+  href="http://download.ros.org/schema/package_format3.xsd"
+  schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
   <name>catkin</name>
   <version>0.7.18</version>
   <description>Low-level build system macros and infrastructure for ROS.</description>
   <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
   <license>BSD</license>
 
-  <url type="website">http://www.ros.org/wiki/catkin</url>
+  <url type="website">http://wiki.ros.org/catkin</url>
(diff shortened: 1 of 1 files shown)


dirk-thomas:
@ros-pull-request-builder retest this please

dirk-thomas:
@ros-pull-request-builder retest this please

dirk-thomas:
With ros-infrastructure/ros_buildfarm_config#138 the PR jobs pass now.

package.xml
```diff
@@ -15,21 +18,24 @@
   <author>Brian Gerkey</author>
   <author>Dirk Thomas</author>
 
+  <depend condition="$ROS_PYTHON_VERSION == 2">python-argparse</depend>
+  <depend condition="$ROS_PYTHON_VERSION == 2" version_gt="0.4.3">python-catkin-pkg</depend>
+  <depend condition="$ROS_PYTHON_VERSION == 3" version_gt="0.4.3">python3-catkin-pkg</depend>
+  <depend condition="$ROS_PYTHON_VERSION == 2">python-empy</depend>
+  <depend condition="$ROS_PYTHON_VERSION == 3">python3-empy</depend>
```
sloretz, line 25: Is `empy` needed at runtime? It looks like it was listed as `<build_depend>` + `<build_export_depend>` before.

dirk-thomas: `build_export_depend` is technically a run dependency. The only difference is that `build_export_depend` implies that it is used on the architecture the code is built on whereas `exec_depend` implies the target architecture. For Python packages (without C extensions) that distinction is irrelevant. Therefore I chose to simplify the dependency declaration.

sloretz (approved):

PR #1027, opened 2019-08-14
Add a blacklist option to make_isolated

### LINKED ISSUE #1026: [Request] Add a 'blacklist' option to catkin_make_isolated
While `catkin_make` has support for white/blacklist, `catkin_make_isolated` only supports whitelist through the `--pkg` option. A blacklist option comes handy when dealing with large workspace.

artivis:
Add a blacklist option to `catkin_make_isolated`, e.g.

```terminal
catkin_make_isolated --ignore-pkg foo bar
```

Fix #1026

### DIFF
--- a/bin/catkin_make_isolated
+++ b/bin/catkin_make_isolated
@@ -57,7 +57,7 @@ def parse_args(args=None):
     pkg.add_argument('--pkg', nargs='+', metavar='PKGNAME', dest='packages',
                      help="Process only specific packages "
                           "(only after catkin_make_isolated has been invoked before with the same install flag)")
-    pkg.add_argument('--ignore-pkg', nargs='+', metavar='PKGNAME', dest='package_blacklist',
+    pkg.add_argument('--ignore-pkg', nargs='+', metavar='PKGNAME', dest='ignore_packages',
                      help="Ignore specific packages.")
     pkg.add_argument('--from-pkg', metavar='PKGNAME', dest='from_package',
                      help='Restart catkin_make_isolated at the given package continuing from there '
@@ -148,7 +148,7 @@ def main():
         force_cmake=opts.force_cmake,
         colorize=not opts.no_color,
         build_packages=opts.packages or ([] if opts.from_package is None else [opts.from_package]),
-        blacklisted=opts.package_blacklist[0].split(';') if opts.package_blacklist is not None else None,
+        blacklisted=opts.ignore_packages,
         quiet=opts.quiet,
         cmake_args=cmake_args,
         make_args=opts.make_args,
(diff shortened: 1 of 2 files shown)


python/catkin/builder.py
```diff
@@ -971,7 +972,7 @@ def build_workspace_isolated(
             sys.exit('Packages not found in the workspace: %s' % ', '.join(unknown_packages))
 
     # Report topological ordering
-    ordered_packages = topological_order_packages(packages)
+    ordered_packages = topological_order_packages(packages, None, blacklisted)
```
dirk-thomas: Please pass keyword arguments explicitly: `topological_order_packages(packages, blacklisted=blacklisted)`.

bin/catkin_make_isolated
```diff
@@ -146,6 +148,7 @@ def main():
         force_cmake=opts.force_cmake,
         colorize=not opts.no_color,
         build_packages=opts.packages or ([] if opts.from_package is None else [opts.from_package]),
+        blacklisted=opts.package_blacklist[0].split(';') if opts.package_blacklist is not None else None,
```
dirk-thomas: There is no need to split arguments by semicolon. Argparse will return a list already since the argument uses `nargs='+'`.

bin/catkin_make_isolated
```diff
@@ -57,6 +57,8 @@ def parse_args(args=None):
     pkg.add_argument('--pkg', nargs='+', metavar='PKGNAME', dest='packages',
                      help="Process only specific packages "
                           "(only after catkin_make_isolated has been invoked before with the same install flag)")
+    pkg.add_argument('--ignore-pkg', nargs='+', metavar='PKGNAME', dest='package_blacklist',
```
dirk-thomas: Please use `ignore_packages` as the `dest` to match the naming of the option.

python/catkin/builder.py
```diff
@@ -809,6 +809,7 @@ def build_workspace_isolated(
     force_cmake=False,
     colorize=True,
     build_packages=None,
+    blacklisted=None,
```
dirk-thomas: Use the same terminology here: `ignore_packages`.

dirk-thomas (commented):
Thanks for working on this patch.

Just as a note: `catkin_make_isolated` is not being developed anymore. So you might want to look into using alternatives (`colcon`, `catkin_tools`) which already offer this (and many more) features.

artivis:
@dirk-thomas Thanks for the review, PR updated accordingly.

python/catkin/builder.py
```diff
@@ -972,7 +972,7 @@ def build_workspace_isolated(
             sys.exit('Packages not found in the workspace: %s' % ', '.join(unknown_packages))
 
     # Report topological ordering
-    ordered_packages = topological_order_packages(packages, None, blacklisted)
+    ordered_packages = topological_order_packages(packages, whitelisted=None, blacklisted=ignore_packages)
```
dirk-thomas: As in my [above](https://github.com/ros/catkin/pull/1027#discussion_r314084468) example there is no need to pass `whitelisted=None`. Please remove it from the list of arguments.

artivis: Oups, done.

PR #1029, opened 2019-08-22
[windows][kinetic-devel] update a typo in local_setup.bat.in

seanyen:
On Windows, the file extension for `setup` script should be `.bat`.


dirk-thomas:
Thanks for the fix.

PR #1033, opened 2019-10-07
support for pytest 5.1+ jUnit result files

### LINKED ISSUE #1032: pytest 5.1.0 generates incompatible xml output for catkin
From pytest 5.1.0 (2019-08-15) the xml output of pytest changed [see this PR](https://github.com/pytest-dev/pytest/pull/5550). The output looks as follows:
```xml
<testsuites>
  <testsuite tests="5" failures="3" errors="1" skip="2" time="35" name="AllTests"></testsuite>
</testsuites>'
```
Note that the root element does not contain a summary, which catkin expects.

I would like to contribute a PR. Against which repo should it be, pytest or catkin?

dirk-thomas:
Fixes #1032.

### DIFF
--- a/python/catkin/test_results.py
+++ b/python/catkin/test_results.py
@@ -109,14 +109,43 @@ def read_junit2(filename):
     :param filename: str junit xml file name
     :returns: num_tests, num_errors, num_failures, num_skipped
     :raises ParseError: if xml is not well-formed
+    :raises ValueError: if root the node is neither a testsuites tag nor a testsuite tag
     :raises IOError: if filename does not exist
     """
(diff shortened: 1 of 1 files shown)


Rayman (approved):
I've tested this change with pytest 5.2.1 and it solves my problem. Thanks!

mathias-luedtke:
This fix does not seem to work with gtest results (example: https://github.com/ros-industrial/industrial_ci/issues/599#issuecomment-769643722) and counts all tests twice.
`<testsuites>` already summarizes the contained `<testsuite>` entries, so the code should not sum them up again.

PR #1034, opened 2019-10-08
Check for INTERFACE library type in add_library function …

johnwason:
defined in windows.cmake. Currently this function will fail with INTERFACE libraries.

### DIFF
--- a/cmake/platform/windows.cmake
+++ b/cmake/platform/windows.cmake
@@ -43,8 +43,9 @@ if(BUILD_SHARED_LIBS)
       # Check if its an external, imported library (e.g. boost libs via cmake module definition)
       list(FIND ARGN "IMPORTED" FIND_IMPORTED)
       list(FIND ARGN "ALIAS" FIND_ALIAS)
+	  list(FIND ARGN "INTERFACE" FIND_INTERFACE)
       _add_library(${ARGV0} ${ARGN})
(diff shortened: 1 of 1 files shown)


cmake/platform/windows.cmake
```diff
@@ -43,8 +43,9 @@ if(BUILD_SHARED_LIBS)
       # Check if its an external, imported library (e.g. boost libs via cmake module definition)
       list(FIND ARGN "IMPORTED" FIND_IMPORTED)
       list(FIND ARGN "ALIAS" FIND_ALIAS)
+	  list(FIND ARGN "INTERFACE" FIND_INTERFACE)
```
dirk-thomas: Please use spaces to indent the line to the same level as the surrounding.

dirk-thomas:
Thanks for patch.

Since a new release of catkin was made earlier this day it might be a while until the next patch release.

johnwason:
Since this only affects Windows maybe you should ask the Microsoft team to update their package.

dirk-thomas:
> Since this only affects Windows maybe you should ask the Microsoft team to update their package.

Please go ahead and do as suggested. I don't know how they are building packages or who the right person to notify is.

johnwason:
https://github.com/ms-iot/ROSOnWindows/issues/158

PR #1037, opened 2019-10-17
Correcly use ${prefix} variable in generated pkg-config

### LINKED ISSUE #1036: Generated .pc files should use ${prefix}
The pkg-config file generated by `catkin_package()` defines a `prefix` variable, but do not make use of it and repeat `@PROJECT_SPACE_DIR` multiple times in the file.

Here in the [template file](https://github.com/ros/catkin/blob/kinetic-devel/cmake/em/pkg.pc.em):

```em
prefix=@PROJECT_SPACE_DIR
//...
Libs: -L@PROJECT_SPACE_DIR/lib @(' '.join(PKG_CONFIG_LIBRARIES_WITH_PREFIX))
```
vs
```em
prefix=@PROJECT_SPACE_DIR
// ...
Libs: -L${prefix}/lib @(' '.join(PKG_CONFIG_LIBRARIES_WITH_PREFIX))
```

This matters, because pkg-config can fix relocated packages with the `--define-prefix` flag, which cause the `prefix` variable to be redefined. But it is unused by the generated files.

To fix it, we should simply replace `@PROJECT_SPACE_DIR` by `${prefix}` in the file, and when building the `PKG_CONFIG_LIBRARIES_WITH_PREFIX` and `PROJECT_PKG_CONFIG_INCLUDE_DIRS` variables.

Hugal31:
When generating the package files for the install space, correctly use
${prefix} instead of repeating the package root dir.

Fixes  #1036.

### DIFF
--- a/cmake/catkin_package.cmake
+++ b/cmake/catkin_package.cmake
@@ -401,7 +401,7 @@ function(_catkin_package)
   set(INSTALLSPACE TRUE)
 
   set(PROJECT_SPACE_DIR ${CMAKE_INSTALL_PREFIX})
-  set(PKG_INCLUDE_PREFIX ${PROJECT_SPACE_DIR})
+  set(PKG_INCLUDE_PREFIX "\\\${prefix}")
 
   # absolute path to include dir under install prefix if any include dir is set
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Thanks for the patch.

PR #1046, opened 2020-01-14
Update gtest config documentation

matthew-reynolds:
Following up on fkie/catkin_lint#76 and ros-controls/ros_control#408:

This doc is a bit out of date, and the `rosunit` dependency is no longer needed for the `catkin_add_XXXtest()` commands.

### DIFF
--- a/doc/howto/format1/gtest_configuration.rst
+++ b/doc/howto/format1/gtest_configuration.rst
@@ -8,14 +8,6 @@ pure C++ framework.  No ROS environment is available.  See:
 :ref:`rostest_configuration` if your tests need a running roscore_.
 
 
-package.xml
-:::::::::::
-
-The rosunit_ package is needed for testing::
-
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Please update the same docs in the sibling directory for `format1` accordingly.

matthew-reynolds:
Done in 41b6692. (Force pushed to squash changes to both files into the same commit)

dirk-thomas:
Thanks for the patch.

> Force pushed to squash changes to both files into the same commit

Don't worry about this in the future. Most PRs are being squash merged anyway. And when PRs get bigger and are being iterated on it is usually more valuable to be able to only review the incremental change since the last review.

matthew-reynolds:
> Don't worry about this in the future. Most PRs are being squash merged anyway

Good point, had my mindset on ros_control where we haven't been squashing PRs.

Thanks for the speedy review & merge :)

PR #1048, opened 2020-01-17
Prefer setuptools with Python 3

sloretz:
This is a follow up to https://github.com/ros/genmsg/pull/90#issuecomment-575702761

Currently catkin requires `distutils` and optionally supports `setuptools`; however, `distutils` for Python 3 requires a dependency on `python3-distutils` in Ubuntu Focal and Debian Buster. This PR makes catkin prefer setuptools and optionally support distutils when using Python 3.

* catkin itself uses setuptools in its own setup.py
* When interrogating another package's `setup.py`, don't fail if `distutils.core` can't be imported
* When using Python 3, add a `buildtool_export_depend` on `setuptools` since something like it must be available downstream
* Update documentation to use `setuptools` instead of `distutils.core`.

With this PR I think the effect on Python 2 ROS distros is `python-setuptools` would be installed where it wasn't before, but no downstream changes are required. The effect on Noetic is downstream packages depending on `catkin` either need to depend on `python3-distutils` or use `setuptools` instead.

### DIFF
--- a/doc/user_guide/setup_dot_py.rst
+++ b/doc/user_guide/setup_dot_py.rst
@@ -12,11 +12,6 @@ packages define the installation files in a file called ``setup.py``
 in the project root. The setup.py file uses Python to describe the
 Python content of the stack.
 
-We recommend to prefer distutils package over setuptools/distribute,
-because with distutils we can avoid the creation of egg-info folders
-in the project source folder. The setup.cfg file of distutils2 is not
-supported by catkin.
-
 Catkin allows you to specify the installation of your python files in
 this setup.py and reuse some of the information in your CMakeLists.txt.
 
@@ -26,10 +21,10 @@ You can do so by including the line::
 
 in the CMakeLists.txt of your project.
 
-catkin will execute setup.py with a hot-patched version of distutils
-to read the arguments to set up the devel space, and execute setup.py
-with suitable arguments to install to the catkin install space under
-``CMAKE_INSTALL_PREFIX``.
+catkin will execute setup.py with a hot-patched version of distutils or
+setuptools to read the arguments to set up the devel space, and execute
+setup.py with suitable arguments to install to the catkin install space
+under ``CMAKE_INSTALL_PREFIX``.
 
 This means that if you execute your
 setup.py using::
@@ -50,7 +45,7 @@ pypi is not very useful for ROS nodes.
 For the develspace, the following setup.py arguments to setup() will
 be used by catkin::
 
-  from distutils.core import setup
+  from setuptools import setup
 
   setup(
       version='...',
@@ -101,7 +96,7 @@ Writing a setup.py file without duplicating information contained in
 the package.xml is possible using a catkin_pkg convenience function
 like this::
 
-  from distutils.core import setup
+  from setuptools import setup
   from catkin_pkg.python_setup import generate_distutils_setup
 
   d = generate_distutils_setup(
@@ -130,7 +125,7 @@ one distributes to pypi.
 .. note::
 
   See the note in previous section about the otherwise useful
-  field ``requires`` that's usually a part of distutils setup.
+  field ``requires`` that's usually a part of setuptools setup.
   *Do not use it* in ROS.
 
 
--- a/cmake/interrogate_setup_dot_py.py
+++ b/cmake/interrogate_setup_dot_py.py
@@ -37,7 +37,7 @@ import sys
 
 setup_modules = []
 
-try
+try:
     import distutils.core
     setup_modules.append(distutils.core)
 except ImportError:
--- a/cmake/interrogate_setup_dot_py.py
+++ b/cmake/interrogate_setup_dot_py.py
@@ -35,11 +35,21 @@ import os
 import runpy
 import sys
 
-import distutils.core
-try:
-    import setuptools
+setup_modules = []
+
+try
+    import distutils.core
+    setup_modules.append(distutils.core)
 except ImportError:
     pass
+
+try:
+    import setuptools
+    setup_modules.append(setuptools)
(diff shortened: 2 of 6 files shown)


sloretz:
CI failure on Noetic appears to be due to CMake warning unrelated to this PR: http://build.ros.org/job/Npr__catkin__ubuntu_focal_amd64/4/warnings2Result/new/

doc/user_guide/setup_dot_py.rst
```diff
@@ -12,11 +12,6 @@ packages define the installation files in a file called ``setup.py``
 in the project root. The setup.py file uses Python to describe the
 Python content of the stack.
 
-We recommend to prefer distutils package over setuptools/distribute,
-because with distutils we can avoid the creation of egg-info folders
```
dirk-thomas, line 16: Does this limitation still apply with `setuptools`?

dirk-thomas: @sloretz Friendly ping.

sloretz: Maybe not? I built all Noetic repos with source entries using Buster. The only three using `setuptools` in my workspace are `catkin`, `genmsg`, and `python_qt_binding`. After building with this branch all three have clean source spaces, and `find . -name "*egg*"` in `src` shows no files. I assume that means whatever avoids `egg-info` folder creation was able to do so.

dirk-thomas (approved):

mikepurvis:
This change is causing a bit of chaos for us as the install directory has changed from `lib/python3/dist-packages/catkin` to `lib/python3/dist-packages/catkin-0.8.0-py3.5.egg`, with an `lib/python3/dist-packages/easy-install.pth` file generated that adds the egg path to `sys.path` on interpreter startup.

Part of the reason this is problematic is because we build in isolation with custom post-build merging, but even in a more conventional build scenario, there's a conflict over this file as it could be provided by multiple packages with varying contents.

Switching to `setuptools` makes sense, but I think we should be passing it the `--single-version-externally-managed` flag.

FYI @tspicer01

doc/user_guide/setup_dot_py.rst
matthew-reynolds: `egg-info` folders are still created when building Melodic packages w/ python 2 on Bionic (See https://travis-ci.org/github/ros-controls/ros_control/jobs/661982516?utm_medium=notification&utm_source=github_status)

Is this expected?

mikepurvis: That's weird that it tried to create it in the source. In all my testing, the `egg-info` folder always ends up in the installspace.

matthew-reynolds: I see the same thing (`egg-info` folders in the source space) when building locally as well, for all packages that have been migrated to `setuptools`. I tested `angles`, `ros_control`, and `catkin` itself.

sloretz: ~~I don't think that's expected, and I also wouldn't expect the behavior to have been changed by this PR. [`controller_manager_msgs` is still using `distutils` on `melodic-devel`](https://github.com/ros-controls/ros_control/blob/f211ee381d4f40967db980601e9a94ac40f3c8c0/controller_manager_msgs/setup.py#L3).~~ **Edit**: Nevermind, I see the PR being tested is using setuptools.

sloretz: @matthew-reynolds maybe [`python_distutils_install.sh`](https://github.com/ros/catkin/blob/86439ec5d2010d5c47c30b815aa97e525037930f/cmake/templates/python_distutils_install.sh.in#L24-L32) needs to set [`--egg-base`](https://setuptools.readthedocs.io/en/latest/setuptools.html#other-egg-info-options) ? I'm not sure that option is also available in `distutils` though.

mikepurvis: The svem part is no longer relevant, but if you need to resuscitate the notion of setuptools-specific flags, the pieces of this diff that detect it may be useful:

https://github.com/ros/catkin/compare/noetic-devel...mikepurvis:add-setuptools-svem

sloretz: @matthew-reynolds would you mind opening an issue with steps to reproduce? I'm unable to reproduce using ROS Melodic on Ubuntu Bionic with :

```
cd /tmp
mkdir -p mel_ws/src
cd mel_ws/src
git clone git@github.com:ros/angles.git
chmod a-w -R angles
cd /tmp/mel_ws
catkin_make_isolated
catkin_make
```

I was expecting `catkin_make_isolated` or `catkin_make` to fail, but they both succeeded. If they tried to create `egg_info` folders in the source directory then they should get a permission denied error since it's read-only.

matthew-reynolds: Interesting - I did a little more testing, this seems to only occur when installing.

`catkin_make install`, `catkin_make_isolated --install`, and `catkin config --install; catkin build` all fail.

```
cd /tmp
mkdir -p mel_ws/src
cd mel_ws/src
git clone git@github.com:ros/angles.git
chmod a-w -R angles
cd /tmp/mel_ws
catkin_make install                                 # Fails
catkin_make_isolated --install                      # Fails
catkin clean; catkin config --install; catkin build # Fails
```

I will open an issue to continue this conversation :+1:

matthew-reynolds: #1069

ros-discourse:
This pull request has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/ros-1-and-python-3-10s-deprecation-of-distutils-core/29834/15

PR #1050, opened 2020-01-20
Fix if statement in catkin_libraries

poggenhans:
Currently, when `catkin_package(LIBRARIES ...)` is called with an interface library that has more than one `INTERFACE_LINK_LIBRARIES`, catkin chokes on it with the error:

```
CMake Error at /opt/mrtros/share/catkin/cmake/catkin_libraries.cmake:140 (if):
  if given arguments:

    "library1" "library2"

  Unknown arguments specified
```

This PR fixes the syntax error that causes this problem. All followings commands seem to be able to handle multiple values in `${${lib}_interface_link_libraries}`.


dirk-thomas:
Thank you for the fix.

Issue #1047, opened 2020-01-17
Failure passing MAKEFLAGS to catkin_make_isolated on FreeBSD


dodsonmg:
I'm attempting to install ROS Kinetic on FreeBSD 12.1 (at my peril).

I'm following the 'from source' instructions at:
http://wiki.ros.org/kinetic/Installation/Source

I am performing step 2.1.3 (Building the catkin workspace), which directs invoking `catkin_make_isolated` as follows:
`./src/catkin/bin/catkin_make_isolated --install -DCMAKE_BUILD_TYPE=Release`

I get the following output:
```
==> Processing catkin package: 'catkin'
...
==> make -j4 -l4 in '$PATH/ros_catkin_ws/build_isolated/catkin'
...
<== Failed to process package 'catkin':
Command '['make', '-j4', '-l4']' returned non-zero exit status 2
```

This seems to be due to FreeBSD `make` not having the loading average flag `-l`.

I have tried both setting the `MAKEFLAGS` and `ROS_PARALLEL_JOBS` environment variables to avoid setting the `-l` flag.  E.g., `set ROS_PARALLEL_JOBS='-j4'`.

The result is that it sends a flag `-pn` to `make`, as follows:
```
==> Processing catkin package: 'catkin'
...
==> make -j4 in '$PATH/ros_catkin_ws/build_isolated/catkin'
...
<== Failed to process package 'catkin':
Command '['make', '-pn']' returned non-zero exit status 2
```

Interestingly, it appears that it's invoking `make -j4`, but then says that the actual flag sent to `make` is `-pn`.

I have also attempted to set the flags when invoking `catkin_make_isolated`, but get the same result (with `-pn`).  For example:
`./src/catkin/bin/catkin_make_isolated --install -DCMAKE_BUILD_TYPE=Release -j4`

dirk-thomas:
The make flags `-j` and `-j` are only added conditionally, see https://github.com/ros/catkin/blob/4d4467502337a76b23a14372cc6659f12ee68d5b/python/catkin/builder.py#L323-L325. If you pass job flags in any other way the logic won't be triggered (and doesn't add the `-l` flag which isn't supported on your platform).

Possible options:
* you pass `--make-args -j4`
  * (the argument must contain a valid job argument otherwise the default logic might get triggered)
* set `MAKEFLAGS=-j4` (the command won't show these since they are set as an environment variable)
  * (the argument must contain a valid job argument otherwise the default logic might get triggered)
[...]

PR #1051, opened 2020-01-22
Support use of gmake on FreeBSD

dodsonmg:
FreeBSD `make` is not GNU `make`, and doesn't support certain commands used by `catkin` (specifically, `-l` and `-p`).

GNU `make` is available in the FreeBSD ports tree as `gmake`.

This PR makes the changes necessary to specify `--use-gmake` for `catkin_make` and `catkin_make_isolated`.

See #1047

### DIFF
--- a/bin/catkin_make
+++ b/bin/catkin_make
@@ -178,8 +178,9 @@ def main():
         if args.use_ninja:
             cmake_args += ['-G', 'Ninja']
         elif args.use_nmake:
-            cmake_args += ['-G', 'NMake Makefiles']
+            cmake_args += ['-G', 'NMake Makefiles']            
         else:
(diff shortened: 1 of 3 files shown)


bin/catkin_make
```diff
@@ -178,8 +178,9 @@ def main():
         if args.use_ninja:
             cmake_args += ['-G', 'Ninja']
         elif args.use_nmake:
-            cmake_args += ['-G', 'NMake Makefiles']
+            cmake_args += ['-G', 'NMake Makefiles']
```
dirk-thomas: Please remove unrelated trailing whitespaces.

dodsonmg: Done.

dirk-thomas:
Thanks for the patch.

PR #1052, opened 2020-01-22
bump CMake minimum version to use new behavior of CMP0048

sloretz:
CMP0048 NEW behavior sets PROJECT_VERSION variables to empty strings if
project() is called with no VERSION argument. This could be a problem
for packages that set VERSION variables prior to project() because
project() would then clear them, but catkin does not do this.

**edit:** This bumps the minimum CMake version to 3.0.2, which is the [minimum supported by ROS Kinetic](https://www.ros.org/reps/rep-0003.html#kinetic-kame-may-2016-may-2021) and new enough to default to the NEW behavior of [CMP0048](https://cmake.org/cmake/help/v3.0/policy/CMP0048.html). This avoids a CMake warning when building and testing this package in Debian Buster and Ubuntu Focal.

Fixes this CMake warning: http://build.ros.org/job/Npr__catkin__ubuntu_focal_amd64/4/warnings2Result/new/

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,4 @@
-cmake_minimum_required(VERSION 2.8.3)
-cmake_policy(SET CMP0048 NEW)
+cmake_minimum_required(VERSION 3.0.2)
 
 set(catkin_EXTRAS_DIR ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,5 @@
 cmake_minimum_required(VERSION 2.8.3)
+cmake_policy(SET CMP0048 NEW)
 
 set(catkin_EXTRAS_DIR ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 


dirk-thomas:
This needs a rebase to pass CI.

dirk-thomas:
While this should work I am worried to require this change in every ROS package. Maybe we should explicitly set the policy (to enabled?)?

sloretz:
> While this should work I am worried to require this change in every ROS package. Maybe we should explicitly set the policy (to enabled?)?

I'm not sure how to do that automatically for downstream projects. IIUC it looks like the [warning is issued in the `project()` command itself](https://gitlab.kitware.com/cmake/cmake/blob/v3.13.4/Source/cmProjectCommand.cxx#L297-299). The policy would have to be set before then, but I think the earliest catkin could set something is in `catkinConfig.cmake` when `find_package(catkin)` is called. That is [expected to happen after calling `project()`](http://wiki.ros.org/catkin/CMakeLists.txt#Overall_Structure_and_Ordering).

dirk-thomas:
I bumped the CMake minimum version to something at least 3.0 (which is when CMP0048 was introduced) but not higher than 3.0.2 which is the CMake version on Debian Jessie targeted by ROS Kinetic.

sloretz:
> I bumped the CMake minimum version to something at least 3.0 (which is when CMP0048 was introduced) but not higher than 3.0.2 which is the CMake version on Debian Jessie targeted by ROS Kinetic.

:heavy_check_mark: works on my machine

dirk-thomas (approved):

sloretz:
**Edit:**
It looks like CMP0048 only is warned if `project()` would clear some variables. The reason the variables are set on the build farm is the packages are built in two steps: the first step invokes CMake to build tests, and the second step invokes CMake again to run them. In the second step, the following variables have been set in the CMake cache and would be overridden by the `project()` call.

```
CMake Warning (dev) at CMakeLists.txt:2 (project):
  Policy CMP0048 is not set: project() command manages VERSION variables.
  Run "cmake --help-policy CMP0048" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.

  The following variable(s) would be set to empty:

    CMAKE_PROJECT_VERSION
    CMAKE_PROJECT_VERSION_MAJOR
    CMAKE_PROJECT_VERSION_MINOR
    CMAKE_PROJECT_VERSION_PATCH
```

PR #1061, opened 2020-02-05
[Windows][noetic-devel] Generate executables without extension name.

seanyen:
For scripts ending with `.py`, `add_python_executable` will take the full filename (e.g., `abc.py`) and produce the executable ending with `.py.exe` (e.g., `abc.py.exe`) and which is not the expected script alias.

The pull request is proposing to extract the filename without the last extension by using [`NAME_WLE`](https://cmake.org/cmake/help/v3.14/command/get_filename_component.html) first and then generate the correct executable with the name without the extra `.py`.

And it would be great this can be backport to `kinetic` branch. Thanks!

### DIFF
--- a/cmake/platform/windows.cmake
+++ b/cmake/platform/windows.cmake
@@ -130,9 +130,12 @@ function(add_python_executable)
   endif()
 
   if(WIN32)
+    get_filename_component(WRAPPER_NAME
+      "${ARG_TARGET_NAME}"
+      NAME_WLE)
     set(
       WRAPPER_SOURCE
-      "${CMAKE_CURRENT_BINARY_DIR}/catkin_generated/add_python_executable/${ARG_TARGET_NAME}/${ARG_SCRIPT_NAME}.cpp")
+      "${CMAKE_CURRENT_BINARY_DIR}/catkin_generated/add_python_executable/${ARG_TARGET_NAME}/${WRAPPER_NAME}.cpp")
     configure_file(
       "${catkin_EXTRAS_DIR}/templates/python_win32_wrapper.cpp.in"
       "${WRAPPER_SOURCE}"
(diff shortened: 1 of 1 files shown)


seanyen:
@dirk-thomas This is ready for review and merge. Thanks!

cmake/platform/windows.cmake
```diff
@@ -130,20 +130,23 @@ function(add_python_executable)
   endif()
 
   if(WIN32)
+    get_filename_component(WRAPPER_NAME
+      "${ARG_TARGET_NAME}"
+      NAME_WLE)
```
dirk-thomas: This option is only available as of CMake 3.14. I suggest to use `NAME_WE` instead.

cmake/platform/windows.cmake
```diff
@@ -130,20 +130,23 @@ function(add_python_executable)
   endif()
 
   if(WIN32)
+    get_filename_component(WRAPPER_NAME
+      "${ARG_TARGET_NAME}"
```
dirk-thomas: The target name is different from the script name. And the latter is what needs to be used at least for the `OUTPUT_NAME` below.

dirk-thomas (changes requested):

cmake/platform/windows.cmake
seanyen: Done. I switched to use the `NAME_WE`.

cmake/platform/windows.cmake
seanyen: @dirk-thomas Sorry I don't follow here. Can you elaborate more what change I should make here?

dirk-thomas: The name of the installed script is passed to the function as `SCRIPT_NAME`. Changing it to use `TARGET_NAME` instead is wrong. E.g.:

`add_python_executable(SCRIPT_NAME "name_the_user_sees.py" TARGET_NAME cmake_target_name_a_user_does_not_care_about DESTINATION ...)`

seanyen: @dirk-thomas Ahh... I see what's the problem there. Now it is fixed.

dirk-thomas:
Thanks for the patch.

dirk-thomas (approved):

dirk-thomas:
Cherry-picked to `kinetic-devel`: 51b6b6a660730828043f8dbab99869af52780d97.

PR #1062, opened 2020-02-06
[Noetic] Global python in develspace

### LINKED ISSUE #1830: [py3] test_rostest test node run with wrong python version
## Problem description

It looks like `rostest` tests in this repo are using the wrong python version when `ROS_PYTHON_VERSION` is `3`. To reproduce, edit [`test_rostest` `just_advertise`](https://github.com/ros/ros_comm/blob/melodic-devel/test/test_rostest/test/just_advertise) to check the python version.

```diff
sloretz@nxt ~/w/n/s/ros_comm> git diff
diff --git a/test/test_rostest/test/just_advertise b/test/test_rostest/test/just_advertise
index 48507030f..6bda8abe5 100755
--- a/test/test_rostest/test/just_advertise
+++ b/test/test_rostest/test/just_advertise
@@ -44,6 +44,11 @@ import roslib.message
 
 
 if __name__ == '__main__':
+    import os
+    import sys
+    sys.stderr.write(repr(sys.version) + " " + repr(os.environ['ROS_PYTHON_VERSION']) + "\n")
+    assert sys.version[0] == os.environ['ROS_PYTHON_VERSION']
+
     rospy.init_node('just_advertise')
     msg_name = rospy.get_param('~msg_name')
     msg_class = roslib.message.get_message_class(msg_name)
```

In a `noetic` Python 3 workspace (I'm using Buster, but I'd expect the same issue on Bionic), run the tests using Python 3.

```
$ echo $ROS_PYTHON_VERSION
3
$ python3 src/catkin/bin/catkin_make_isolated --pkg test_rostest --cmake-args -DCATKIN_ENABLE_TESTING=1 --make-args run_tests
```

The test node is using python 2

```
'2.7.16 (default, Apr  6 2019, 01:42:57) \n[GCC 8.3.0]' '3'
  File "/home/sloretz/ws/noetic/src/ros_comm/test/test_rostest/test/just_advertise", line 50, in <module>
    assert sys.version[0] == os.environ['ROS_PYTHON_VERSION']
```

<details><summary>Full test output</summary>

```
==> Processing catkin package: 'test_rostest'
==> Building with env: '/home/sloretz/ws/noetic/devel_isolated/test_rosservice/env.sh'
Makefile exists, skipping explicit cmake invocation...
==> make cmake_check_build_system in '/home/sloretz/ws/noetic/build_isolated/test_rostest'
==> make run_tests -j8 -l8 in '/home/sloretz/ws/noetic/build_isolated/test_rostest'
Built target tests
Removing test result files from '/home/sloretz/ws/noetic/build_isolated/test_rostest/test_results/test_rostest'
- removing '/home/sloretz/ws/noetic/build_isolated/test_rostest/test_results/test_rostest/rostest-test_publishtest.xml'
- removing '/home/sloretz/ws/noetic/build_isolated/test_rostest/test_results/test_rostest/rosunit-publishtest.xml'
Built target clean_test_results_test_rostest
-- run_tests.py: execute commands
  /usr/bin/python3 /home/sloretz/ws/noetic/src/ros_comm/tools/rostest/scripts/rostest --pkgdir=/home/sloretz/ws/noetic/src/ros_comm/test/test_rostest --package=test_rostest --results-filename test_publishtest.xml --results-base-dir "/home/sloretz/ws/noetic/build_isolated/test_rostest/test_results" /home/sloretz/ws/noetic/src/ros_comm/test/test_rostest/test/publishtest.test 
... logging to /home/sloretz/.ros/log/rostest-43822c0b032c-32069.log
[ROSUNIT] Outputting test results to /home/sloretz/ws/noetic/build_isolated/test_rostest/test_results/test_rostest/rostest-test_publishtest.xml
'2.7.16 (default, Apr  6 2019, 01:42:57) \n[GCC 8.3.0]' '3'
  File "/home/sloretz/ws/noetic/src/ros_comm/test/test_rostest/test/just_advertise", line 50, in <module>
    assert sys.version[0] == os.environ['ROS_PYTHON_VERSION']
AssertionError
[Testcase: testpublishtest] ... ok

[ROSTEST]-----------------------------------------------------------------------

[test_rostest.rosunit-publishtest/test_publish][passed]

SUMMARY
 * RESULT: SUCCESS
 * TESTS: 1
 * ERRORS: 0
 * FAILURES: 0

rostest log file is in /home/sloretz/.ros/log/rostest-43822c0b032c-32069.log
-- run_tests.py: verify result "/home/sloretz/ws/noetic/build_isolated/test_rostest/test_results/test_rostest/rostest-test_publishtest.xml"
Built target _run_tests_test_rostest_rostest_test_publishtest.test
Built target _run_tests_test_rostest_rostest
Built target _run_tests_test_rostest
Built target run_tests
```
</details>

In this specific case it doesn't matter if the `just_advertise` node is using Python 2 since that's not what is being tested. However, this represents a larger problem with `rostest` and Python. `rostest` [requires the test node to be executable](https://github.com/ros/ros/blob/2b307e1854ebf3954a5b4110f891278f555e79a6/core/roslib/src/roslib/packages.py#L387-L404), and the [shebang in `src` space says `python`](https://github.com/ros/ros_comm/blob/6b1ee1a43b5b0a970d85612435e269818c89a152/test/test_rostest/test/just_advertise#L1). On Buster, `python` is Python 2.

### Require rostest users to add `launch-prefix="python$(env ROS_PYTHON_VERSION)"

One possible fix is to use `launch-prefix="python$(env ROS_PYTHON_VERSION)"` ([used by `tf2_geometry_msgs`](https://github.com/ros/geometry2/blob/dbc1254d45e71fb072dfb42abc2b9cf1e272ee76/tf2_geometry_msgs/test/test_python.launch#L2)).

```diff
diff --git a/test/test_rostest/test/publishtest.test b/test/test_rostest/test/publishtest.test
index b65e81b51..fbf129b8d 100644
--- a/test/test_rostest/test/publishtest.test
+++ b/test/test_rostest/test/publishtest.test
@@ -11,7 +11,7 @@
   </node>
 
   <node name="talker_2"
-        pkg="test_rostest" type="just_advertise">
+        pkg="test_rostest" type="just_advertise" launch-prefix="python$(env ROS_PYTHON_VERSION)">
     <rosparam>
       msg_name: std_msgs/String
     </rosparam>
```

However, since that was added to `geometry2`, it's been hypothesized this won't work on Windows because `python2` or `python3` commands won't exist. Currently on Windows there is a hard-coded check for files ending in `.py`, but that doesn't explain how the extension-less `just_advertise` runs. If `launch-prefix` does interfere with how Windows picks the python executable, then it will need to be understood to come up with a new workaround.

https://github.com/ros/ros_comm/blob/89ca9a7bf288b09d033a1ab6966ef8cfe39e1002/tools/roslaunch/src/roslaunch/node_args.py#L263-L266

### Provide `python` in `PATH` 

A possible fix is to prepend a path to an executable named `python`. On linux, any shebang using `#!/usr/bin/env python` would pick the `python` provided in the ROS workspace. This `python` executable would exec the python interpreter according to `ROS_PYTHON_VERSION`. How python is invoked on platforms that don't make use of the shebang, like Windows, would be unchanged. There are different places this approach could be applied.

#### A. Provide `python` in `PATH` when any ROS workspace is sourced

The program loader will pick the right version of python regardless of it being run through `roslaunch`, `rosrun`, or manually invoked. Any python executables created as child processes will use the correct version of python too.

It could interfere with someone expecting `python` to be different from `ROS_PYTHON_VERSION`. For example, some users put `. /opt/ros/.../setup.bash` into their `~/.bashrc`. If they do, then `python` would always refer to the version of python preferred by that `ROS` version instead of the system default.

#### B. Provide `python` in `PATH` when a ROS devel space is sourced

Whenever a devel space is sourced, the program loader will pick the right version of python regardless of it being run through `roslaunch`, `rosrun`. Any python executables created as child processes will use the correct version of python too.

This option is proposed to only apply to the devel space, not an install space. I'm assuming `devel` spaces are less likely to be automatically sourced, so the change to the `python` executable is less likely to affect non-ROS work. However, it means the python executable in the `devel` space may be different from the `python` executable in the install space. If users don't do `catkin_install_python()`, then the installed scripts may not work because the shebang has not been rewritten.

Implementing this option seems to be build tool specific. It looks like `catkin_make` and `catkin_make_isolated` create the devel space [here](https://github.com/ros/catkin/blob/dfc93c72d9799f57442d08c2f6e9fd494bba95c1/python/catkin/builder.py#L813), but catkin-tools does it [here](https://github.com/catkin/catkin_tools/blob/8f6add406d211d726d86b98a3c49cab7e1be7984/catkin_tools/verbs/catkin_build/build.py#L177).

#### C. `roslaunch` provides `python` in `PATH` when launching executables

Another option is to make `roslaunch` provide a `python` executable and prepend that to `PATH`. This limits the change to just scripts run through `roslaunch` (like `rostest` tests), but affects launch files in both the `devel` space and `install` space. Scripts invoked via `rosrun` or manually would not use this `python` executable.

### `roslaunch` reads shebang and calls right interpreter

This option would make `roslaunch` read the magic bytes on an executable, replacing what the linux program loader would normally do. If the first two bytes are `#!` then it would parse up to `\n` looking for shebangs that match expected shebang's for python. Then it would add `sys.executable` as a prefix to the launch command. It's unclear what to do with a manually specified `launch-prefix`. If the `launch-prefix` is gdb, then it would be desirable to put this python prefix between `launch-prefix` and the path to the python script. If the launch prefix is another python interpreter, then no extra prefixing should be done.

This would affect launch files in both the devel and install space, but would have an advantage that the existing method of executing python scripts on windows could be replaced with this approach. Scripts invoked via `rosrun` or manually would not use this `python` executable. A downside is python scripts installed in the install space lacking rewritten shebangs may work when used with roslaunch but not with rosrun.

sloretz:
Fixes ros/ros_comm#1830

This provides `python` in `$PATH` when the devel space is sourced by creating a symlink to `python` in `CATKIN_DEVEL_PREFIX/CATKIN_GLOBAL_BIN_DESTINATION`. I haven't seen unit tests run using python 2 with this fix.

Using `catkin_make` this creates a single symlink `<your-ws>/devel/bin/python`. With `catkin_make_isolated` this creates a symlink in `<your-ws>/devel_isolated/<package>/bin/python`, which I think means `$PATH` has more entries because every package now has a global bin destination. I have not yet tested with `catkin-tools`.

I've tested using `catkin_make_isolated` to build a Noetic workspace and run tests. I've also sourced that `devel_isolated` workspace and built an overlay using `catkin_make`. I have not yet tested with an overlay of an install space.

I moved some symlink creation code into a new function `_catkin_create_symlink` so it could be reused. The only change to it is `@CMAKE_COMMAND@` became `${CMAKE_COMMAND}` to avoid a [CMP0053 warning](https://cmake.org/cmake/help/v3.1/policy/CMP0053.html).

### DIFF
--- a/cmake/catkin_generate_environment.cmake
+++ b/cmake/catkin_generate_environment.cmake
@@ -57,6 +57,12 @@ function(catkin_generate_environment)
         @ONLY)
     endforeach()
 
+    # Not windows? There might be multiple python major versions installed
+    # Generate symlink to python so tests in devel space use correct version
+    get_filename_component(absolute_python_executable "${PYTHON_EXECUTABLE}" ABSOLUTE)
+    file(MAKE_DIRECTORY "${CATKIN_DEVEL_PREFIX}/${CATKIN_GLOBAL_BIN_DESTINATION}")
+    _catkin_create_symlink("${absolute_python_executable}" "${CATKIN_DEVEL_PREFIX}/${CATKIN_GLOBAL_BIN_DESTINATION}/python")
+
   else()
     # windows
     # generate env
--- a/cmake/catkin_symlink_install.cmake
+++ b/cmake/catkin_symlink_install.cmake
@@ -48,3 +48,32 @@ if(CATKIN_SYMLINK_INSTALL)
   # register script for being executed at install time
   install(SCRIPT "${CATKIN_SYMLINK_INSTALL_INSTALL_SCRIPT}")
 endif()
+
+function(_catkin_create_symlink absolute_file symlink)
+  # avoid any work if correct symlink is already in place
+  if(EXISTS "${symlink}" AND IS_SYMLINK "${symlink}")
+    get_filename_component(destination "${symlink}" REALPATH)
+    get_filename_component(real_absolute_file "${absolute_file}" REALPATH)
+    if(destination STREQUAL real_absolute_file)
+      message(STATUS "Up-to-date symlink: ${symlink}")
+      return()
+    endif()
+  endif()
+
+  message(STATUS "Symlinking: ${symlink}")
+  if(EXISTS "${symlink}" OR IS_SYMLINK "${symlink}")
+    file(REMOVE "${symlink}")
+  endif()
+
+  execute_process(
+    COMMAND "${CMAKE_COMMAND}" "-E" "create_symlink"
+      "${absolute_file}"
+      "${symlink}"
+  )
+  # the CMake command does not provide a return code so check manually
+  if(NOT EXISTS "${symlink}" OR NOT IS_SYMLINK "${symlink}")
+    get_filename_component(destination "${symlink}" REALPATH)
+    message(FATAL_ERROR
+      "Could not create symlink '${symlink}' pointing to '${absolute_file}'")
+  endif()
+endfunction()
--- a/cmake/symlink_install/catkin_symlink_install.cmake.in
+++ b/cmake/symlink_install/catkin_symlink_install.cmake.in
@@ -266,33 +266,8 @@ function(_catkin_symlink_install_create_symlink absolute_file symlink)
   file(APPEND "${CMAKE_CURRENT_BINARY_DIR}/symlink_install_manifest.txt"
     "${symlink}\n")
 
-  # avoid any work if correct symlink is already in place
-  if(EXISTS "${symlink}" AND IS_SYMLINK "${symlink}")
-    get_filename_component(destination "${symlink}" REALPATH)
-    get_filename_component(real_absolute_file "${absolute_file}" REALPATH)
-    if(destination STREQUAL real_absolute_file)
-      message(STATUS "Up-to-date symlink: ${symlink}")
-      return()
-    endif()
-  endif()
-
-  message(STATUS "Symlinking: ${symlink}")
-  if(EXISTS "${symlink}" OR IS_SYMLINK "${symlink}")
-    file(REMOVE "${symlink}")
-  endif()
-
-  execute_process(
-    COMMAND "@CMAKE_COMMAND@" "-E" "create_symlink"
-      "${absolute_file}"
-      "${symlink}"
-  )
-  # the CMake command does not provide a return code so check manually
-  if(NOT EXISTS "${symlink}" OR NOT IS_SYMLINK "${symlink}")
-    get_filename_component(destination "${symlink}" REALPATH)
-    message(FATAL_ERROR
-      "Could not create symlink '${symlink}' pointing to '${absolute_file}'")
-  endif()
-endfunction()
+  _catkin_create_symlink("${absolute_file}" "${symlink}")
+endfunction
 
 # end of template
 


dirk-thomas:
See https://github.com/ros/ros_comm/issues/1830#issuecomment-583167335

sloretz:
Closing in favor of https://github.com/ros/ros_comm/pull/1879

PR #1063, opened 2020-02-07
[Windows][noetic] a follow up fix to the generated executable change.

seanyen:
After more validation to my earlier changes of ["Generate executables without extension name"](https://github.com/ros/catkin/pull/1061), I overlooked a required change in `python_win32_wrapper.cpp.in` side. Please excuse me causing any inconvenience.

### DIFF
--- a/cmake/templates/python_win32_wrapper.cpp.in
+++ b/cmake/templates/python_win32_wrapper.cpp.in
@@ -70,22 +70,35 @@ int main(int argc, char* argv[]) try
     const auto FindPythonScript = [](const std::string& exeName) -> std::string
     {
         // implement a heuristic here to execute a Python script of the same name
-        // when the executable's name is <name>.exe, assume the Python script to be <name>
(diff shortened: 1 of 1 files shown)


seanyen:
@dirk-thomas This is ready for review and merge. Thanks!

dirk-thomas:
Why should the code need to guess? The script name is known at configure time - if it does or does not contain a `.py` extension. How about embedding that knowledge into the `.cpp`file?

seanyen:
@dirk-thomas Sure, I changed the template `.cpp` to refer to `@ARG_SCRIPT_NAME@` for the script name.

dirk-thomas:
Thanks for the patch.

dirk-thomas:
Also cherry-picked to `kinetic-devel`: ebaa3c5ee91363ea066d8f3e9970051440871ab9.

seanyen:
@dirk-thomas Thanks!

PR #1065, opened 2020-02-10
catkin_make: allow flexible CMake minimum version in metapackage CMake code

### LINKED ISSUE #1064: catkin_make expects cmake_minimum_required 2.8.3 for metapackages
It seems that `catkin_make` always expects `cmake_minimum_required(VERSION 2.8.3)` for metapackages.

It checks that the metapackage CMakeLists.txt conforms exactly to the expected format [here](https://github.com/ros/catkin/blob/noetic-devel/cmake/catkin_workspace.cmake#L56-L72) and the build fails if it doesn't. The format has a fixed cmake version that seems to come from [here](https://github.com/ros-infrastructure/catkin_pkg/blob/master/src/catkin_pkg/templates/metapackage.cmake.in).

What is the rationale for imposing a fixed version? If someone wants to force a higher minimum version for certain packages then they won't be able to do the same for the metapackage containing those packages.

I see that recently the minimum version was bumped for a number of packages and metapackages. (https://github.com/ros/catkin/pull/1052, https://github.com/ros/ros_comm/pull/1869 and other PRs). All these will probably build with with `catkin_make_isolated`, but I think they will no longer build with `catkin_make`.

dirk-thomas:
Fixes #1064. Needs to be backported to `kinetic-devel` branch.

### DIFF
--- a/cmake/catkin_workspace.cmake
+++ b/cmake/catkin_workspace.cmake
@@ -64,8 +64,13 @@ function(catkin_workspace)
         if(EXISTS ${CMAKE_SOURCE_DIR}/${path}/CMakeLists.txt)
           # compare CMakeLists.txt with standard content
           file(STRINGS ${CMAKE_CURRENT_BINARY_DIR}/catkin_generated/metapackages/${name}/CMakeLists.txt generated_cmakelists)
+          # allow any of CMake minimum version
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Backported to `kinetic-devel`: 38378ed907e73c43aa7d0544c8bda3aee37750ca.

PR #1071, opened 2020-03-19
[Windows] Accommodate different drives for --root

seanyen:
This is a follow-up fix for #1070 

In cases of that `CMAKE_INSTALL_PREFIX` resides in the different drive from the source code, the previous fix will result in the `Python` module to be installed at the same drive where the source code sits. For example, if ROS source code is checked out on `C:` drive and `cmake` install into `D:\opt\instsall`, the `Python` module will be mistakenly installed to `c:\opt\install\`.

This pull request is to take the target drive into consideration. And hope this can be cherry-picked into `kinectic-devel`.


seanyen:
@dirk-thomas This is ready for review and merge. Thanks for your feedback on the drive name thing. Problem do manifest when the source workspace is on different drives from the install space.

dirk-thomas:
Thanks for the patch.

That was exactly what I was concerned about in the previous patch: https://github.com/ros/catkin/pull/1070#discussion_r393912628

dirk-thomas:
Cherry-picked to `kinetic-devel`: 39bbd52cb69c567f915a372906216e6c9ac8b226.

Issue #1069, opened 2020-03-13
Using setuptools generates egg-info folders in source space when installing


matthew-reynolds:
Continuing the conversation from https://github.com/ros/catkin/pull/1048#discussion_r392361424. Using `setuptools`, as is now recommended:

`egg-info` folders are created in the source space when building Melodic packages w/ python 2 on Bionic. This seems to only happen when installing (ie. `catkin_make install`). (See https://travis-ci.org/github/ros-controls/ros_control/jobs/661982516?utm_medium=notification&utm_source=github_status)

Steps to reproduce:

```
cd /tmp
mkdir -p mel_ws/src
cd mel_ws/src
git clone git@github.com:ros/angles.git
chmod a-w -R angles
cd /tmp/mel_ws
catkin config --install; catkin build         # Fails
catkin clean
catkin_make install                           # Fails
catkin_make_isolated --install                # Fails
```

matthew-reynolds:
FYI @sloretz @mikepurvis

mikepurvis:
Oh, it's a Py2 issue—that might explain why I'm not hitting it as we're now migrated.

dirk-thomas:
What `setuptools` version are you using (`pip freeze --all | grep setuptools`)?

sloretz:
> What setuptools version are you using (pip freeze --all | grep setuptools)?

I'm able to reproduce with `setuptools==39.0.1`

matthew-reynolds:
> What setuptools version are you using (pip freeze --all | grep setuptools)?

I'm using `setuptools==41.0.1`. Not sure what version the Travis build I linked used though.

Edit: I can still reproduce the issue with `setuptools=44.0.0`, which AFAIK is the latest version that is Py2.7 compatible.

dirk-thomas:
Addressed in #1073.

PR #1073, opened 2020-04-02
Set egg-base for setuptools packages

matthew-reynolds:
Addresses #1069.

Based on @sloretz's [suggestion](https://github.com/ros/catkin/pull/1048#discussion_r392337794) of setting `--egg-base`, with some code loosely adapted from @mikepurvis's [suggestion](https://github.com/ros/catkin/pull/1048#discussion_r392345260).

Tested on Melodic with each tool on both a distutils and a setuptools package:
```
catkin_make
catkin_make install
catkin_make_isolated
catkin_make_isolated --install
catkin config --no-install; catkin build
catkin config --install; catkin build
colcon build
colcon build --merge-install
```

I haven't had a chance to test this with Python 3, would appreciate help there.

### DIFF
--- a/cmake/catkin_python_setup.cmake
+++ b/cmake/catkin_python_setup.cmake
@@ -31,29 +31,6 @@ function(catkin_python_setup)
     message(FATAL_ERROR "catkin_python_setup() called without 'setup.py' in project folder ' ${${PROJECT_NAME}_SOURCE_DIR}'")
   endif()
 
-  assert(PYTHON_INSTALL_DIR)
-  set(INSTALL_CMD_WORKING_DIRECTORY ${${PROJECT_NAME}_SOURCE_DIR})
-  if(NOT WIN32)
-    set(INSTALL_SCRIPT
-      ${CMAKE_CURRENT_BINARY_DIR}/catkin_generated/python_distutils_install.sh)
-    configure_file(${catkin_EXTRAS_DIR}/templates/python_distutils_install.sh.in
-      ${INSTALL_SCRIPT}
-      @ONLY)
-  else()
-    # need to convert install prefix to native path for python setuptools --prefix (its fussy about \'s)
-    file(TO_NATIVE_PATH ${CMAKE_INSTALL_PREFIX} PYTHON_INSTALL_PREFIX)
-    set(INSTALL_SCRIPT
-      ${CMAKE_CURRENT_BINARY_DIR}/catkin_generated/python_distutils_install.bat)
-    configure_file(${catkin_EXTRAS_DIR}/templates/python_distutils_install.bat.in
-      ${INSTALL_SCRIPT}
-      @ONLY)
-  endif()
-
-  # generate python script which gets executed at install time
-  configure_file(${catkin_EXTRAS_DIR}/templates/safe_execute_install.cmake.in
-    ${CMAKE_CURRENT_BINARY_DIR}/catkin_generated/safe_execute_install.cmake)
-  install(SCRIPT ${CMAKE_CURRENT_BINARY_DIR}/catkin_generated/safe_execute_install.cmake)
-
   # interrogate setup.py
   stamp(${${PROJECT_NAME}_SOURCE_DIR}/setup.py)
   assert(CATKIN_ENV)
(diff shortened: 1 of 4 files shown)


mikepurvis:
Thanks for updating the tests as well. :)

Overall this looks reasonable, though we definitely need to understand if this fix is needed for Python 3. If it's a 2 only issue, then it may make sense to target the Melodic branch directly with the change (or simply accept it as a feature, given how potentially disruptive the fix is).

If you're on Bionic/Melodic, it's pretty easy to test Py3 locally— just make a small workspace in a terminal that hasn't sourced anything else ROS, and do one of either:

```
catkin_make_isolated -DPYTHON_EXECUTABLE=python3
catkin config --cmake-args -DPYTHON_EXECUTABLE=python3
```

matthew-reynolds:
> We definitely need to understand if this fix is needed for Python 3

Agreed. I created a little Py3 Melodic workspace to test and found that this issue seems to exist in Py3 as well.

```
mkdir -p ~/ros_py3/src
cd ~/ros_py3
rosinstall_generator angles --rosdistro melodic --deps --upstream-devel > dependencies.rosinstall
vcs import --input dependencies.rosinstall ./src
export ROS_PYTHON_VERSION=3
python3 ./src/catkin/bin/catkin_make_isolated --install -DCMAKE_BUILD_TYPE=Release --catkin-make-args all

find ./src -name *egg-info*
```

When using the `kinetic-devel` branch, I end up with a whole bunch of `*.egg-info` folders in the source space, but with this patch my source space remains untouched.

matthew-reynolds:
Whoops, forgot the last bit of my previous comment.

@mikepurvis You mentioned in https://github.com/ros/catkin/pull/1048#discussion_r392325758 that you were never seeing these egg-info folders appear in the source space using Python 3 - Do you see any differences between our setups that might be causing our different results?

cmake/catkin_python_setup.cmake
```diff
@@ -163,6 +140,34 @@ function(catkin_python_setup)
       TARGET_NAME ${PROJECT_NAME}_${name}_exec_install
       DESTINATION ${CATKIN_GLOBAL_BIN_DESTINATION})
   endforeach()
+
+  if(${PROJECT_NAME}_SETUP_PY_SETUP_MODULE STREQUAL setuptools)
```
dirk-thomas: Please put quotes around the string `setuptools` to avoid it being interpreted as a variable.

cmake/catkin_python_setup.cmake
```diff
@@ -163,6 +140,34 @@ function(catkin_python_setup)
...
+    message(STATUS "Appending egg base")
```
dirk-thomas: While this was useful for debugging I would suggest to remove it from the patch.

cmake/catkin_python_setup.cmake
```diff
@@ -163,6 +140,34 @@ function(catkin_python_setup)
...
+    set(SETUPTOOLS_EGG_INFO "egg_info --egg-base ${CMAKE_INSTALL_PREFIX}/${PYTHON_INSTALL_DIR}")
+  endif()
+
+  assert(PYTHON_INSTALL_DIR)
```
dirk-thomas: Since line 146 already uses the variable this `assert` must be moved above the previous `if`.

cmake/interrogate_setup_dot_py.py
```diff
@@ -92,7 +92,7 @@ def _get_locations(pkgs, package_dir):
     return locations
 
 
-def generate_cmake_file(package_name, version, scripts, package_dir, pkgs, modules):
+def generate_cmake_file(setup_module, package_name, version, scripts, package_dir, pkgs, modules):
```
dirk-thomas: To avoid breaking the API please add the new argument at the end and make it optional (and only add the variable below if a value was passed.

cmake/catkin_python_setup.cmake
```diff
@@ -163,6 +140,34 @@ function(catkin_python_setup)
       TARGET_NAME ${PROJECT_NAME}_${name}_exec_install
       DESTINATION ${CATKIN_GLOBAL_BIN_DESTINATION})
   endforeach()
+
+  if(${PROJECT_NAME}_SETUP_PY_SETUP_MODULE STREQUAL setuptools)
+    message(STATUS "Appending egg base")
+    set(SETUPTOOLS_EGG_INFO "egg_info --egg-base ${CMAKE_INSTALL_PREFIX}/${PYTHON_INSTALL_DIR}")
```
dirk-thomas, line 146: Please set the variable to an empty string in the `else` case to ensure it is initialized before it is used.

cmake/interrogate_setup_dot_py.py
matthew-reynolds: Resolved in 00847412

cmake/catkin_python_setup.cmake
matthew-reynolds: Resolved in 0c9d05d4

cmake/catkin_python_setup.cmake
matthew-reynolds: Resolved in 0c9d05d4

cmake/catkin_python_setup.cmake
matthew-reynolds: Resolved in 0c9d05d4

cmake/catkin_python_setup.cmake
matthew-reynolds: Resolved in 0c9d05d4

dirk-thomas:
Thanks for the patch and the updates. This works for me in Noetic as desired.

mikepurvis (approved):

matthew-reynolds:
Thanks @dirk-thomas. Hopefully this can get backported to `kinetic-devel` as well, I saw you tagged it as such :+1:. This issue currently blocking us from switching to setuptools in Melodic (in preparation of branching for Noetic). 

Is there anything on my end I can do to help? Is it useful to open a PR with the commit against `kinetic-devel`, or is it easier to just leave it for the maintainers to decide whether to cherry pick it back?

dirk-thomas:
@sloretz Is about to do some more tests with Melodic (and eventually Kinetic) to check that it is safe to backport this change. I will do the cherry-pick when we are confident.

matthew-reynolds:
Sounds great - thanks!

dirk-thomas:
This caused a regression: #1087.

PR #1075, opened 2020-04-06
stamp env hook before copying

### LINKED ISSUE #1072: catkin_add_env_hooks does not copy updated files
`catkin_add_env_hooks` uses  `file(COPY ...)` macro to copy env hook files (https://github.com/ros/catkin/blob/noetic-devel/cmake/catkin_add_env_hooks.cmake#L97).
`file` macro copies files at generation time and it does not copy files at build time.

dirk-thomas:
Fixes #1072.


dirk-thomas:
Cherry-picked to `kinetic-devel` in bea5d4fafb4324937cc9b5c7f1bb6cf7c02c7359.

Issue #1080, opened 2020-04-14
Catkin cannot handle when the CMake language is set to none


BryceStevenWilley:
Setting the CMake Language to none (which makes sense if this is a catkin package that is solely providing resources, meshes, configurations, etc.) causes CMake to fail. 

The smallest CMakeLists.txt that will cause the issue:
```
cmake_minimum_required(VERSION 3.0.2)
project(hello_test LANGUAGES NONE)

find_package(catkin REQUIRED)
catkin_package()
```
This will result in two different errors, one regarding FindThreads and GTest:
```
CMake Error at /usr/share/cmake-3.10/Modules/FindThreads.cmake:54 (message):
  FindThreads only works if either C or CXX language is enabled
Call Stack (most recent call first):
  /usr/share/cmake-3.10/Modules/FindGTest.cmake:204 (find_package)
  /opt/ros/melodic/share/catkin/cmake/test/gtest.cmake:304 (find_package)
  /opt/ros/melodic/share/catkin/cmake/all.cmake:164 (include)
  /opt/ros/melodic/share/catkin/cmake/catkinConfig.cmake:20 (include)
  CMakeLists.txt:6 (find_package)
```
and another regarding RT (if built with `CATKIN_ENABLE_TESTING=False`):
```
CMake Error at /opt/ros/melodic/share/catkin/cmake/assert.cmake:17 (message):


  Assertion failed: check for file existence, but filename
  (RT_LIBRARY-NOTFOUND) unset.  Message: RT Library

Call Stack (most recent call first):
  /opt/ros/melodic/share/catkin/cmake/tools/rt.cmake:42 (assert_file_exists)
  /opt/ros/melodic/share/catkin/cmake/all.cmake:164 (include)
  /opt/ros/melodic/share/catkin/cmake/catkinConfig.cmake:20 (include)
  CMakeLists.txt:4 (find_package)
```

The main issue is that catkin tries to find a lot of C++ libraries when it doesn't need to (and when CMake can't). 

The easy fix is to just make the project's languages `CXX`, but to "proper" (and probably too much effort) fix is to not include all of the C++ specific libraries that are necessary.

dirk-thomas:
It should be possible to wrap the specific logic which assumes that either C or C++ is an enabled language in a conditional checking the global CMake variable `ENABLED_LANGUAGES`. A pull request to support this would be appreciated.

sloretz:
[...]

PR #1083, opened 2020-04-30
Stops catkin from trying to find C++ libraries if not needed

BryceStevenWilley:
Specifically, when CXX is not in the `ENABLED_LANGUAGES` global
property.

Addresses https://github.com/ros/catkin/issues/1080.

### DIFF
--- a/cmake/test/gtest.cmake
+++ b/cmake/test/gtest.cmake
@@ -1,6 +1,9 @@
 _generate_function_if_testing_is_disabled(
   "catkin_add_gtest"
   "catkin_add_gmock")
+_generate_function_if_no_cxx_language(
+  "catkin_add_gtest"
+  "catkin_add_gmock")
 
 #
 # Add a GTest based test target.
@@ -299,6 +302,12 @@ function(catkin_find_google_test_source gtest_path googletest_path
     endif()
   endif()
 endfunction()
+
+get_property(ENABLED_LANGUAGES_LIST GLOBAL PROPERTY ENABLED_LANGUAGES)
+list(FIND ENABLED_LANGUAGES_LIST "CXX" _cxx_index)
+if ("${_cxx_index}" EQUAL -1) # if C++ isn't an enabled language
+  return()
+endif()
 
 find_package(GMock QUIET)
 find_package(GTest QUIET)
--- a/cmake/test/gtest.cmake
+++ b/cmake/test/gtest.cmake
@@ -1,9 +1,15 @@
 _generate_function_if_testing_is_disabled(
   "catkin_add_gtest"
-  "catkin_add_gmock")
+  "catkin_add_gmock"
+  "catkin_add_executable_with_gtest"
+  "catkin_add_executable_with_gmock"
+  "catkin_find_google_test_source")
 _generate_function_if_no_cxx_language(
   "catkin_add_gtest"
-  "catkin_add_gmock")
+  "catkin_add_gmock"
+  "catkin_add_executable_with_gtest"
+  "catkin_add_executable_with_gmock"
+  "catkin_find_google_test_source")
 
 #
 # Add a GTest based test target.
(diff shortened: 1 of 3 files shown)


BryceStevenWilley:
My only concern is that Windows still might require some CXX libraries due to: https://github.com/ros/catkin/blob/712da602172d9ab0efc3b3108def4c43c0198db8/cmake/platform/windows.cmake#L100. But I'm not familiar with anything Windows, or why that `enable_language` call is there.

dirk-thomas:
The current patch skips including some CMake files so a package trying to call any of the expected API will get a rather opaque error message that the function doesn't exist. I don't think that is a good behavior for users.

Instead the functions should continue to be defined but print a helpful error message why they can't be used.

BryceStevenWilley:
I found the `_generated_function_if_testing_is_disabled` macro, and copied off of that same template if C++ isn't set as a language. Let me know if there are any more changes you would like

cmake/test/gtest.cmake
```diff
@@ -1,6 +1,9 @@
 _generate_function_if_testing_is_disabled(
   "catkin_add_gtest"
   "catkin_add_gmock")
+_generate_function_if_no_cxx_language(
+  "catkin_add_gtest"
+  "catkin_add_gmock")
```
dirk-thomas: This needs to include other public functions from this file, e.g. `catkin_add_executable_with_gtest`.

The same is probably needed above in `_generate_function_if_testing_is_disabled`.

cmake/test/gtest.cmake
```diff
@@ -300,6 +303,12 @@ function(catkin_find_google_test_source gtest_path googletest_path
   endif()
 endfunction()
 
+get_property(ENABLED_LANGUAGES_LIST GLOBAL PROPERTY ENABLED_LANGUAGES)
+list(FIND ENABLED_LANGUAGES_LIST "CXX" _cxx_index)
+if ("${_cxx_index}" EQUAL -1) # if C++ isn't an enabled language
+  return()
+endif()
+
```
dirk-thomas: This logic can never be triggered. The macro `_generate_function_if_no_cxx_language` already contains a `return()` statement which ends the file when it is invoked in line 4.

cmake/test/tests.cmake
```diff
@@ -26,6 +26,23 @@ macro(_generate_function_if_testing_is_disabled)
...
+macro(_generate_function_if_no_cxx_language)
+  get_property(ENABLED_LANGUAGES_LIST GLOBAL PROPERTY ENABLED_LANGUAGES)
+  list(FIND ENABLED_LANGUAGES_LIST "CXX" _cxx_index)
+  if ("${_cxx_index}" EQUAL -1) # if C++ isn't an enabled language
+    foreach(_arg ${ARGN})
+      set(tmp_func_name ${_arg})
+      function(${_arg})
+        message(FATAL_ERROR
+          "${tmp_func_name}() is not available when the CXX language isn't enabled. "
+          "Try adding \"project(project_name LANGUAGES CXX)\" to your cmake.")
```
dirk-thomas: I suggest rephrasing this sentence since it currently makes assumption what the package might do / might want to do, maybe:

> Check the 'LANGUAGES' argument of the 'project()' call to make sure CXX is not excluded.

cmake/test/gtest.cmake
BryceStevenWilley: Think I forgot to remove this from the first attempt, before it was using `_generate_function...`. Thanks

dirk-thomas:
Thanks for the patch and for iterating on it.

dirk-thomas:
Cherry-picked to `kinetic-devel` in 9bd4b21322b21b086f3fa17ab6f52b45c7491e1a.

ros-discourse:
This pull request has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/preparing-for-melodic-sync-2020-06-09/14567/5

PR #1091, opened 2020-05-13
fix GTest detection when cmake-extras is installed

### LINKED ISSUE #1084: cmake-extras package breaks google test in
Ok, this one is oddly-specific, but here we go.
When the cmake-extras package is installed, google test builds will fail with "gtest not found", even though gtest is, in fact, found.  The mechanism is as follows.

The cmake-extras package installs a Find file for GMock here: /usr/share/cmake/GMock/GMockConfig.cmake
This GMockConfig file will add a target for gtest in the find_package(GMock) call on line 303 of gtest.cmake, but the call to find_package(GTest...) will still return "not found".  However, the target already exists, so FORCE_GTEST_GMOCK_FROM_SOURCE never gets set and I end up in this weird state where gtest is reported as not found despite being added to the project.

Not really sure what the best way to fix this is-- in practice, I just simply removed the offending package.

dirk-thomas:
Fixes #1084.

### DIFF
--- a/cmake/test/gtest.cmake
+++ b/cmake/test/gtest.cmake
@@ -301,7 +301,13 @@ function(catkin_find_google_test_source gtest_path googletest_path
 endfunction()
 
 find_package(GMock QUIET)
-find_package(GTest QUIET)
+# the GMockConfig.cmake from the Debian package cmake-extras provides all targets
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Cherry-picked to `kinetic-devel`: 1daa680b32d333dbea7eebf78be57406b8072147.

PR #1106, opened 2020-07-14
explicitly call project() in toplevel CMakeLists.txt

dirk-thomas:
To prevent a warning to be shown with CMake 3.15 and newer:

```
CMake Warning (dev) in CMakeLists.txt:
  No project() command is present.  The top-level CMakeLists.txt file must
  contain a literal, direct call to the project() command.  Add a line of
  code such as

    project(ProjectName)

  near the top of the file, but after cmake_minimum_required().

  CMake is pretending there is a "project(Project)" command on the first
  line.
```

### DIFF
--- a/cmake/toplevel.cmake
+++ b/cmake/toplevel.cmake
@@ -3,6 +3,8 @@
 
 cmake_minimum_required(VERSION 3.0.2)
 
+project(Project)
+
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Cherry-picked to `kinetic-devel` in c964027ac031d31a3b87d51b692e4023d2e1e65b.

PR #1100, opened 2020-06-29
Fix symlink install python when shebang line is rewritten

kenji-miyake:
## Overview

This PR fixes `catkin_install_python` when `CATKIN_SYMLINK_INSTALL=ON`.

~I'm sending this PR to `kinetic-devel` because I'm still using `melodic`, but I believe the same change can be applied to `noetic-devel`.~ -> Changed to `noetic-devel`.

## Problem

In `catkin_install_python.cmake`, new (shebang-replaced) files which are generated by `file(WRITE ...)`, which drops the permissions.
https://github.com/ros/catkin/blob/4fe89dd8a553a74e16e14754021e6550386fe5a3/cmake/catkin_install_python.cmake#L42

This disables python-scripts execution with `rosrun`.

### Supplemental

The permissions of symlink files (inside `install/`) are executable because of `install(PROGRAMS ...)`
https://github.com/ros/catkin/blob/4fe89dd8a553a74e16e14754021e6550386fe5a3/cmake/catkin_install_python.cmake#L53

However, the real files (inside `catkin_generated/installspace`) are not executable.

## Solution

I just added `file(COPY ...)` before `file(WRITE ...)`.
https://github.com/ros/catkin/commit/aead61da62d50404e0c4539173484469de49934b#diff-3cfcccb0a4b1968a8dae7de378653f99R41

I also added a variable `installspace` in order to avoid overwriting the variable, but I can drop this change from this PR.
https://github.com/ros/catkin/commit/8f23bf5300a386f68888c0a3357ded482779ede8

## How to test (Melodic)

```sh-session
###################################
##### Launch docker container #####
###################################
$ docker run --rm -it osrf/ros:melodic-desktop-full

##################################################
##### Check result with this patch (working) #####
##################################################
# apt update && apt install -y python-catkin-tools
# mkdir -p ~/catkin_ws/src
# cd ~/catkin_ws/src
# git clone https://github.com/kenji-miyake/catkin.git --depth 1 -b fix-symlink-install-python-kinetic # Actually based on noetic-devel
# git clone https://github.com/ros/ros_tutorials.git --depth 1 -b melodic-devel
# cd ~/catkin_ws
# catkin config --install
# catkin build rospy_tutorials --cmake-args -DCATKIN_SYMLINK_INSTALL=ON

# ls -l ~/catkin_ws/install/share/rospy_tutorials/001_talker_listener
total 24
[...]

PR #1109, opened 2020-07-22
[noetic-devel][Windows] Fix file COPY cannot read symlink

seanyen:
A recent change from #1100 introduced an usage of `file(COPY)` and unfortunately this idiom has some limitation on Windows. For example, `CMake` doesn't support the case where the source file is a symlink, due to the absent of [`ReadSymlink`](https://gitlab.kitware.com/cmake/cmake/-/issues/20994#note_802160) implementation on Windows.

And when the conditions are met, it will render as errors like below:

```
-- rospy_tutorials: 2 messages, 2 services
CMake Error at C:/opt/ros/melodic/x64/share/catkin/cmake/catkin_install_python.cmake:45 (file):
  file COPY cannot read symlink
  "C:/workspace/rospy_ws/src/ros_tutorials/rospy_tutorials/001_talker_listener/listener"
  to duplicate at
  "C:/workspace/rospy_ws/build_isolated/rospy_tutorials/catkin_generated/installspace/listener":
  No error.
  CMakeLists.txt:14 (catkin_install_python)
```

This pull request is to make the file copying conditional and unblock any build breaks due to the limitation. And please consider to back-port to `kinetic-devel` as well.

### DIFF
--- a/cmake/catkin_install_python.cmake
+++ b/cmake/catkin_install_python.cmake
@@ -40,7 +40,10 @@ function(catkin_install_python signature)
         file(MAKE_DIRECTORY ${rewritten_file})
         # even though the content of the file is overwritten
         # the copy makes sure the file has the right permissions
-        file(COPY ${source_file} DESTINATION ${rewritten_file} USE_SOURCE_PERMISSIONS)
+        if(NOT IS_SYMLINK ${source_file} AND WIN32)
+          # file copying is not supported on Windows symlinks by CMake
+          file(COPY ${source_file} DESTINATION ${rewritten_file} USE_SOURCE_PERMISSIONS)
+        endif()
         set(rewritten_file "${rewritten_file}/${filename}")
         file(WRITE ${rewritten_file} "${data}")
       else()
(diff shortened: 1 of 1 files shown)


cmake/catkin_install_python.cmake
```diff
@@ -40,7 +40,10 @@ function(catkin_install_python signature)
         file(MAKE_DIRECTORY ${rewritten_file})
         # even though the content of the file is overwritten
         # the copy makes sure the file has the right permissions
-        file(COPY ${source_file} DESTINATION ${rewritten_file} USE_SOURCE_PERMISSIONS)
+        if(NOT IS_SYMLINK ${source_file} AND WIN32)
```
dirk-thomas: The condition looks incorrect. The file copy wouldn't be executed on non-Windows anymore?

dirk-thomas: @seanyen Please double check the modified condition: 5895a4b6b6f994d34adaf90477a53522e51c955d.

seanyen: Thanks for the patch! I confirmed the modified version fixes the bug as well.

dirk-thomas:
Thanks for the patch and the quick response.

dirk-thomas:
Cherry-picked to `kinetic-devel` in 232339eed0f04d6325443747faefabd4aa059289.

PR #1110, opened 2020-07-23
[noetic][Windows] Adding .lib into the symlink install file list

seanyen:
This [enhancement](https://github.com/colcon/colcon-ros/pull/109) makes an existing bug in `CATKIN_SYMLINK_INSTALL` surface up. 😉

The problem & solution is the same as https://github.com/ament/ament_cmake/pull/219.

This is a cherry-pick and please consider to back-port it into `kinetic-devel` as well.


dirk-thomas:
Thanks for the patch.

dirk-thomas:
Cherry-picked to `kinetic-devel` in 807c983e2c3191866b40c7e2f79c66d6f0ff8c8e.

PR #1114, opened 2020-08-07
[noetic] Adding GTest v1.8.1-later search logic

seanyen:
This is is motivated by building the `noetic` with the [`Vcpkg`](https://github.com/microsoft/vcpkg) GTest port. So, by the stock [`FindGTest`](https://cmake.org/cmake/help/v3.18/module/FindGTest.html), it wasn't able to find the GTest installed by `Vcpkg`. And this patch is attempting to search for the namespaced CMake targets introduced since `GTest v1.8.1` and convert that into the CMake variables that `gtest.cmake` expects.


cmake/test/gtest.cmake
```diff
@@ -442,6 +442,31 @@ else()
...
+    add_library(gmock ALIAS GTest::gmock)
+    add_library(gmock_main ALIAS GTest::gmock_main)
+    set(GMOCK_INCLUDE_DIRS)
+    set(GMOCK_LIBRARY_DIRS)
+    set(GMOCK_LIBRARIES gmock)
+    set(GMOCK_MAIN_LIBRARIES gmock_main)
+    set(GMOCK_BOTH_LIBRARIES ${GMOCK_LIBRARIES} ${GMOCK_MAIN_LIBRARIES})
+  endif()
+endif()
+
```
dirk-thomas, line 476: The added logic looks completely "standalone". I don't see why this should be necessary. The CMake code before already calls `find_package(GMock QUIET)` / `find_package(GTest QUIET)`. Why it gtest not found by these?

Anyway in case there needs to be custom logic for Windows it should be integrated into the existing logic and it should be obvious why the existing logic isn't sufficient.

seanyen: Thanks. I added some comments on why explicitly check `GTest` config files would help in some cases. Hope that justifies the reason why it is nice to have it.

dirk-thomas: I still don't see why this block should be standalone / an afterthought below all the existing logic.

dirk-thomas: @seanyen Friendly ping.

seanyen: Thanks for the ping. I didn't have a chance to refactor this logic for now and this is workaround by placing the GTest libs locations like other platforms. I am closing it for now (maybe revisit again later) to reduce the noise.

PR #1116, opened 2020-08-16
[Windows] execute the output of _setup_util.py in place

seanyen:
When different isolated `setup.bat` are invoked at the same time (for example, when building in parallel with `colcon`), the `_SETUP_TMP` has chances to end up with the same file path (the file name is generated by `%Time: =0%` and it doesn't give a good quality for uniqueness). And in this case, the file write (creation) and deletion can race together and result in unexpected errors.

This pull request is to address this problem by removing the need of the intermediate temp file and execute the generated batch commands in place directly.

Also, please consider to back-port to kinetic-devel if it makes sense. Thanks!

### DIFF
--- a/cmake/templates/setup.bat.in
+++ b/cmake/templates/setup.bat.in
@@ -39,30 +39,8 @@ if not "!_PYTHON_FOUND!" == "!_PYTHON!" (
 )
 endlocal & set PATH=%PATH%
 
-REM generate pseudo random temporary filename
-:GenerateTempFilename
-REM replace leading space of time with zero
-set _SETUP_TMP=%Time: =0%
-REM remove time delimiters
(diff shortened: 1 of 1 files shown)


seanyen:
This is ready for review.

seanyen:
@dirk-thomas Please excuse me pinging you on this. This is ready for review and merge. We have seen some folks using `colcon` on `noetic` builds and this addition would be an improvement to the first time experience (so people don't need to limit the parallel build to `1` to workaround this.)

dirk-thomas:
Thanks for the patch.

dirk-thomas:
Cherry-picked to `kinetic-devel` in debf48220bb7b38177ab26ec9ef467b5d1083da7.

seanyen:
Thanks for the quick response!

PR #1117, opened 2020-08-18
[Windows] Use raw string literal to accommodate escape characters

seanyen:
On Windows, CMake will use the backslash `\` as the default character to compose the prefix path. However, when it is passed to the template, the `\` means differently as the escape character in Python. Therefore, a wrong string will be used in such case.

Currently, to workaround it, one needs to explicitly define the prefix path like `set CMAKE_PREFIX_PATH=c:/opt/ros/noetic`.

This pull request is to use the raw string literal to remove the need of using forward slash in such case, so the default CMake would work directly with `catkin`.

### DIFF
--- a/cmake/templates/_setup_util.py.in
+++ b/cmake/templates/_setup_util.py.in
@@ -268,7 +268,7 @@ if __name__ == '__main__':
 
         if not args.local:
             # environment at generation time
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thanks for the patch.

dirk-thomas:
Cherry-picked to `kinetic-devel` in 313322176752f3524a33c35f8656c16a2ad10114.

PR #1121, opened 2020-09-27
count 'skipped' tests as skipped

dseifert:
Previously only 'skip' and 'disabled' were counted as skipped tests.
However, junit format explicitly allows 'skipped' (cmp.
https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd#L196)

For reference, also see the same issue in colcon:
- Bug: https://github.com/colcon/colcon-test-result/issues/18
- PR: https://github.com/colcon/colcon-test-result/pull/19

This is based on noetic-devel, but applies to kinetic-devel as well (particularly, we are facing this issue under melodic and would appreciate a backport (or shall I create a new PR?))

### DIFF
--- a/python/catkin/test_results.py
+++ b/python/catkin/test_results.py
@@ -158,7 +158,7 @@ def _get_testsuite_stats(node):
     num_tests = int(node.attrib['tests'])
     num_errors = int(node.attrib['errors'])
     num_failures = int(node.attrib['failures'])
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Thanks for the patch. Yes, support for `skipped` was added to the schema in the recent years.

dirk-thomas:
Cherry-picked to `kinetic-devel` in 55dcf350def0964205b7007502ad3f8190838fa2.

PR #1122, opened 2020-09-28
set CATKIN_PACKAGE_LIBEXEC_DESTINATION which was documented but not set

### LINKED ISSUE #1118: CATKIN_PACKAGE_LIBEXEC_DESTINATION mentioned in docs is not implemented
The documentation mentions the CMake variable `CATKIN_PACKAGE_LIBEXEC_DESTINATION`.
https://github.com/ros/catkin/blob/dfc93c72d9799f57442d08c2f6e9fd494bba95c1/doc/user_guide/variables.rst#L53

In addition, we see the following mentions in configure files:
https://github.com/ros/catkin/blob/dfc93c72d9799f57442d08c2f6e9fd494bba95c1/cmake/templates/cfg-extras.context.py.in#L19
https://github.com/ros/catkin/blob/dfc93c72d9799f57442d08c2f6e9fd494bba95c1/cmake/templates/env-hook.context.py.in#L19

Looking through the code, we see no mention of such a variable.
https://github.com/ros/catkin/blob/dfc93c72d9799f57442d08c2f6e9fd494bba95c1/cmake/catkin_destinations.cmake

Is this a real variable or perhaps something that was envisioned but never implemented?

dirk-thomas:
Fixes #1118.


dirk-thomas:
Cherry-picked to `kinetic-devel` in 457a7cbe1885214905a79cdb11ea7d3b9c9e6d56.

PR #1126, opened 2020-11-19
cmake: symlink_install: detect and properly symlink versioned libraries

xqms:
Summary: versioned libraries do not end in ".so" but in ".so.X.Y.Z" and require an additional unversioned symlink to help CMake and the linker find it.

Here is a (quite popular) package that uses versioned libraries: https://github.com/ros-planning/moveit_visual_tools/blob/c166f8ee6412023b4183fc091abe6acbcdba2448/CMakeLists.txt#L57

Currently, the `catkin_symlink_install_targets` function will not detect the versioned library (in this case `libmoveit_visual_tools.so.3.6.0`) as a library, copying it to the wrong destination. This PR replaces the simple check for ".so" with a regex that also matches versioned libraries.

Furthermore, versioned libraries require an additional symlink from the unversioned name, otherwise CMake and/or the linker won't find the library. CMake's `install(TARGETS)` command seems to do that automatically, so we add logic to do that as well.

I didn't find a test case for symlinked installs - if there is one, I'd be happy to add a unit test.


xqms:
ping - just hit this bug again with https://github.com/ros-planning/moveit/blob/master/moveit_plugins/moveit_simple_controller_manager/CMakeLists.txt. Anything I can do to make review/merge easier?

sloretz:
Fixed in #1197

PR #1127, opened 2020-12-15
Added --force-color flag to catkin_make and catkin_make_isolated

stertingen:
This can be used to enforce colored output even on non-interactive output. `--force-color` can still be overriden by `--no-color`.

This PR is for CI environments where the output is stored in a file but a colored output makes it easier for a user to read the log.

### DIFF
--- a/bin/catkin_make
+++ b/bin/catkin_make
@@ -33,9 +33,14 @@ def main():
     apply_platform_specific_defaults(args)
     cmake_args = args.cmake_args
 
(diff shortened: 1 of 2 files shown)


stertingen:
There are some alternatives to the current implementation:
- `--force-color` overrides `--no-color`
- Error/Warning if `--force-color` and `--no-color` are both set

sloretz:
Thank you for the PR!

PR #1038, opened 2019-10-17
support build_type tag with condition attribute

dirk-thomas:
Based on https://answers.ros.org/question/335437/possible-to-build-pure-python-package-for-ros1-and-ros2-with-one-packagexmlcmakelists-file/

Getting the `build_type` through the `catkin_pkg` API (while evaluating the conditions before) ensures that only matching build types are considered.

Also the dependencies need to take the condition into account.

PR #1134, opened 2021-02-20
Evaluate <depend> conditions in CMake variables

matthew-reynolds:
Adds support for the `condition` attribute for `<*depend>` tags in package.xml format 3 ([REP014](https://www.ros.org/reps/rep-0149.html#id20))

#1038 added support for the `condition` attribute to the `<build_type>` tag, but it was never added to the other `<*depend>` tags.

### Without this change:
In Noetic (py3), package.xml containing 
```
  <exec_depend condition="$ROS_PYTHON_VERSION == 2">python-whatever</exec_depend>
  <exec_depend condition="$ROS_PYTHON_VERSION == 3">python3-whatever</exec_depend>
```
yields CMake variable: `${PROJECT_NAME}_EXEC_DEPENDS = "python-whatever;python3-whatever"`

### After this change:
The same package.xml yields CMake variable: `${PROJECT_NAME}_EXEC_DEPENDS = "python3-whatever"`

---

If approved, I would also like this to be considered for backport to kinetic & melodic.

### DIFF
--- a/cmake/parse_package_xml.py
+++ b/cmake/parse_package_xml.py
@@ -33,6 +33,7 @@
 from __future__ import print_function
 
 import argparse
+import os
 import sys
 from collections import OrderedDict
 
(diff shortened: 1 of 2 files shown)


sloretz:
Thank you for the PR!

sloretz:

* it only supports ROS 1

PR #1137, opened 2021-03-12
catkin pkg config template: performance fix for pack/unpack/append

aurzenligl:
**rationale**
These algorithms may be used O(M^2) times where M is the level of rospackage nesting, and each pack/unpack/append algorithm complexity are O(N^2)/O(N)/O(N^2) where N is libraries count. The fix allows to reduce them to O(1) complexity if we only consider amount of cmake statements executed. In a ros distro where level of nesting may exceed 10 and library count a couple of hundreds, cmake execution time can easily take more than 30 seconds, which becomes prohibitive.

**reproduction**
https://github.com/aurzenligl/study/tree/master/ros-catkin-pkgconfig
```
# noetic-devel
aurzenligl@study:/code/ros-catkin-pkgconfig$ ./compile.sh 
real	0m18.339s
user	0m16.333s
sys	0m1.994s
```
```
# aurzenligl/pkg-config-speedup
aurzenligl@study:/code/ros-catkin-pkgconfig$ ./compile.sh 
real	0m5.504s
user	0m4.378s
sys	0m1.146s
```

**notes**
Notes regarding REMOVE_DUPLICATES in list_append_unique and list_append_deduplicate are incorrect, in case of:
- list_append_unique: REMOVE_DUPLICATES uses set/unordered_set to remember encountered items, all output items are written in order, so the algorithm is "stable", it doesn't shuffle elements
- list_append_deduplicate: REMOVE_DUPLICATES is improper here, as it removes duplicates from the added element list, not the added-to element list, not due to it's "unstability" as note suggests

3.0.2:  https://github.com/Kitware/CMake/blob/v3.0.2/Source/cmListCommand.cxx#L480-L493
3.19.6: https://github.com/Kitware/CMake/blob/v3.19.6/Source/cmAlgorithms.h#L120-L130


aurzenligl:
@tfoote, @mjcarroll, should I open an issue associated with the PR?

sloretz:
Thank you for the PR!

sloretz:

* it only supports ROS 1

PR #1142, opened 2021-04-16
find libraries by name instead of using the imported location

### LINKED ISSUE #1141: catkin_package DEPENDS libs are imported by LOCATION only
External system libraries are exported via catkin_package DEPENDS to dependees via IMPORT_LOCATION only.

For example an external cmake project installed via an debian package with an cmake auto generated export config of the following form:

```cmake
add_library(external SHARED ...)
#will generate libexternal.so pointing to .so.1
#with .so.1 pointing to .so.1.2.3
set_target_properties(external PROPERTIES VERSION 1.2.3 SOVERSION 1)

#Do cmake like auto export with commands like these
install(TARGETS external EXPORT external-export)
install(EXPORT external-export ..

#will generate some external-export.cmake external-export-release.cmake
#with maybe a line like this:
set_target_properties(external PROPERTIES
  IMPORTED_LOCATION_RELEASE "${_IMPORT_PREFIX}/lib/libexternal.so.1.2.3"
  IMPORTED_SONAME_RELEASE "libexternal.so.1"
  )
```
Let's say the ros package 'external_ros' forwards external via depends.
This package is backwards compatible and a newer version maybe installed than a ros package was linked against at build time. The package itself faces no problem as it was correctly linked and will accept any newer package providing the .so.1 file.

But any package that now depends on external_ros as build dependency will now receive the libexternal.so.1.2.3 in their list although it is not present nor "installable". If the SONAME were queried if present the issue would not arise.

I think this happens when this is being called for external targets: https://github.com/ros/catkin/blob/a9672d78ec483c3a991a051c365da329fceaa9f2/cmake/catkin_package.cmake#L258

In the case of non windows targets this line is probably being executed:
https://github.com/ros/catkin/blob/a9672d78ec483c3a991a051c365da329fceaa9f2/cmake/catkin_libraries.cmake#L160

I do not know if there are any side effects for other platforms like windows otherwise I would have committed a pr. Also I have not checked if this affects ament too, but I guess it probably might be an issue there as well. Without a fix I can only recommend for custom libs using a fixed lib version or just the short soversion.

devrite:
Fixes #1141

Instead of directly using an imported location we use
the soname if possible and after that the location name
to find the library using the location directory as hint.

This allows us to use an soname, which may be different
and to the location name (and residing directory). This
allows to use system installed libraries which may be updated
but are compatible with the one used at built time
as long as the location and name in which they were found in
are the same.


cmake/catkin_libraries.cmake
```diff
@@ -114,6 +114,40 @@ function(catkin_unpack_libraries_with_build_configuration VAR)
...
+    list(APPEND imported_location_libnames ${imported_soname})
+  endif()
+  get_target_property(imported_location ${lib} IMPORTED_LOCATION${cfg})
+  if(imported_location)
+    get_filename_component(imported_location_name ${imported_location} NAME)
+    get_filename_component(imported_location_dir ${imported_location} DIRECTORY)
+    list(APPEND imported_location_libnames ${imported_location_name})
+  endif()
+  if(imported_location_libnames)
+    find_library(imported_location_library NAMES ${imported_location_libnames} HINTS ${imported_location_dir})
```
peci1:
```suggestion
    find_library(imported_location_library NAMES ${imported_location_libnames} HINTS ${imported_location_dir} NO_CACHE)
```
Otherwise the first found library will be returned for all calls of this function. See https://github.com/RoboStack/ros-noetic/pull/426 .

peci1 (changes requested):

cmake/catkin_libraries.cmake
devrite: I accepted it, but I did not follow up anymore it. So have not retested my use cases or if there is any situation in which cache needs to be used.

sloretz:
Thank you for the PR!

devrite:
No worries., I agree. If the feature is needed it is at least not lost and may be re-implemented in ament if not already.

Issue #490, opened 2013-07-25
make installed code relocatable


dirk-thomas:
Installed code should not (or as few as possible) contain absolute path information. Currently a lot of code contains the `CMAKE_INSTALL_PREFIX`.

One goal would be to compile a workspace with `DESTDIR=/tmp/destdir catkin_make_isolated --install` and be able to build another workspace on top of that.
At best it should be possible to run everything from the installed workspace and run the tests of workspaces built on top.

The following things need to be updated to not use absolute paths:
- [x] env.sh references setup.sh, 9cfba6ff417b561e2454ed1c8fca819643ec4892
- [x] setup.SHELL references setup.sh, 9cfba6ff417b561e2454ed1c8fca819643ec4892
- [x] setup.sh references _setup_util.py,
  sourced shell scripts can not determine their own path therefore enable the user to override the default path via an environment variable, using `CATKIN_SETUP_DIR`, 9cfba6ff417b561e2454ed1c8fca819643ec4892
- [x] _setup_util.sh contains abs. path to prefix CPP,
  should be passed in from setup.sh, 9cfba6ff417b561e2454ed1c8fca819643ec4892
- [x] generated CMake config files, e.g. include directories, 74af67d80936a1a4c565c25979687002dc0a99fa
- [x] environment hooks,
  require context about their workspace from the outside, setup.sh need to provide that individually for each environment hook, using the environemt variable `CATKIN_ENV_HOOK_WORKSPACE` for that 44b3234dbaa36f3210184ebea93c82d7a5ce3de4
- [x] catkin_make_isolated must set the environment variable to override the base path when building with `DESTDIR`, 7282a4bb832610e16d1a63c8336be41d781cb0e6

Changes required in other packages:
- [x] CMake extra files,
  should utilize `${@(PROJECT_NAME)_DIR}/../../..` to achieve the same as `@(CMAKE_INSTALL_PREFIX)`, ros/gencpp@e5acaf690d6276d427f6b0ecaab67b52f55bbe8a, ros/genlisp@42458c1d0a4ca1d3e78b8b1b2cac5e8fb2169a60, ros/genpy@1bc78ac23e465c1b3ef040d11c53d89cdc00b8ae, ros/ros@3681bd0ebd5d19f26421d8af844009e9970e9b58, ros/ros_comm@e0366e14128ed0083e9450bbe9f98fd9b2806454,
  ros/roslisp@58657c8827c70c3302c8755019500be2c2cdd95b
[...]

Issue #379, opened 2013-03-12
Don't set NO_CMAKE_FIND_ROOT_PATH in find_xxx


herbrechtsmeier:
Don't set NO_CMAKE_FIND_ROOT_PATH in find_xxx functions to enable the cross compiler to set the system roots to the target system root and host system root. Additionally don't assume that the CMAKE_PREFIX_PATH contains absolute paths  (IS_DIRECTORY) as they are relative to a system root.

herbrechtsmeier:
At the moment you have to set CMAKE_PREFIX_PATH to the absolute path (/path/to/machine/sysroot/path/to/ros;/path/to/native/sysroot/path/to/ros) but normally it should only contain the local prefix (/path/to/ros) and the system root paths (/path/to/machine/sysroot; /path/to/native/sysroot) are passed via CMAKE_FIND_ROOT_PATH. This allows the cross compiler framework to install cross compiled libraries and headers into the machine system root and binaries and scripts into the native system root. Additionally the cross compiler framework installs the CMake files into the correspondent system root.

tfoote:
Can you point to docs that we can read more about the best practices?

herbrechtsmeier:
At first you could take a look at the http://www.vtk.org/Wiki/CMake_Cross_Compiling but OpenEmbedded (Yocto) have simplify the usage of executables in the build process. They first build all buildtool_depends as native version and install them into a native sysroot. As next they cross compile all depends and install all header, libraries and CMake files into the cross sysroot. The native sysroot is append to the cross sysroot in the `CMAKE_FIND_ROOT_PATH` and `CMAKE_FIND_ROOT_PATH_MODE_PROGRAM` is set to `ONLY`. This allows the find_program function to find the native programs in the native sysroot. I will try to find a documentation about the OpenEmbedded internals.

herbrechtsmeier:
@tfoote This is the common cmake configuration of yocto: http://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/classes/cmake.bbclass

herbrechtsmeier:
All paths in the CMake files should be generated by the find_xxx macro and don't set NO_CMAKE_FIND_ROOT_PATH outside of the develspace. This allows the CMake files to work with different DESTDIR (sysroots) for programs and libraries.

pkgConfig.cmake.in:

```
if(@DEVELSPACE@)
[...]

PR #1145, opened 2021-05-11
Remove usage of NO_CMAKE_FIND_ROOT_PATH to enable cross compilation.

windelbouwman:
It's not fully clear to me why `NO_CMAKE_FIND_ROOT_PATH` was used. I think it was done for consistency, since it's used everywhere in the same way. I believe this option is not needed, and even should not be there.

The option not being used has these advantages:
- Less code in the cmake files
- Cross compilation works as expected when using a sysroot folder

### DIFF
--- a/cmake/templates/pkgConfig.cmake.in
+++ b/cmake/templates/pkgConfig.cmake.in
@@ -157,7 +157,7 @@ foreach(library ${libraries})
     foreach(path @PKG_CONFIG_LIB_PATHS@)
       find_library(lib ${library}
         PATHS ${path}
-        NO_DEFAULT_PATH NO_CMAKE_FIND_ROOT_PATH)
+        NO_DEFAULT_PATH)
       if(lib)
         set(lib_path ${path})
         break()


meyerj:
I agree with @windelbouwman that the `NO_CMAKE_FIND_ROOT_PATH` flags should probably be removed: Without, cross-compilation or staged installation of a full workspace would work more or less out of the box, by setting the `DESTDIR` environment variable and the [`CMAKE_FIND_ROOT_PATH`](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH.html) CMake variable to the location of the staged install tree. Unless someone explicitly sets `CMAKE_FIND_ROOT_PATH`, adding `NO_CMAKE_FIND_ROOT_PATH` should have no effect anyway. And if it is set, than finding packages, libraries, headers etc. at that location is exactly what we want, even though `CMAKE_PREFIX_PATH` or `<pkg>_DIR` may say something different.

From the [documentation of `find_package()`](https://cmake.org/cmake/help/latest/command/find_package.html):
> The CMake variable [CMAKE_FIND_ROOT_PATH](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH.html#variable:CMAKE_FIND_ROOT_PATH) specifies one or more directories to be prepended to all other search directories. This effectively "re-roots" the entire search under given locations. Paths which are descendants of the [CMAKE_STAGING_PREFIX](https://cmake.org/cmake/help/latest/variable/CMAKE_STAGING_PREFIX.html#variable:CMAKE_STAGING_PREFIX) are excluded from this re-rooting, because that variable is always a path on the host system. By default the [CMAKE_FIND_ROOT_PATH](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH.html#variable:CMAKE_FIND_ROOT_PATH) is empty.
> 
> The [CMAKE_SYSROOT](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSROOT.html#variable:CMAKE_SYSROOT) variable can also be used to specify exactly one directory to use as a prefix. Setting [CMAKE_SYSROOT](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSROOT.html#variable:CMAKE_SYSROOT) also has other effects. See the documentation for that variable for more.
>
> These variables are especially useful when cross-compiling to point to the root directory of the target environment and CMake will search there too.

So why would one want to explicitly disable that behavior?

But the occurence of `NO_CMAKE_FIND_ROOT_PATH` in [`pkgConfig.cmake.in`](https://github.com/ros/catkin/blob/2612c4ce82a285cac13263576f8a85b8a6d6cd3d/cmake/templates/pkgConfig.cmake.in) is not the only one, and also others in [`catkinConfig.cmake`](https://github.com/ros/catkin/blob/2612c4ce82a285cac13263576f8a85b8a6d6cd3d/cmake/catkinConfig.cmake.in) would need to be removed. Next to that, in [some places `catkin`](https://github.com/ros/catkin/blob/2612c4ce82a285cac13263576f8a85b8a6d6cd3d/cmake/all.cmake#L60) or also other packages' CMake code (e.g. `genmsg`) expect to find files at paths below an entry of `CMAKE_PREFIX_PATH` or `CATKIN_WORKSPACES`, without using `find_*()` commands and without also checking the same paths prefixed by [`CMAKE_FIND_ROOT_PATH`](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH.html). That is doable with a few more patches, for example by including the prefixed paths to `CATKIN_WORKSPACES` if those directories exist, and keep `CMAKE_PREFIX_PATH` as a list of underlays without the staging prefix, that is used by CMake internally according to the docs and for generating setup and config files.

There is no problem with the current version of catkin to build and install _package by package_ with `DESTDIR`, and to install each package _individually_ before building the next one. In this case catkin and CMake can find already installed packages at their final location. But in order to build a full workspace at once, each package needs to be able to find its dependencies that have been installed to `DESTDIR` before. And even if they can be found, because `CMAKE_PREFIX_PATH` or `_CATKIN_SETUP_DIR` is set such that they include the staging dir, those path components must be stripped again from all generated files. If I only set `DESTDIR` with `catkin_make_isolated` the build succeeds (likely thanks to , but after copying the resulting install-space to the final location and sourcing `setup.sh`, `CMAKE_PREFIX_PATH` and other environment variables contain both, the final locations and the original staging locations.

Or does anyone have a working solution without the need to patch `catkin`? Older issues and especially https://github.com/ros/catkin/issues/490 suggest that this use case should already be supported. But I could not get it to work without setting `CMAKE_FIND_ROOT_PATH` and without still finding traces of the `DESTDIR` in the installed files, especially in `<prefix>/_setup_util.py` and generated `<pkg>Config.cmake`. I remember that it was already working for me in the past (https://github.com/ros/catkin/pull/553), but maybe I did not pay attention to the content of generated files back then.

[`CMAKE_STAGING_PREFIX`](https://cmake.org/cmake/help/latest/variable/CMAKE_STAGING_PREFIX.html) may be another approach to divert the install space and affect `find_*()` commands. It is not clear to me yet to what degree it differs in its intended usage from `DESTDIR` and [`CMAKE_FIND_ROOT_PATH`](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_ROOT_PATH.html). Both seem to have similar goals. If only `CMAKE_STAGING_PREFIX` is set, but not `DESTDIR`, some files are still generated in `CMAKE_INSTALL_PREFIX` directly.

Is there still interest in such a patch, given the that this pull request is already open for more than a year and the last release was in 2021? I got quite far with only minimal changes, basically those removals of `NO_CMAKE_FIND_ROOT_PATH` and checking `CMAKE_FIND_ROOT_PATH` when populating `CATKIN_WORKSPACES` from `CMAKE_PREFIX_PATH` [here](https://github.com/ros/catkin/blob/2612c4ce82a285cac13263576f8a85b8a6d6cd3d/cmake/all.cmake#L60). But it would require some time to finalize and document the patch, and eventually consider other variables like [`CMAKE_SYSROOT`](https://cmake.org/cmake/help/latest/variable/CMAKE_SYSROOT.html), too.

I personally do not have the use case of cross-compiling in mind, as mentioned in the title of this PR. Instead I want to build full workspace binary distributions for later installation, eventually in multiple stages without mixing the new packages with others already installed at the final destination using [docker multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/). I assume that the requirements are very similar, and that both use cases would work if catkin could distinguish cleanly between paths used at build time (e.g. `DESTDIR`, `CMAKE_FIND_ROOT_PATH`, `CATKIN_WORKSPACES`, `CMAKE_SYSROOT`, ...) and paths where the packages will later be installed and found (`CMAKE_PREFIX_PATH`, `CMAKE_INSTALL_PREFIX`, ...).

Related issues:
- https://github.com/ros/catkin/issues/379
- https://github.com/ros/catkin/issues/490
- https://github.com/ros/catkin/pull/553
- https://github.com/ros/catkin/pull/811

sloretz:
Thank you for the PR!

sloretz:

* it only supports ROS 1

PR #1153, opened 2021-10-16
Fix handling of multiple namespaced python packages in devel space

rhaschke:
If there are multiple python packages to be installed below a [common namespace](https://packaging.python.org/guides/packaging-namespace-packages/#pkgutil-style-namespace-packages), only the last one installed is found in devel space, because the path extension is always overwritten.
This PR augments the existing `__extended_path` instead of overriding it.
We need this feature for MoveIt, which has multiple python packages to be installed in the common namespace `moveit`.


rhaschke:
Hm. Unfortunately, this works only with catkin tools _and_ merged devel space as well as with `catkin_make`. :frowning: 
`catkin tools` with a linked devel space fails because linking multiple `__init__.py` files to the same final devel location fails.
`catkin_make_isolated` fails because the init files are installed in different locations and only the first one on `PYTHONPATH` will define its src extension path.

sloretz:
Thank you for the PR!

Issue #1155, opened 2021-10-25
`catkin_install_python` fails when shebang is followed by a space


maetulj:
I am porting a python script from melodic to noetic. To convert the shebang from python to python3 I am using `catkin_install_python` as suggested in [the migration guide](http://wiki.ros.org/UsingPython3/SourceCodeChanges#Changing_shebangs).

The function successfully converted all the shebangs from `#!/usr/bin/env python` to `#!/usr/bin/env python3` except for one file where it failed. After comparing it for some time I noticed that the erroneous script has a shebang ending with a space character before the new line (represented here as &#9608;):
`#!/usr/bin/env python`&#9608;`\n`

This breaks the `catkin_install_python` function and the "installed" file starts with an incorrect  shebang.

Rayman:
If you want, you could contribute a PR to adjust the regex for this:
https://github.com/ros/catkin/blob/7fa3eb3508ba12c34d85b8e54bbdf4bbdc60a5c1/cmake/catkin_install_python.cmake#L30

ivanpauno:
Fixed in https://github.com/ros/catkin/pull/1156.

PR #1156, opened 2021-10-27
Fix catkin install python shebang whitespace

maetulj:
Fixes the issue #1155 where a whitespace character prevents `catkin_install_python` function to update the shebang line with the correct python version. The regex searching for the shebang line has been updated to match the shebang line even if it is followed by multiple spaces and/or tabs.

The line must end with a new line and no other characters except a space and/or tabs are allowed between the shebang and the newline.

The spaces and/or tabs after the shebang are preserved in the converted file (lookahead in regex which would prevent it to be included [is not supported in cmake](https://cmake.org/Bug/view.php?id=15961)).

### DIFF
--- a/cmake/catkin_install_python.cmake
+++ b/cmake/catkin_install_python.cmake
@@ -27,7 +27,7 @@ function(catkin_install_python signature)
       stamp(${source_file})
       # read file and check shebang line
       file(READ ${source_file} data)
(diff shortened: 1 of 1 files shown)


ivanpauno (approved):
lgtm!

thanks for the contribution!

PR #1171, opened 2022-04-13
Mark include directories as system directories

kunaltyagi:
Allows use of warning flags without the compiler complaining about gtest/gmock related issues

### DIFF
--- a/cmake/test/gtest.cmake
+++ b/cmake/test/gtest.cmake
@@ -175,7 +175,7 @@ function(_catkin_add_executable_with_google_test type target)
   endif()
 
   # create the executable, with basic + gtest/gmock build flags
-  include_directories(${${type_upper}_INCLUDE_DIRS})
+  include_directories(SYSTEM ${${type_upper}_INCLUDE_DIRS})
   link_directories(${${type_upper}_LIBRARY_DIRS})
   add_executable(${target} ${ARG_UNPARSED_ARGUMENTS})
   if(ARG_EXCLUDE_FROM_ALL)


gbiggs:
Can you please provide some sample output of before and after this change that demonstrates what effect it has?

kunaltyagi:
Haven't seen this issue with `catkin build`, but with `colcon build`. 

I've seen two difference between their behaviors:
* catkin created an imported target for gtest/gmock etc, while colcon creates a new target per package with the tests
* warnings (via compiler flags) are propagated across `include <header >` for colcon, but not catkin

I don't know if these 2 are related, but adding SYSTEM seems to have stopped the warnings for me

ivanpauno:
We have avoided using `SYSTEM` in the past as it has some unexpected side effects, see for example https://github.com/ament/ament_cmake/pull/321#issuecomment-830817601.

kunaltyagi:
Reading that chain, towards the end, after investigation, they have accepted that this makes sense for ament.

I don't feel there would be any unexpected side-effects (as pointed out later):
* We can't control what flags are used by users
* Errors/Warnings in gtest shouldn't affect user code unless someone has that in the same overlay
* Later inclusions on both -I and -isystem override the earlier ones (left to right). This allows proper usage with overlays

sloretz:
Thank you for the PR!

Unfortunately I don't think we should merge this one. Catkin workspaces are very sensitive to `-I` vs `-isystem` with regards to overriding packages. Given that [ROS Noetic will reach end-of-life on May 31st, 2025](https://discourse.ros.org/t/ros-noetic-end-of-life-may-31-2025/43160), and that this change seems likely to create a hard to find regression, I'm going to close this one.

PR #1176, opened 2022-06-25
Import setuptools before distutils

lopsided98:
Setuptools wants to be imported before distutils, and prints a warning if it is not. Starting with a recent version (>57.2, <=61.2), not following this rule causes catkin's own monkey patching to fail. This commit reverses the import order to fix the warning and patching issue.


gbiggs:
Can you please supply a list of commands to run and the necessary environment to verify this change works?

sloretz:
Thank you for the PR!

sloretz:

* it only supports ROS 1

PR #1180, opened 2022-09-19
Default to number of usable CPUs if possible

wico-silva:
`multiprocessing.cpu_count()` returns the number of CPUs, usable or not. While in most situations this is good enough it fails when not all CPUs are usable. An example using docker:

1. Start a container: `docker run -it --rm --cpuset-cpus="0" ros:melodic-perception-bionic python3`
2. Run:
```
import multiprocessing
multiprocessing.cpu_count()
```

It will return the number of CPUs of the whole system instead of `1`, the number of CPUs that are available.
Consequently, when using `catkin_make` it tries to use more CPUs than the ones available and the build ends up failing.

Expected behavior is:
```
import os
len(os.sched_getaffinity(0))
```
Returns `1`.

### DIFF
--- a/python/catkin/builder.py
+++ b/python/catkin/builder.py
@@ -322,7 +322,14 @@ def handle_make_arguments(input_make_args, append_default_jobs_flags=True):
             elif append_default_jobs_flags:
                 # Else Use the number of CPU cores
                 try:
-                    jobs = multiprocessing.cpu_count()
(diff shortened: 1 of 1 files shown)


gbiggs (approved):

gbiggs:
Thanks for the contribution!

PR #1186, opened 2023-05-06
Build with cmake to include setup.sh

jspricke:


sloretz:
Could you explain what you mean by including `setup.sh`? Catkin has a special release process where prefix-level setup scripts are installed via this patch https://github.com/ros-gbp/catkin-release/blob/patches/debian/noetic/catkin/0001-Patch-rules-template-to-create-setup-files.patch , so I'm very hesitant to change the build type of the package without understanding more

jspricke:
This is basically doing the same as your patch if you compare: https://github.com/ros-infrastructure/bloom/blob/master/bloom/generators/debian/templates/catkin/rules.em#L35 and https://github.com/ros-infrastructure/bloom/blob/master/bloom/generators/debian/templates/cmake/rules.em#L38

AFAIR both have two consequences:
- it allows bootstrapping the ROS world, as catkin can't depend on itself (and does not).
- it includes a setup.sh in the catkin package (as is the case with your patch).

sloretz:

* it only supports ROS 1

PR #1193, opened 2024-04-04
Use the newer unittest.mock from standard library on Python 3

a-detiste:
https://github.com/testing-cabal/mock

mock is now part of the Python standard library, available as unittest.mock in Python 3.3 onwards.


sloretz (approved):
Thank you for the PR!

Issue #1164, opened 2022-02-11
`distutils`: Deprecation warning


acxz:
```
/opt/ros/noetic/share/catkin/cmake/interrogate_setup_dot_py.py:43: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
  import distutils.core
```

catkin version: 0.8.10
py version: 3.10.2

gbiggs:
As there are still likely to be a a _lot_ of packages using `distutils` in Noetic, I do not think it is safe to remove support for it at this time. Python 3.10 has just been released and 3.12 is not yet scheduled, so we still have a long time before it will become unavailable in the most up-to-date version on Python. On top of that, Noetic targets Python 3.8, so it is not required to handle a Python version where `distutils` has been removed.

A better time to do this is probably in a post-Noetic version of catkin.

tfoote:
Also before removing the support it would be best to provide a deprecation warning to users with a stated interval by which they will need to have migrated forward.

acxz:
> packages using distutils in Noetic

Do you know of a smart way to find out which Noetic packages use distutils?

> provide a deprecation warning

I can create a PR with a deprecation warning when distutils gets imported. Is there documentation on a proper format to coding this deprecation warning?

sloretz:
ROS 1 is end-of-life (EOL) as of today, May 31st 2025. I am archiving this repository because:

* it only supports ROS 1
* it isn't needed anymore in ROS 2

[If you still rely on ROS 1, read this page to learn about your options.](https://bit.ly/NoeticEOL)

Issue #1162, opened 2022-01-14
How to support poetry / pyproject.toml?


dimaqq:
Hi, I'd love to be able to write ros2/catkin-able components that use modern dependency management tools.

Specifically, I'd love to specify dependencies in `pyproject.toml` and use `poetry` to maintain/build/test the package.
(that would be as an alternative to `setup.py` and `

I could totally contribute some code towards this.

It's not yet supported, is it?

How do I go about that?

Rayman:
Do you want to use `pyproject.yaml` to install dependencies? If so you should forward your request to https://github.com/ros-infrastructure/rosdep

Or to you want to replace `setup.py`? I don't think that will get implemented for ROS1 because it's end of line. You have a better chance to have https://github.com/colcon support it.

dimaqq:
I was thinking about the latter.
Allowing modern 3p packages to be used in ros2 projects, for robotics devs.

The earlier is interesting too, and quite straightforward, may be a good first step 😉

Rayman:
I found a related issue in the colcon project: https://github.com/colcon/colcon-core/issues/208

ivanpauno:
> Allowing modern 3p packages to be used in ros2 projects, for robotics devs.

For ros2 see:

- https://github.com/colcon/colcon-core/issues/208
- https://github.com/colcon/colcon-core/issues/454

For ros1, I don't think this is going to be done, as noetic was the last release.

PR #1194, opened 2024-04-10
Avoid distutils-related build time spam on Python 3.11

mikepurvis:
Without this change, Python 3.11 platforms get this on every build— not just of `catkin` itself, but every package that uses it:

```
/home/mikepurvis/roslib_ws/install/share/catkin/cmake/interrogate_setup_dot_py.py:43: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
  import distutils.core
```

This has been discussed previously in https://github.com/ros/catkin/issues/1164#issuecomment-1039650945, but I'm hopeful that this change will resolve the issue for Python 3.11+ users while allowing any remaining distutils packages to continue functioning.

Note that this does not address the other deprecation warning related to catkin invoking `setup.py` directly. There was some discussion of that in https://github.com/ros/catkin/issues/1162, but it will definitely be a separate PR, attempting to move `cmake/templates/python_distutils_install.sh.in` to a pypi/build workflow.

### DIFF
--- a/cmake/interrogate_setup_dot_py.py
+++ b/cmake/interrogate_setup_dot_py.py
@@ -32,26 +32,14 @@
 
 from __future__ import print_function
 
+import importlib
+import contextlib
 import os
 import runpy
 import sys
 from argparse import ArgumentParser
 
-setup_modules = []
-
-try:
-    import distutils.core
-    setup_modules.append(distutils.core)
-except ImportError:
-    pass
-
-try:
-    import setuptools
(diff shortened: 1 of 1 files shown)


mikepurvis:
Building our larger workspace with this change has revealed that it definitely still breaks with ROS Python packages using distutils— I'll modify the patch to conditionally import distutils in the case where the package's `setup.py` actually uses it.

